PR #2, opened 2016-06-15
Don't support DWARF 1

tromey:
I think DWARF 1 was different enough that it can't be parsed by a DWARF 2+ parser.  So, reject it.


fitzgen:
Gracias

PR #19, opened 2016-07-20
Parse abbreviations on a per-unit basis

### LINKED ISSUE #5: Need a method to obtain the abbreviations for a CompilationUnit
Should be able to replace https://github.com/philipc/rust-dwarf/blob/master/benches/bench.rs#L48-L50 with
`let abbrevs = unit.abbreviations(&debug_abbrev);`
or
`let abbrevs = debug_abbrev.abbreviations(unit.debug_abbrev_offset());`
(not sure which way is best).

fitzgen:
Previously, we would only parse abbreviations in an all-or-nothing
manner from the start of the `.debug_abbrev` section. This isn't even
valid in the face of padding between units' abbreviations. This commit
uses the unit header's `debug_abbrev_offset` to find the correct
abbreviations for that unit.

Fixes #5

Does this look correct @philipc?

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -25,11 +25,18 @@ fn read_section(section: &str) -> Vec<u8> {
 
 #[bench]
 fn bench_parsing_debug_abbrev(b: &mut test::Bencher) {
+    let debug_info = read_section("debug_info");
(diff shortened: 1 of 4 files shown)


philipc:
LGTM

fitzgen:
Thanks!

PR #22, opened 2016-07-25
Delete input structs

philipc:
@fitzgen are these leftover from using nom?


fitzgen:
Yes these are leftovers that I forgot to remove, thanks!

PR #23, opened 2016-07-26
Abbrev module

philipc:
This is an attempt at creating an abbrev module with a tests submodule. The abbrev module has required me to make various things in parser public, and the tests submodule has done the same for things in the abbrev module. So as a result, I need to restrict what gets exported in the toplevel lib.rs. It's all a bit ugly though. I need to think about this for a bit to see if there's a better way. Interested in your thoughts.

There's a few other useful changes in this pull request too. I'm happy to rebase them if we decide not to do the module changes.

### DIFF
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -48,7 +48,7 @@ impl<'input, Endian> DebugAbbrev<'input, Endian>
                          debug_abbrev_offset: DebugAbbrevOffset)
                          -> ParseResult<Abbreviations> {
         let input: &[u8] = self.debug_abbrev_section.into();
-        parse_abbreviations(&input[debug_abbrev_offset.0 as usize..]).map(|(_, abbrevs)| abbrevs)
+        Abbreviations::parse(&input[debug_abbrev_offset.0 as usize..]).map(|(_, abbrevs)| abbrevs)
     }
 }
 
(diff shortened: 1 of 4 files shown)


philipc:
I'm not sure I fully understand the visibility rules yet, but I have learned some more.

It appears the `use super::*` glob only uses public functions, but you can still access private functions if you use them directly. Listing out all the private functions in the test module is a bit awkard though.

Alternatively, we can change the functions to be private methods instead. I think this makes a lot of sense, and I've pushed another commit to do that.

That fixes up the abbrev/tests module. There's still all the uses listed out in lib.rs. I think this is a sign that the modules aren't structured properly, and we can probably avoid them by making abbrev a submodule of parser.

fitzgen:
> It appears the use super::* glob only uses public functions, but you can still access private functions if you use them directly.

Oh, wow I didn't know that.

src/abbrev.rs
```diff
@@ -0,0 +1,761 @@
...
+    }
+
+    /// Get this abbreviation's tag.
+    #[inline]
+    pub fn tag(&self) -> constants::DwTag {
+        self.tag
+    }
+
+    /// Return true if this abbreviation's type has children, false otherwise.
+    #[inline]
```
fitzgen: Do all these `#[inline]` annotations improve any of the extant benches?

src/abbrev.rs
```diff
@@ -0,0 +1,761 @@
...
+        }
+    }
+
+    /// Parse a series of attribute specifications, terminated by a null attribute
+    /// specification.
+    fn parse_attributes(mut input: &[u8]) -> ParseResult<(&[u8], Vec<AttributeSpecification>)> {
+        let mut attrs = Vec::new();
+
+        loop {
+            let result = AttributeSpecification::parse_null(input).map(|(rest, _)| (rest, None));
```
fitzgen: Unrelated to this PR: I wonder if the performance issues boil down to attempting to `parse_null` and then attempting to `parse` (non-null) if that failed so we end up calling `parse_unsigned_leb` twice as many times as needed. We could fix this by combining the two methods into one that returns `ParseResult<(&[u8], Option<Abbreviation>)>` where the `Option` part is `None` if it is a null abbreviation.

fitzgen:
Thanks!

src/abbrev.rs
philipc: Yes, the annotations on name()/form() were required to due to the switch to using them in parser.rs, and others improved the improved the benchmark. I think a micro benchmark for a single attribute would help highlight these.

src/abbrev.rs
philipc: Yeah, I wondered that too. I did attempt it for null abbreviations, but I don't think they have much impact on the current benchmark, so it was hard to tell and I left that for later.

PR #24, opened 2016-07-26
Dont reparse same lebs multiple times

fitzgen:
Before:
```
test bench_parsing_debug_abbrev ... bench:      33,643 ns/iter (+/- 6,243)
test bench_parsing_debug_info   ... bench:   5,914,364 ns/iter (+/- 819,837)
```

After:
```
test bench_parsing_debug_abbrev ... bench:      29,905 ns/iter (+/- 7,143)
test bench_parsing_debug_info   ... bench:   5,913,803 ns/iter (+/- 894,585)
```

Looks like a little bit of a speed up for parsing abbreviations, but its within the margin of error. Either way, I think this actually cleans up the loop-until-we-parse-a-null-thing code a bit.

What do you think @philipc?

### DIFF
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -95,10 +95,7 @@ impl Abbreviations {
         let mut abbrevs = Abbreviations::empty();
 
         loop {
-            let result = Abbreviation::parse_null(input).map(|(rest, _)| (rest, None));
-            let result =
-                result.or_else(|_| Abbreviation::parse(input).map(|(rest, a)| (rest, Some(a))));
-            let (rest, abbrev) = try!(result);
+            let (rest, abbrev) = try!(Abbreviation::parse(input));
             input = rest;
 
             match abbrev {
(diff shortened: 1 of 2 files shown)


src/abbrev.rs
```diff
@@ -683,34 +648,17 @@ mod tests {
...
-            otherwise => panic!("Unexpected result: {:?}", otherwise),
-        };
-    }
-
-    #[test]
     fn test_parse_attribute_specifications_name_zero() {
         let buf = [0x00, 0x01, 0x00, 0x00];
         match AttributeSpecification::parse(&buf) {
-            Err(Error::AttributeNameZero) => {}
+            Err(Error::ExpectedZero) => {}
```
philipc: Not sure about this change since it makes the handling for zeroes in name/form asymmetrical. As it is, the AttributeNameZero enum value is now unused and could be deleted.

philipc:
Nice cleanup

fitzgen:
Thanks for checking this out! I removed the unused error variants, and I'm going to merge now.

src/abbrev.rs
fitzgen: I don't feel strongly about one error variant being `ExpectedEof` and the other being `AttributeFormZero` -- if you do, please push follow ups :)

PR #26, opened 2016-07-31
Cursor performance

philipc:
This gives significant improvements to the debug_info benchmark (see numbers in the commits).

I haven't updated any of the examples to use `current_ref`, because it is quite unergonomic, due to the borrow checker: you can't call `next_dfs` and `next_sibling` while you still have a reference to the current entry. See the changes in `bench.rs` for example. It's also much worse if you want to use `while` instead of `loop`.

So we probably need to rethink the EntriesCursor API. I think the best way will be to avoid the user needing to call `current_ref`, and instead `next_dfs` and `next_sibling` should return the entry ref too. This means we need to change them to allow them to return the first entry when they are first called, rather than immediately skipping the current entry. This is more in line with how the `next` method works for iterators. I'm open to other ideas though, since I don't know what requirements you had when designing the current API.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2046,10 +2046,7 @@ pub struct EntriesCursor<'input, 'abbrev, 'unit, Endian>
     input: &'input [u8],
     unit: &'unit UnitHeader<'input, Endian>,
     abbreviations: &'abbrev Abbreviations,
-    cached_current: RefCell<Option<ParseResult<DebuggingInformationEntry<'input,
-                                                                         'abbrev,
-                                                                         'unit,
-                                                                         Endian>>>>,
+    cached_current: RefCell<Option<DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>>>,
 }
 
 impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endian>
@@ -2064,8 +2061,7 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
         {
             let cached = self.cached_current.borrow();
             if let Some(ref cached) = *cached {
-                debug_assert!(cached.is_ok());
-                return Some(cached.clone());
+                return Some(Ok(cached.clone()));
             }
         }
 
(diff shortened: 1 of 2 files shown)


philipc:
It's also worth noting that prior to these changes, the check for `current.after_attrs.get()` in `next_dfs` never succeeded, because the caller always iterated attributes on a clone.

src/parser.rs
```diff
@@ -2083,14 +2083,25 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
...
     }
 
+    /// Get the entry that the cursor is currently pointing to.
+    pub fn current<'me>
+        (&'me mut self)
+         -> Option<ParseResult<DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>>> {
+        match self.current_ref() {
+            Some(Ok(current)) => Some(Ok(current.clone())),
+            Some(Err(e)) => Some(Err(e)),
+            None => None,
```
fitzgen: I guess this is more readable than using `map`:

```rust
self.current_ref().map(|res| res.map(|cur| cur.clone()))
```

Are there type errors if we do

```rust
match self.current_ref() {
    Some(Ok(current)) => Some(Ok(current.clone())),
    otherwise => otherwise,
}
```

instead? I guess the type of `otherwise` will be `Option<ParseResult<&DebuggingInformationEntry>>` and not `Option<ParseResult<DebuggingInformationEntry>>` even though we know it doesn't actually contain the variant that differs between the two.

src/parser.rs
```diff
@@ -2219,19 +2219,20 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
...
+                    for _ in self.cached_current.as_ref().unwrap().attrs() {
                     }
-                    current.after_attrs
+                    self.cached_current.as_ref().unwrap().after_attrs
                         .get()
                         .expect("should have after_attrs after iterating attrs")
                 };
 
-                let mut delta_depth = if current.abbrev.has_children() {
+                let mut delta_depth = if self.cached_current.as_ref().unwrap().abbrev.has_children() {
```
fitzgen: So many `self.cached_current.as_ref().unwrap()` -- perhaps we should make a little `unwrap_current_ref()` helper method?

fitzgen:
> It's also worth noting that prior to these changes, the check for current.after_attrs.get() in next_dfs never succeeded, because the caller always iterated attributes on a clone.

D'oh... Good catch.

fitzgen:
> So we probably need to rethink the EntriesCursor API. I think the best way will be to avoid the user needing to call current_ref, and instead next_dfs and next_sibling should return the entry ref too. This means we need to change them to allow them to return the first entry when they are first called, rather than immediately skipping the current entry. This is more in line with how the next method works for iterators. I'm open to other ideas though, since I don't know what requirements you had when designing the current API.

This sounds good to me. I didn't have any requirements other than a vague idea that by breaking things apart we could provide better performance than when they were stuck together, and that we could later provide nicer iterators (eg `DfsIter` and `SiblingsIter`) on top of this that maybe sacrificed a little perf for the nice API. If my inkling was way off mark, then lets definitely change things up!

fitzgen:
Going to go ahead and merge this and we can do follow ups as needed.

src/parser.rs
philipc: Agreed that we need to do something... I didn't spend much time on making it clean because it still needs other changes. Same for your other comment. I hadn't even run rustfmt on this. I'll push a commit for rustfmt so that it doesn't conflict too much with your branch.

PR #27, opened 2016-07-31
Add support for .debug_line and line number programs

fitzgen:
Very much still a work in progress. So far this can parse the headers in front of each line number program (see section 6.2.4 of the standard). Doesn't actually unpack and execute the line number program itself yet.

@philipc, want to take a look? Just want to give a chance to provide feedback on API and types and all that before I get too far.


tests/parse_self.rs
```diff
@@ -50,3 +53,46 @@ fn test_parse_self_debug_info() {
...
+
+        let mut cursor = unit.entries(&abbrevs);
+
+        let unit_entry = cursor.current_ref()
+            .expect("Should have a root entry")
+            .expect("and it should parse OK");
+
+        if let Some(attr) = unit_entry.attrs()
+            .map(Result::unwrap)
+            .find(|at| at.name() == DW_AT_stmt_list) {
```
philipc: Use `unit_entry.attr_value(DW_AT_stmt_list)` instead.

tests/parse_self.rs
```diff
@@ -50,3 +53,46 @@ fn test_parse_self_debug_info() {
...
+            // For whatever reason, rustc generated DW_FORM_data4 typed
+            // attributes for DW_AT_stmt_list attributes, when it seems to me
+            // like the more correct choice would be DW_FORM_sec_offset (the
+            // spec seems to agree -- see pages 214-215). Because of this, we
+            // have to turn the data into an offset.
```
philipc: I think rustc defaults to DWARF 2.0, which did not have DW_FORM_sec_offset.  We probably should include test fixtures for both versions. This is also why there is no .debug_types section.

philipc:
I think the primary API that the user will want is a cursor for line numbers, which they should be able to obtain using `compilation_unit.line_numbers(debug_line)`. The cursor should provide `next_line` as the lowest level operation, with `next_sequence` and `next_basic_block` built on top of that.

I imagine there should also be a fast way to obtain a cursor at the line corresponding to a given address, but I haven't read enough of the spec to know how to do that.

fitzgen:
Thanks for the feedback. Pushed a new commit fixing the things you pointed out.

Agreed that we want to provide a low level operation and build others on top of it. Not 100% sure whether that low level operation is `next_line` or maybe `next_row` for each row of the matrix that is being expanded by the line number program. There are many more rows in the matrix than lines, I think. Hopefully things will clear up some more as I dig in.

fitzgen:
@philipc -- I think this is pretty much ready to rebase on master and merge, want to take a final look? I separated the opcode parser from the execution of opcodes, which forced writing out a proper enum for opcodes, which involves a lot of busy work in the library but makes things cleaner overall (and especially for library users). The lowest level building blocks are the opcode parser and the iterator yielding each row of the line number matrix.

Note that the API is not how I want it to remain. Right now, you have to pull the data out of the compilation unit header and compilation unit DIE to create a LineNumberProgramHeader. It is not pretty. We should just encapsulate all that in a nice little helper method for library consumers. I plan to do that in a follow up, but I just want to get this merged into master and not worry about diverging anymore. I also plan on adding more doc examples and unit tests in follow ups.

src/line.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// # let buf = [0x00, 0x01, 0x02, 0x03];
+    /// # let read_debug_line_section_somehow = || &buf;
+    /// let debug_line = DebugLine::<LittleEndian>::new(read_debug_line_section_somehow());
+    /// ```
+    pub fn new(debug_line_section: &'input [u8]) -> DebugLine<'input, Endian> {
+        DebugLine { debug_line_section: EndianBuf(debug_line_section, PhantomData) }
+    }
+}
+
+/// Executes a `LineNumberProgram` to recreate and the mapping from instructions
```
tromey: This sentence could use a tweak.

tromey:
@fitzgen - I read through the rest a bit and I think everything looks pretty great.

fitzgen:
@tromey thanks for giving it a look over!

philipc:
Having the separate opcode iterator is good.

I switched the Entry/Attrs iterators to return `ParseResult<Option<...>>`. Probably should do the same for the iterators here.

It would be nice to avoid copying the registers for every line; it should be up to the caller to decide if a clone is needed. We could do that by storing the current `LineNumberRow` in `StateMachine`, and have `execute()` operate on the registers in that row, instead of having a separate copy. `StateMachine::next()` would return a reference to current `LineNumberRow`.  Would need to find some way to defer the reset of discriminator/basic_block/prologue_end/epilogue_begin though.

Both of those can be follow ups though. Looks good to me.

src/parser.rs
```diff
@@ -84,75 +94,40 @@ impl error::Error for Error {
...
 
-pub fn parse_u8(input: &[u8]) -> ParseResult<(&[u8], u8)> {
+/// Parse a `i8` from the input.
+#[doc(hidden)]
+pub fn parse_i8(input: &[u8]) -> ParseResult<(&[u8], i8)> {
     if input.len() == 0 {
         Err(Error::UnexpectedEof)
     } else {
-        Ok((&input[1..], input[0]))
+        Ok((&input[1..], unsafe { mem::transmute(input[0]) }))
```
philipc: Can you use `input[0] as i8` instead of `transmute`?

fitzgen: Ok I totally misunderstood `as` conversions/coercions -- I thought that `u8 as i8` would always result in either a positive `i8` or a panic if the `u8` was greater than `std::i8::MAX` but actually it does reinterpret the bits directly, just like this transmute intended to do.

PR #28, opened 2016-08-02
Change `EntriesCursor` to parse in `next_*` instead of `current`

philipc:
The main changes are:

- delete `current_ref`
- `current` now returns a reference to the cached entry, and never parses anything.
- add `next_entry`, which parses a single entry and caches it. The cached entry may be `None` after calling this if the entry was a null.
- `next_dfs` and `next_sibling` now parse the next entry and cache it.
- reaching the last child in `next_sibling` no longer exhausts the cursor. You can call `next_dfs` for example.
- for a new cursor, `current` has no meaning until you to call  `next_dfs` to parse the first entry

I think this is an improvement over the current syntax, but it's subjective, and not as much improvement as I was hoping. For example, you can iterate with:

    while let Some(depth) = try!(cursor.next_dfs()) {
        let entry = cursor.current().unwrap();
        ...
    }

This also gives a small performance improvement for `bench_parsing_debug_info` (order of 5%).

The main thing lacking is that I wanted this syntax instead:
`while let Some((entry, depth)) = try!(cursor.next_dfs()) { ... }`
but I had borrow checker problems trying to implement this... maybe need non-lexical borrows, not sure. I'll try another day. I'm not actually convinced this will help performance though.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2054,26 +2054,24 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
     /// Get a reference to the entry that the cursor is currently pointing to.
     pub fn current_ref<'me>
         (&'me mut self)
-         -> Option<ParseResult<&'me DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>>> {
+         -> ParseResult<Option<&'me DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>>> {
 
         // First, check for a cached result.
         {
             if let Some(ref cached) = self.cached_current {
-                return Some(Ok(cached));
+                return Ok(Some(cached));
             }
         }
 
         if self.input.len() == 0 {
-            return None;
+            return Ok(None);
         }
 
-        match parse_unsigned_leb(self.input) {
-            Err(e) => Some(Err(e)),
-
+        match try!(parse_unsigned_leb(self.input)) {
             // Null abbreviation is the lack of an entry.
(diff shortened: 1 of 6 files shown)


philipc:
I added some more commits to return the entry from `next_dfs` and `next_sibling`.

fitzgen:
Hey! Went on a short vacation, but I'm back now. Taking a look...

src/parser.rs
```diff
@@ -2200,78 +2217,61 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
...
                 }
 
-                self.cached_current = None;
-
-                if self.input.len() > 0 {
-                    Some(delta_depth)
-                } else {
-                    None
-                }
+                // Must have been a strange leb128 encoding of zero.
```
fitzgen: I don't understand when this can happen -- if there are multiple 0 digits with the LEB128 continuation bit set? Eg the LEB128 equivalent of "00000000000"? If this is indeed the case that this code is handling, then can you add a more explicit comment about that? If not, then what is it?

fitzgen:
Looks great, thanks! I'm going to go ahead and merge, but please clarify that comment in a follow up, as it isn't clear to potential readers (such as myself :P) what is going on there.

src/parser.rs
philipc: Yes, for example [0x80, 0x00]. I doubt this ever happens in practice, but we do need to handle this case somehow. We could return on error instead. Or we could remove the micro optimization and make this the normal case.

I've updated the comments in 2339ca1.

PR #29, opened 2016-08-06
AttrsIter::next() now returns ParseResult<Option<Attribute>>

### LINKED ISSUE #21: Handle errors in attribute iteration
`EntriesCursor::next_dfs` doesn't handle errors:
https://github.com/fitzgen/gimli/blob/8e4eb3ce54a966b5bd17b396ff279858d794b162/src/parser.rs#L3388-L3389

`EntriesCursor::next_sibling` has a related problem at: https://github.com/fitzgen/gimli/blob/8e4eb3ce54a966b5bd17b396ff279858d794b162/src/parser.rs#L3526

Part of the problem is that `AttrsIter::next` leaves the iterator in a state that won't continue to give errors: https://github.com/fitzgen/gimli/blob/8e4eb3ce54a966b5bd17b396ff279858d794b162/src/parser.rs#L3091. So if an error does occur, the next DIE will start being parsed at wherever the error occurred, with unpredictable results.

(Note: the above is all from code inspection, I haven't tested this happens yet, so I may be mistaken.)

It's easy to get this wrong when using `Option<Result<...>>`. We could switch to `Result<Option<...>>`, but that loses the ergonomics of `for` loops.  Some of the iterator functions can be retained using https://github.com/sfackler/rust-fallible-iterator.

philipc:
This makes error handling more reliable.

Fixes #21

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -57,10 +57,10 @@ fn bench_parsing_debug_info(b: &mut test::Bencher) {
                 .expect("Should parse abbreviations");
 
             let mut cursor = unit.entries(&abbrevs);
-
             while let Some((_, entry)) = cursor.next_dfs().expect("Should parse next dfs") {
-                for attr in entry.attrs() {
-                    test::black_box(attr.expect("Should parse entry's attribute"));
+                let mut attrs = entry.attrs();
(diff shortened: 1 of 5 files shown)


fitzgen:
Awesome!

PR #32, opened 2016-08-11
Implement support for .debug_str

khuey:
This implements basic support for .debug_str, with a lookup method to give you an ffi::CStr in exchange for a DebugStrOffset.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,10 +68,11 @@ pub use endianity::{Endianity, EndianBuf, LittleEndian, BigEndian};
 
 mod parser;
 pub use parser::{Error, ParseResult, Format};
-pub use parser::{DebugStrOffset, DebugLocOffset, DebugMacinfoOffset, UnitOffset};
+pub use parser::{DebugLocOffset, DebugMacinfoOffset, UnitOffset};
 pub use parser::{DebugInfo, DebugInfoOffset, UnitHeadersIter, UnitHeader};
 pub use parser::{DebugTypes, DebugTypesOffset, TypeUnitHeadersIter, TypeUnitHeader};
 pub use parser::{EntriesCursor, DebuggingInformationEntry, AttrsIter, Attribute, AttributeValue};
+pub use parser::{DebugStr, DebugStrOffset};
 
 mod abbrev;
 pub use abbrev::{DebugAbbrev, DebugAbbrevOffset, Abbreviations, Abbreviation,
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2859,3 +2859,51 @@ fn test_parse_type_unit_header_64_ok() {
         otherwise => panic!("Unexpected result: {:?}", otherwise),
     }
 }
+
+/// The `DebugStr` struct represents the DWARF strings
+/// found in the `.debug_str` section.
+#[derive(Debug, Clone, Copy)]
+pub struct DebugStr<'input, Endian>
+    where Endian: Endianity
+{
+    debug_str_section: EndianBuf<'input, Endian>,
+}
+
+impl<'input, Endian> DebugStr<'input, Endian>
+    where Endian: Endianity
+{
+    /// Construct a new `DebugStr` instance from the data in the `.debug_str`
+    /// section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_str` section and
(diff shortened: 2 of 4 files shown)


philipc:
Hi, thanks for the PR!

What the benefit of the `FromBytesUntilNull` trait? It's better to keep things simple unless there is a need.

src/lib.rs
```diff
@@ -68,10 +68,11 @@ pub use endianity::{Endianity, EndianBuf, LittleEndian, BigEndian};
 
 mod parser;
 pub use parser::{Error, ParseResult, Format};
-pub use parser::{DebugStrOffset, DebugLocOffset, DebugMacinfoOffset, UnitOffset};
+pub use parser::{DebugLocOffset, DebugMacinfoOffset, UnitOffset};
 pub use parser::{DebugInfo, DebugInfoOffset, UnitHeadersIter, UnitHeader};
 pub use parser::{DebugTypes, DebugTypesOffset, TypeUnitHeadersIter, TypeUnitHeader};
 pub use parser::{EntriesCursor, DebuggingInformationEntry, AttrsIter, Attribute, AttributeValue};
+pub use parser::{DebugStrings, DebugStrOffset};
```
philipc: We should use consistent names for `DebugStrings` and `DebugStrOffset`. I have a mild preference for `DebugStr` since that corresponds to the dwarf section name.

src/parser.rs
```diff
@@ -2875,3 +2875,54 @@ fn test_parse_type_unit_header_64_ok() {
...
+    ///
+    /// # let buf = [0x01, 0x02, 0x00];
+    /// # let offset = DebugStrOffset(0);
+    /// # let read_debug_str_section_somehow = || &buf;
+    /// # let debug_str_offset_somehow = || offset;
+    /// let debug_strings = DebugStrings::<LittleEndian>::new(read_debug_str_section_somehow());
+    /// println!("Found string {:?}", debug_strings.get_str(debug_str_offset_somehow()));
+    /// ```
+    pub fn get_str(&self, offset: DebugStrOffset) -> &ffi::CStr {
+        let result = ffi::CStr::from_bytes_until_nul(&self.debug_str_section[offset.0 as usize ..]);
```
philipc: It's important that the library does not panic for corrupt dwarf data. So you need to check that `offset.0` is in range and return a `ParseResult`.

khuey:
It's not strictly necessary.  I wrote it separately before noticing there was existing code that does the same sort of thing.  Does seem like a missing piece of the standard library though.

src/parser.rs
```diff
@@ -2875,3 +2875,54 @@ fn test_parse_type_unit_header_64_ok() {
...
+        match result {
+            Ok((cstr, _)) => cstr,
+            otherwise => panic!("Bad offset? {:?}", otherwise),
```
philipc: Similarly, don't panic here.

philipc:
Ok, I think it would be better to drop that commit and just use `parse_null_terminated_string` then.

khuey:
All comments addressed, I believe.  Not sure if there's a better way to write that kinda-ugly double-match in dwarfdump.rs ... this is the first rust I've ever written.

src/parser.rs
```diff
@@ -2859,3 +2859,54 @@ fn test_parse_type_unit_header_64_ok() {
...
+    /// # let debug_str_offset_somehow = || offset;
+    /// let debug_str = DebugStr::<LittleEndian>::new(read_debug_str_section_somehow());
+    /// println!("Found string {:?}", debug_str.get_str(debug_str_offset_somehow()));
+    /// ```
+    pub fn get_str(&self, offset: DebugStrOffset) -> ParseResult<&ffi::CStr> {
+        let result = parse_null_terminated_string(&self.debug_str_section[offset.0 as usize ..]);
+        match result {
+            Ok((_, cstr)) => Ok(cstr),
+            Err(e) => Err(e),
+        }
```
philipc: You should be able to do `result.map(|(_, cstr)| cstr)`

khuey: That's a neat trick.

fitzgen:
Thanks for the PR @khuey! I had been hoping to get to this soon, and then all of a sudden there's not one but two PRs for it :)

Thanks for doing the review @philipc!

philipc:
FYI I push 3 follow up commits at 5eccd6d.

PR #37, opened 2016-08-13
.debug_aranges

### LINKED ISSUE #10: Implement support for .debug_aranges
Section 6.1.2 of the DWARF 4 standard.

> For lookup by address, a table is maintained in a separate object file section called .debug_aranges. The table consists of sets of variable length entries, each set describing the portion of the program’s address space that is covered by a single compilation unit.

khuey:
Fixes #10.

Tested to produce reasonably sane output on the firefox binary.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -3,7 +3,7 @@
 extern crate gimli;
 extern crate test;
 
-use gimli::{DebugAbbrev, DebugInfo, DebugLine, DebugLineOffset, LineNumberProgramHeader,
+use gimli::{DebugAbbrev, DebugAranges, DebugInfo, DebugLine, DebugLineOffset, LineNumberProgramHeader,
             LittleEndian, StateMachine};
 
 use std::env;
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -68,6 +68,19 @@ fn bench_parsing_debug_info(b: &mut test::Bencher) {
     });
 }
 
+#[bench]
+fn bench_parsing_debug_aranges(b: &mut test::Bencher) {
+    let debug_aranges = read_section("debug_aranges");
+    let debug_aranges = DebugAranges::<LittleEndian>::new(&debug_aranges);
+
+    b.iter(|| {
+        let mut aranges = debug_aranges.aranges();
+        while let Some(arange) = aranges.next_arange().expect("Should parse arange OK") {
+            // Not really anything else we can check right now.
+        }
+    });
+}
+
 // We happen to know that there is a line number program and header at
 // offset 0 and that address size is 8 bytes. No need to parse DIEs to grab
 // this info off of the compilation units.
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -200,14 +200,12 @@ fn dump_line<Endian>(file: &obj::File, debug_abbrev: gimli::DebugAbbrev<Endian>)
 fn dump_aranges<Endian>(file: &obj::File)
     where Endian: gimli::Endianity
 {
-    let debug_aranges = obj::get_section(file, ".debug_aranges")
-        .expect("Does not have .debug_aranges section");
-    println!(".debug_aranges");
-    let debug_aranges = gimli::DebugAranges::<Endian>::new(debug_aranges);
+    if let Some(debug_aranges) = obj::get_section(file, ".debug_aranges") {
+        println!(".debug_aranges");
+        let debug_aranges = gimli::DebugAranges::<Endian>::new(debug_aranges);
 
-    let mut aranges = debug_aranges.aranges();
-    while let Some(arange) = aranges.next() {
-        for arange in arange.expect("Should parse arange OK") {
+        let mut aranges = debug_aranges.aranges();
+        while let Some(arange) = aranges.next_arange().expect("Should parse arange OK") {
             println!("arange starts at {}, length of {}, cu_die_offset = {:?}",
                      arange.start(), arange.len(), arange.debug_info_offset());
         }
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -30,6 +30,7 @@ fn dump_file<Endian>(file: obj::File)
     dump_info(&file, debug_abbrev, debug_str);
     dump_types(&file, debug_abbrev, debug_str);
     dump_line(&file, debug_abbrev);
+    dump_aranges::<Endian>(&file);
 }
 
 fn dump_info<Endian>(file: &obj::File, debug_abbrev: gimli::DebugAbbrev<Endian>,
@@ -192,6 +193,23 @@ fn dump_line<Endian>(file: &obj::File, debug_abbrev: gimli::DebugAbbrev<Endian>)
                     println!("  {}", opcode);
                 }
             }
+        }
+    }
+}
+
+fn dump_aranges<Endian>(file: &obj::File)
+    where Endian: gimli::Endianity
+{
+    let debug_aranges = obj::get_section(file, ".debug_aranges")
+        .expect("Does not have .debug_aranges section");
+    println!(".debug_aranges");
+    let debug_aranges = gimli::DebugAranges::<Endian>::new(debug_aranges);
+
+    let mut aranges = debug_aranges.aranges();
+    while let Some(arange) = aranges.next() {
+        for arange in arange.expect("Should parse arange OK") {
+            println!("arange starts at {}, length of {}, cu_die_offset = {:?}",
+                     arange.start(), arange.len(), arange.debug_info_offset());
         }
     }
 }
--- a/src/aranges.rs
+++ b/src/aranges.rs
@@ -1,0 +1,249 @@
+#![deny(missing_docs)]
+
+use endianity::{Endianity, EndianBuf};
+use parser::{parse_u16, parse_uN_as_u64, parse_unit_length, parse_debug_info_offset, parse_address_size, DebugInfoOffset, Format, ParseResult, Error};
+use std::cmp::{Ordering};
+use std::marker::PhantomData;
+use std::rc::Rc;
+
+/// The `DebugAranges` struct represents the DWARF address range information
+/// found in the `.debug_aranges` section.
+#[derive(Debug, Clone, Copy)]
+pub struct DebugAranges<'input, Endian>
+    where Endian: Endianity
+{
+    debug_aranges_section: EndianBuf<'input, Endian>,
+}
+
+impl<'input, Endian> DebugAranges<'input, Endian>
+    where Endian: Endianity
+{
+    /// Construct a new `DebugAranges` instance from the data in the `.debug_aranges`
+    /// section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_aranges` section and
+    /// present it as a `&[u8]` slice. That means using some ELF loader on
+    /// Linux, a Mach-O loader on OSX, etc.
+    ///
+    /// ```
+    /// use gimli::{DebugAranges, LittleEndian};
+    ///
+    /// # let buf = [];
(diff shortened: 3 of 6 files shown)


fitzgen:
Still reading the PR, but an initial thought:

Why should parse_address_size take an EndianBuf? The address size is a u8 so it doesn't matter what endianity the buffer is, and since that info isn't needed, I think it makes sense to leave the function in its current simpler form. It's pretty easy to do `my_endian_buf.0` or `my_endian_buf.into()` when passing in an `EndianBuf`'s `&[u8]` to `parse_address_size` and `let rest = EndianBuf::new(rest)` after parsing, if you need to rewrap the rest of the input back into an `EndianBuf` again.

khuey:
Mostly because I think it is a PITA to chain parses that way. ;)  I'll revert it if you like.

examples/dwarfdump.rs
```diff
@@ -196,6 +197,23 @@ fn dump_line<Endian>(file: &obj::File, debug_abbrev: gimli::DebugAbbrev<Endian>)
     }
 }
 
+fn dump_aranges<Endian>(file: &obj::File)
+    where Endian: gimli::Endianity
+{
+    let debug_aranges = obj::get_section(file, ".debug_aranges")
+        .expect("Does not have .debug_aranges section");
```
fitzgen: We don't require the `.debug_aranges` section to be able to dump the rest of the debug info (unlike say `.debug_abbrev`) so it makes more sense to use an `if let` rather than `expect("...")` or `unwrap()`.

```rust
if let Some(debug_aranges) = obj::get_section(file, ".debug_aranges") {
    // ...
}
```

fitzgen:
Still looking at the second commit, will look at how chaining interacts there.

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+            None
+        } else {
+            match Arange::parse(self.input) {
+                Ok((remaining_input, aranges)) => {
+                    self.input = remaining_input;
+                    Some(Ok(aranges))
+                }
+                Err(e) => {
+                    self.input = self.input.range_to(..0);
+                    Some(Err(e))
```
fitzgen: Fallible iterators are kind of hard to get right (see https://github.com/fitzgen/gimli/issues/21) so we've been switching from iterators that return `Option<ParseResult<T>>` to `next_whatever` methods that return `ParseResult<Option<T>>` which subtly encourage robustness in the face of errors. Basically, it is easy to keep iterating without handling an error, when as soon as we yeild an error we can't meaningfully yield anything else after that.

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+
+        let header_length = match format {
+            Format::Dwarf32 => 8,
+            Format::Dwarf64 => 12,
+        };
+        let dividing_line: usize = try!(length.checked_sub(header_length).ok_or(Error::BadLength)) as usize;
+
+        Ok((rest.range_from(dividing_line..),
+            rest.range_to(..dividing_line),
+            Rc::new(ArangeHeader { format: format,
```
fitzgen: Let's put the buffer of aranges to be parsed in this set into the header so that (a) it is easier to keep track of them since there's only one thing to juggle instead of two that need to be juggled together, and (b) it is consistent with all our usual parse signatures that return `(rest, parsed_val)`.

See `LineNumberProgramHeader` which does exactly this, if you want an example.

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+        let address_size = header.address_size;
+        let segment_size = header.segment_size; // May be zero!
+
+        let (rest, segment) = try!(parse_uN_as_u64(segment_size, input));
+        let (rest, offset) = try!(parse_uN_as_u64(address_size, rest));
+        let (rest, length) = try!(parse_uN_as_u64(address_size, rest));
+
+        Ok((rest, match (segment, offset, length) {
+            (0, 0, 0) => None,
+            _ => Some(Arange { segment: segment,
```
fitzgen: Nice.

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+    format: Format,
+    length: u64,
+    version: u16,
+    offset: DebugInfoOffset,
+    address_size: u8,
+    segment_size: u8,
+}
+
+#[derive(Debug, Clone, Eq)]
+pub struct Arange
```
fitzgen: Nit: let's name this `ArangeHeader` for consistency with the rest of the lib

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+impl PartialEq for Arange
+{
+    fn eq(&self, other: &Arange) -> bool
+    {
+        // The expected comparison, but verify that header matches if everything else does.
+        match (self.segment == other.segment,
+               self.offset == other.offset,
+               self.length == other.length) {
+            (true, true, true) => {
+                assert!(self.header == other.header);
```
fitzgen: Nit: `debug_assert!(...)`

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+    ///   let arange = parse_result.unwrap();
+    ///   println!("arange starts at {}, has length {}", arange.start(), arange.len());
+    /// }
+    /// ```
+    pub fn aranges(&self) -> ArangesIter<Endian> {
+        ArangesIter { input: self.debug_aranges_section }
+    }
+}
+
+pub struct ArangesIter<'input, Endian>
```
fitzgen: Nit: let's rename this `ArangesHeaderIter`

src/aranges.rs
```diff
@@ -0,0 +1,249 @@
...
+            _ => Some(Arange { segment: segment,
+                               offset: offset,
+                               length: length,
+                               header: header.clone() } ),
+        }))
+    }
+
+    /// Parse a single set of aranges, including a header and all the aranges specified. Returns
+    /// the remaining arange sets and the series of parsed aranges.
+    fn parse<Endian>(input: EndianBuf<Endian>) -> ParseResult<(EndianBuf<Endian>, Vec<Arange>)>
```
fitzgen: Rather than eagerly parsing all address ranges, let's expose an iterator[0] (can be named `ArangesIter` after the extant one is renamed) so that consumers can choose whether to heap allocate at all, and also can bail out early if they find the info they're looking for. Think of an `addr2line` or `atos` clone: it will take an address as input, iterate aranges until it finds the cu for the address, and then execute just enough of that cu's line number program to find the source/line for the input address. It doesn't want a whole vec of aranges, or the whole line number matrix, it just wants to parse/execute enough to get the one bit of info it needs and then stop.

[0] again, not an actual iterator that returns `Option<ParseResult<T>>` but an iterator-like that has a `next_arange` method that returns `ParseResult<Option<T>>`

fitzgen:
Ok, all done reviewing! Overall, this looks great, but just a bunch of little nits and things to make this more consistent with the rest of the lib and one note about forcing eager parsing and a heap allocation on consumers who potentially don't want either of those things.

Additionally, please extend `./tests/parse_self.rs` with a test that we can parse the Real World .debug_aranges section in `./fixtures/self/debug_aranges`. Same thing for `./benches/bench.rs` so we can track performance of changes to the parsing code.

I'm fine leaving the parse_address_size change in.

Thanks!

src/aranges.rs
khuey: Can't tell if serious or ...

Do you mean ArangeEntry or something similar?

src/aranges.rs
khuey: The reason I didn't do something like this originally is because there's nothing in the final parsed structs (`ArangeHeader`s and `Arange`s) that require keeping alive the original input buffer once parsing is complete.  Whereas making this change would require that.  Did you consider this tradeoff?

src/aranges.rs
khuey: Srsly.  I patted myself on the back for that one.

src/aranges.rs
khuey: Hmm.  If I were to implement `addr2line` (which I plan to!) and I were going to deal with more than a handful of addresses as inputs I would parse all of `.debug_aranges` into a tree of some sort and then run my queries against the tree.  I don't have a good feel for rust performance characteristics but a series of function calls to parse each and every 16 byte segment feels suboptimal.

khuey:
Can you clarify the `Option<ParseResult<T>>` vs `ParseResult<Option<T>>` thing a bit more?  Is the concern just that you are checking things in the "wrong" order in the first case?  i.e. it makes more sense to handle errors, then the end, then a value than to handle the end, then any errors, and then a value?

philipc:
@khuey The problem is that `Result` is defined to be [must_use](https://doc.rust-lang.org/src/core/up/src/libcore/result.rs.html#248) but `Option` is not, so if the `Result` is nested in the `Option`, it is too easy to inadvertently miss checking it. A real example of the problem this can cause is that calling `count()` on an iterator that encounters an error may either return a short count, or get stuck in an infinite loop, depending on the error. See http://sfackler.github.io/rust-fallible-iterator/doc/v0.1.2/fallible_iterator/ for a more detailed explanation.

src/aranges.rs
fitzgen: ... yes ...

Jumping around too fast.

src/aranges.rs
fitzgen: Indeed things change if you want to symbolicate many addresses. This is why I prefer the library leave these kinds of decisions about whether to stop in the middle after finding a specific address range or building up a query-able data structure to the consumer of the library.

I'm not too worried about the function call for each 16 byte segment: the `next_arange` method can be marked `#[inline]` and there shouldn't be much penalty.

On the other hand, if you want to come back with bench numbers, I'm happy to defer to them.

src/aranges.rs
fitzgen: Ah, usually this isn't a concern because we are returning references into the input left and right. However, given that for debug_aranges we don't do any of that, I think this is fine as is.

khuey:
Ah, yes, that makes perfect sense.

khuey:
I think I addressed or refuted all the comments.  Lmk if there are more changes you want to see, or if we're good.  I can squash commits tonight.

benches/bench.rs
```diff
@@ -68,6 +68,19 @@ fn bench_parsing_debug_info(b: &mut test::Bencher) {
...
 
+#[bench]
+fn bench_parsing_debug_aranges(b: &mut test::Bencher) {
+    let debug_aranges = read_section("debug_aranges");
+    let debug_aranges = DebugAranges::<LittleEndian>::new(&debug_aranges);
+
+    b.iter(|| {
+        let mut aranges = debug_aranges.aranges();
+        while let Some(arange) = aranges.next_arange().expect("Should parse arange OK") {
+            // Not really anything else we can check right now.
```
fitzgen: The preferred style is to prefix unused variables with `_` or just name them `_`, which tells the compiler to shut up about not using them.

I'll fix this in a follow up after merging.

tests/parse_self.rs
```diff
@@ -92,3 +92,14 @@ fn test_parse_self_debug_line() {
...
 }
+
+#[test]
+fn test_parse_self_debug_aranges() {
+    let debug_aranges = read_section("debug_aranges");
+    let debug_aranges = DebugAranges::<LittleEndian>::new(&debug_aranges);
+
+    let mut aranges = debug_aranges.aranges();
+    while let Some(_) = aranges.next_arange().expect("Should parse arange OK") {
+        // Not really anything else we can check right now.
```
fitzgen: Generally, we would do `test::black_box(arange)` to make sure that the whole thing didn't get optimized away. See https://doc.rust-lang.org/book/benchmark-tests.html for more info.

I'll merge and fix this in a follow up myself, though.

fitzgen:
Thanks, Kyle!

khuey:
Thanks for the review/merge!

PR #38, opened 2016-08-13
Print DIE offset in dwarfdump instead of abbreviation code

jvns:
here's an attempt at resolving https://github.com/fitzgen/gimli/issues/35! This would make the dwarfdump example match better.

I compared the output to what dwarfdump prints and the offsets look right to me. there is probably some terrible Rust in here because my Rust is actually awful :)

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -731,8 +731,12 @@ impl<'input, Endian> UnitHeader<'input, Endian>
         self.format
     }
 
+    fn header_size(&self) -> usize {
+        Self::size_of_header(self.format)
+    }
+
     fn is_valid_offset(&self, offset: UnitOffset) -> bool {
-        let size_of_header = Self::size_of_header(self.format);
+        let size_of_header = self.header_size();
         if (offset.0 as usize) < size_of_header {
             return false;
         }
@@ -960,6 +964,7 @@ pub struct DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>
     where 'input: 'unit,
           Endian: Endianity + 'unit
 {
+    offset: usize,
     attrs_slice: &'input [u8],
     after_attrs: Cell<Option<&'input [u8]>>,
     code: u64,
@@ -973,6 +978,11 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev, 
     /// Get this entry's code.
     pub fn code(&self) -> u64 {
         self.code
+    }
+
+    /// Get this entry's offset.
+    pub fn offset(&self) -> usize {
+        self.offset
     }
 
     /// Get this entry's `DW_TAG_whatever` tag.
@@ -1947,6 +1957,7 @@ fn test_attrs_iter() {
                0xaa, 0xaa];
 
     let entry = DebuggingInformationEntry {
+        offset: 0,
         attrs_slice: &buf,
         after_attrs: Cell::new(None),
         code: 1,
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -2123,11 +2135,19 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
         }
     }
 
+    /// Return the current offset in bytes inside the compilation unit
+    pub fn current_offset(&self) -> usize {
+        let ptr = self.input.as_ptr() as *const u8 as usize;
+        let start_ptr = self.unit.entries_buf.as_ptr() as *const u8 as usize;
+        ptr - start_ptr + self.unit.header_size()
```
khuey: Not sure how @fitzgen feels but you may want to do checked arithmetic here to handle bogus values in the DWARF without panicking.

src/parser.rs
```diff
@@ -2123,11 +2135,19 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
...
+    }
+
     /// Move the cursor to the next DIE in the tree.
     ///
     /// Returns `Some` if there is a next entry, even if this entry is null.
     /// If there is no next entry, then `None` is returned.
     pub fn next_entry(&mut self) -> ParseResult<Option<()>> {
+        let offset = self.current_offset();
```
philipc: Move this down to about line 2158. The offset needs to be calculated after we call `after_entry`. The problem is that we lazily parse the attributes for the previous entry, so we need to finish parsing them first. It doesn't affect the dwarfdump example because it displays every attribute, but that won't always be the case.

src/parser.rs
philipc: @khuey In general you are right, but in this case the values are based only on the pointers, not what they point to. The cursor will always point into the unit `entries_buf`, so any error would be a logic bug in gimli, and panicking is correct.

src/parser.rs
```diff
@@ -2123,11 +2135,19 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
         }
     }
 
+    /// Return the current offset in bytes inside the compilation unit
+    pub fn current_offset(&self) -> usize {
```
philipc: I don't think this should be public. As a user, I would expect this to be equivalent to `current().offset()`, but it isn't, because by the time the user can call this, `self.input` will already be pointing somewhere after the start of the current entry.

jvns: addressed your two comments! let me know if you prefer them to be squashed

src/parser.rs
```diff
@@ -2156,6 +2155,7 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
             return Ok(None);
         }
 
+        let offset = self.current_offset();
```
philipc: Sorry, I just realized I my previous comment wasn't complete. It also needs to calculate the offset using the value that `after_entry()` returns.

jvns: done!

philipc:
LGTM, thanks! I'm not sure if we've decided yet if we want commits squashed, @fitzgen ?

fitzgen:
> LGTM, thanks! I'm not sure if we've decided yet if we want commits squashed, @fitzgen ?

I think as long as they are nice, atomic commits where tests are passing on each one so bisect does not break, there is no need to squash. On the other hand, when the ducks are not in as nice of a row as that, we should squash. Overall, I don't think it is the biggest deal in the world.

jvns:
squashed :)

fitzgen:
Looks great! Thanks!

PR #39, opened 2016-08-14
Line fixes

### LINKED ISSUE #31: Investigate avoiding copying StateMachineRegisters when executing line number programs
From https://github.com/fitzgen/gimli/pull/27#issuecomment-238742174:

> It would be nice to avoid copying the registers for every line; it should be up to the caller to decide if a clone is needed. We could do that by storing the current LineNumberRow in StateMachine, and have execute() operate on the registers in that row, instead of having a separate copy. StateMachine::next() would return a reference to current LineNumberRow. Would need to find some way to defer the reset of discriminator/basic_block/prologue_end/epilogue_begin though.

I just looked into this a little and my conclusion is that we would need something like (wrapping?) `std::cell::Ref` but with a custom `Drop` impl to do the resetting.

Not going to dig any further at the moment.

philipc:
Fixes #31 

Not sure about a402366, happy to drop it if you prefer.

There's a few other fixes in the later patches too.  Some of them seem to have improved performance, but I didn't figure out which.  I've also been slack about adding tests for these... Can do it another day if you don't get to it first, not sure if you're currently working on this, so don't want to do too much that is going to conflict.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -95,8 +95,8 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
 
     fn adjust_opcode(&self, opcode: u8) -> i64 {
         let opcode = opcode as i64;
-        let line_base = self.row.header.line_base as i64;
-        opcode - line_base
+        let opcode_base = self.row.header.opcode_base as i64;
+        opcode - opcode_base
     }
 
     /// Section 6.2.5.1
--- a/src/line.rs
+++ b/src/line.rs
@@ -850,7 +850,7 @@ pub struct LineNumberProgramHeader<'input, Endian>
     /// standard opcodes. The first element of the array corresponds to the
     /// opcode whose value is 1, and the last element corresponds to the opcode
     /// whose value is `opcode_base - 1`."
-    standard_opcode_lengths: Vec<u64>,
+    standard_opcode_lengths: &'input [u8],
 
     /// > Entries in this sequence describe each path that was searched for
     /// > included source files in this compilation. (The paths include those
@@ -960,7 +960,7 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
     }
 
     /// The byte lengths of each standard opcode in this header's line program.
-    pub fn standard_opcode_lengths(&self) -> &[u64] {
+    pub fn standard_opcode_lengths(&self) -> &'input [u8] {
         &self.standard_opcode_lengths[..]
     }
 
@@ -1028,13 +1028,12 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
         let (rest, line_range) = try!(parser::parse_i8(rest));
         let (rest, opcode_base) = try!(parser::parse_u8(rest));
 
-        let mut rest = EndianBuf::<Endian>::new(rest);
-        let mut standard_opcode_lengths = Vec::with_capacity((opcode_base - 1) as usize);
-        for _ in 1..opcode_base {
-            let (rest1, opcode_length) = try!(parser::parse_unsigned_leb(rest.0));
-            rest = EndianBuf::new(rest1);
-            standard_opcode_lengths.push(opcode_length);
+        let standard_opcode_count = opcode_base as usize - 1;
+        if rest.len() < standard_opcode_count {
+            return Err(parser::Error::UnexpectedEof);
         }
+        let standard_opcode_lengths = &rest[..standard_opcode_count];
+        let mut rest = EndianBuf::<Endian>::new(&rest[standard_opcode_count..]);
 
         let mut include_directories = Vec::new();
         loop {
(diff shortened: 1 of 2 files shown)


fitzgen:
It looks like this isn't compiling yet?

```
error: no method named `push` found for type `&[u8]` in the current scope

    --> src/line.rs:1443:40

     |

1443 |         header.standard_opcode_lengths.push(0);

     |                                        ^^^^

error: no method named `push` found for type `&[u8]` in the current scope

    --> src/line.rs:1457:40

     |

1457 |         header.standard_opcode_lengths.push(1);

     |                                        ^^^^

error: no method named `push` found for type `&[u8]` in the current scope

    --> src/line.rs:1471:40

     |

1471 |         header.standard_opcode_lengths.push(3);

     |                                        ^^^^
```

fitzgen:
Ah, you must have made this change in parallel with my new parse tests. I think you'll need to update them.

fitzgen:
> I've also been slack about adding tests for these... Can do it another day if you don't get to it first, not sure if you're currently working on this, so don't want to do too much that is going to conflict.

Hold off on it -- I'm going to write (at least initial) tests for line number programs. In fact, I already have mostly complete unit tests for parsing, execution is next up.

In general, I think I should have written these tests before merging, but I got ahead of myself once I got the integration tests passing and the results seemed sane. I need to hold myself to a higher standard.

Going to give this PR a once over now...

src/line.rs
```diff
@@ -63,197 +62,154 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
...
-
-        self.registers.discriminator = 0;
-
-        row
+        let operation_advance = adjusted_opcode / self.row.header.line_range;
+        self.apply_operation_advance(operation_advance as u64);
     }
 
     /// Execute the given opcode, if a new row in the line number matrix is
     /// generated, return it.
```
fitzgen: This comment needs to be updated. Also, not sure whether `Option<()>` is the best return type. Not sure if `bool` is any more satisfying. Maybe a custom `enum`? Can't really think of a good name for one...

src/line.rs
```diff
@@ -277,14 +233,24 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
...
-    pub fn next_row(&mut self) -> parser::ParseResult<Option<LineNumberRow<'input, 'header, Endian>>> {
+    pub fn next_row(&mut self)
+                    -> parser::ParseResult<Option<&LineNumberRow<'input, 'header, Endian>>> {
+        if self.row.end_sequence {
+            self.row.reset();
+        } else {
+            self.row.discriminator = 0;
+            self.row.basic_block = false;
+            self.row.prologue_end = false;
+            self.row.epilogue_begin = false;
```
fitzgen: A comment here about Steps 3-7 from 6.5.2.1 would be nice. Is it valid to always apply this after yielding a row, not just for special opcodes?

fitzgen:
> Not sure about a402366, happy to drop it if you prefer.

I'm not excited about this change, because I liked how the code matched 1:1 with the standard, but I don't feel strongly about it and it is still pretty obvious how the code matches the standard.

/me shrugs

src/line.rs
philipc: Yeah I wasn't sure either. I think `bool` is slightly better, so I changed it to that. I also made it private while I was at it.

src/line.rs
philipc: There's 3 opcodes that can yield a new row: EndSequence, Special, and Copy.  The reset behaviour is the same for Special and Copy. I've added comments.

philipc:
> I'm not excited about this change, because I liked how the code matched 1:1 with the standard, but I don't feel strongly about it and it is still pretty obvious how the code matches the standard.

Dropped it then.

philipc:
> Ah, you must have made this change in parallel with my new parse tests. I think you'll need to update them.

Yeah I hadn't noticed them. Should be all fixed now.

src/line.rs
fitzgen: Thanks for double checking!

src/line.rs
```diff
@@ -277,14 +235,31 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
...
+        // Perform any reset that was required after copying the previous row.
+        if self.row.registers.end_sequence {
+            // Previous opcode was EndSequence, so reset everything
+            // as specified in Section 6.2.5.3.
+            self.row.registers.reset(self.row.header.default_is_stmt);
+        } else {
+            // Previous opcode was one of:
+            // - Special - specified in Section 6.2.5.1, steps 4-7
+            // - Copy - specified in Section 6.2.5.2
+            // The reset behaviour is the same in both cases.
```
fitzgen: Great comment, thanks!

fitzgen:
Merged in 5aed77e1a8f1f37623dc4522e521ecb5e832e252 -- thanks!

PR #42, opened 2016-08-15
doc: Fix example formatting for aranges

kamalmarhubi:


kamalmarhubi:
Decided to do the impl instead, see #43.

kamalmarhubi:
Apparently the impl wasn't included intentionally so I fixed the example instead.

fitzgen:
Looks great, thanks!

PR #45, opened 2016-08-15
debug_str: Extend lifetime of returned &CStr to match input

kamalmarhubi:


fitzgen:
Thanks!

khuey:
For my benefit, could someone explain what this change actually does? :)

fitzgen:
On phone, so terse explanation: explicit lifetime elision will choose the
input lifetime (&self) for the output lifetime, but in this case we know
that it is valid as long as the underlying input buffer is valid. In
practice, likely very similar lifetimes but not necessarily.

I don't know all the details of lifetime elision so someone correct me if
that's not technically correct...

khuey:
Ah, interesting.

kamalmarhubi:
Ah yeah I should have explained better in the commit message. @fitzgen has it right: the selected lifetime is that of `&self`. In my program I have a hashmap with keys from the name attribute. Sometimes these are `AttributeValue::String`, in which case they have the input lifetime; in others they are `AttributeValue::DebugStrRef`, in which case they need to be looked up in the `DebugStr`. These were coming out with the wrong lifetime, which got in the way of doing what I wanted. :-)

kamalmarhubi:
(The aim of this was to eliminate copies in https://github.com/jvns/ruby-stacktrace)

kamalmarhubi:
Looking a bit more closely, the reason the lifetime can end up wrong is that `DebugStr` is `Copy`, so it's possible to pass it by copy instead of a reference. That messes it all up, and can make the lifetime of the `DebugStr` really really short.

fitzgen:
Interesting... I've kind of wondered whether we should be deriving Copy for
a lot of these things, because it seems like it is easy to make copies
unknowingly... Will file an issue.


_Nick_

kamalmarhubi:
Yeah I think they probably shouldn't for that reason.

kamalmarhubi:
For linkiness, the relevant issue is #48

PR #46, opened 2016-08-16
Convert the starts-at-1 file register into an array index before using it to index an array.

khuey:
"There are two hard problems in computer science: cache invalidation, naming things, and off-by-one errors."

Easy fix.  Came up while testing my basic addr2line clone.


fitzgen:
You make a great fuzzer ;)

PR #47, opened 2016-08-16
Implement addr2line

khuey:
This definitely isn't ready for merging yet, but it might be ready to look at.

Notable todos:

- Should we factor out object loading support for the examples/ somewhere?
- Should parsing DW_AT_stmt_list be handled inside gimli somewhere? (probably)
- Get the directory name instead of just printing the index

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ version = "0.6.0"
 byteorder = "0.5.3"
 leb128 = "0.2.1"
 
+[dev-dependencies]
+getopts = "0.2"
+
 [target.'cfg(target_os="linux")'.dev-dependencies]
 elf = "0.0.9"
 
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -1,0 +1,217 @@
+extern crate gimli;
+extern crate getopts;
+
+use std::env;
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let mut opts = getopts::Options::new();
+    opts.optopt("e", "exe", "Set the input file name (default is a.out)", "<executable>");
+
+    let matches = opts.parse(&args[1..]).unwrap();
+    let file_path = matches.opt_str("e").unwrap_or("a.out".to_string());
+    let file = obj::open(&file_path);
+    if obj::is_little_endian(&file) {
+        symbolicate::<gimli::LittleEndian>(&file, &matches);
+    } else {
+        symbolicate::<gimli::BigEndian>(&file, &matches);
+    }
+}
+
+fn parse_uint_from_hex_string(string: &str) -> u64
+{
+    if string.len() > 2 && string.starts_with("0x") {
+        u64::from_str_radix(&string[2..], 16).expect("Failed to parse address")
+    } else {
+        u64::from_str_radix(string, 16).expect("Failed to parse address")
+    }
+}
+
+fn entry_offsets_for_addresses<Endian>(file: &obj::File, addrs: &Vec<u64>)
+                                       -> Vec<Option<gimli::DebugInfoOffset>>
+    where Endian: gimli::Endianity
+{
+    let aranges = obj::get_section(file, ".debug_aranges").expect("Can't addr2line with no aranges");
+    let aranges = gimli::DebugAranges::<Endian>::new(aranges);
+    let mut aranges = aranges.aranges();
+
+    let mut dies: Vec<Option<gimli::DebugInfoOffset>> = (0..addrs.len()).map(|_| None).collect();
+    while let Some(arange) = aranges.next_arange().expect("Should parse arange OK") {
+        let start = arange.start();
+        let end = start + arange.len();
+
+        for (i, addr) in addrs.iter().enumerate() {
+            if *addr >= start && *addr < end {
+                dies[i] = Some(arange.debug_info_offset());
+            }
+        }
+    }
+
+    dies
+}
+
+fn line_offset_for_entry<Endian>(abbrevs: &gimli::DebugAbbrev<Endian>, header: &gimli::UnitHeader<Endian>)
+                                 -> Option<gimli::DebugLineOffset>
+    where Endian: gimli::Endianity
+{
+    let abbrev = abbrevs.abbreviations(header.debug_abbrev_offset()).expect("Fail");
+    let mut entries = header.entries(&abbrev);
+    let (_, entry) = entries.next_dfs()
+        .expect("Should parse first entry OK")
+        .expect("And first entry should exist!");
+    match entry.attr_value(gimli::DW_AT_stmt_list) {
+        Some(gimli::AttributeValue::DebugLineRef(offset)) => Some(offset),
+        _ => None,
+    }
+}
+
+fn display_file<Endian>(header: &gimli::LineNumberProgramHeader<Endian>, row: &gimli::LineNumberRow<Endian>)
+    where Endian: gimli::Endianity
+{
+    let file = row.file().unwrap();
+    let directory_index = file.directory_index();
+    if directory_index > 0 {
+        println!("{}/{}:{}",
+                 header.include_directories()[directory_index as usize - 1].to_string_lossy(),
+                 file.path_name().to_string_lossy(),
+                 row.line().unwrap());
+    } else {
+        println!("{}:{}", file.path_name().to_string_lossy(), row.line().unwrap());
+    }
+}
+
+fn symbolicate<Endian>(file: &obj::File, matches: &getopts::Matches)
+    where Endian: gimli::Endianity
+{
+    let addrs: Vec<u64> = matches.free.iter().map(|x| parse_uint_from_hex_string(x)).collect();
+
+    let offsets = entry_offsets_for_addresses::<Endian>(&file, &addrs);
+    let debug_info = obj::get_section(file, ".debug_info")
+        .expect("Can't addr2line without .debug_info");
+    let debug_info = gimli::DebugInfo::<Endian>::new(debug_info);
+    let debug_abbrev = obj::get_section(&file, ".debug_abbrev")
+        .expect("Can't addr2line without .debug_abbrev");
+    let debug_abbrev = gimli::DebugAbbrev::<Endian>::new(debug_abbrev);
+    let debug_line = obj::get_section(file, ".debug_line")
+        .expect("Can't addr2line without .debug_line");
+    let debug_line = gimli::DebugLine::<Endian>::new(&debug_line);
+
+    for (info_offset, addr) in offsets.iter().zip(addrs.iter()) {
+        match *info_offset {
+            None => println!("Found nothing"),
+            Some(d) => match debug_info.header_from_offset(d) {
+                Err(_) => println!("Couldn't get DIE header"),
+                Ok(h) => {
+                    let line_offset = line_offset_for_entry(&debug_abbrev, &h)
+                        .expect("No offset into .debug_lines!?");
+                    let header =
+                        gimli::LineNumberProgramHeader::new(debug_line, line_offset, h.address_size());
+                    if let Ok(header) = header {
+                        let mut state_machine = gimli::StateMachine::new(&header);
+                        match state_machine.run_to_address(addr) {
+                            Err(_) => println!("Failed to run line number program!"),
+                            Ok(None) => println!("Failed to find matching line for {}", *addr),
+                            Ok(Some(row)) =>  display_file(&header, &row),
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+// All cross platform / object file format compatibility stuff should
+// be contained in the `obj` module. Each supported platform / object
+// file format should implement the `obj` module with an identical
+// interface, but with the `pub type File` changing as needed. Hooray
+// duck typing!
+
+#[cfg(target_os="linux")]
+mod obj {
+    extern crate elf;
+    use std::path::Path;
+
+    /// The parsed object file type.
+    pub type File = elf::File;
+
+    /// Open and parse the object file at the given path.
+    pub fn open<P>(path: P) -> File
+        where P: AsRef<Path>
+    {
+        let path = path.as_ref();
+        elf::File::open_path(path).expect("Could not open file")
+    }
+
+    /// Get the contents of the section named `section_name`, if such
+    /// a section exists.
+    pub fn get_section<'a>(file: &'a File, section_name: &str) -> Option<&'a [u8]> {
+        file.sections
+            .iter()
+            .find(|s| s.shdr.name == section_name)
+            .map(|s| &s.data[..])
+    }
+
+    /// Return true if the file is little endian, false if it is big endian.
+    pub fn is_little_endian(file: &File) -> bool {
+        match file.ehdr.data {
+            elf::types::ELFDATA2LSB => true,
+            elf::types::ELFDATA2MSB => false,
+            otherwise => panic!("Unknown endianity: {}", otherwise),
+        }
+    }
+}
+
+#[cfg(target_os="macos")]
+mod obj {
+    extern crate mach_o;
+
+    use std::ffi::CString;
+    use std::fs;
+    use std::io::Read;
+    use std::mem;
+    use std::path::Path;
+
+    pub type File = Vec<u8>;
+
+    pub fn open<P>(path: P) -> File
+        where P: AsRef<Path>
+    {
+        let mut file = fs::File::open(path).expect("Could not open file");
+        let mut buf = Vec::new();
+        file.read_to_end(&mut buf).expect("Could not read file");
+        buf
+    }
+
+    // Translate the "." prefix to the "__" prefix used by OSX/Mach-O, eg
+    // ".debug_info" to "__debug_info".
+    fn translate_section_name(section_name: &str) -> CString {
+        let mut name = Vec::with_capacity(section_name.len() + 1);
+        name.push(b'_');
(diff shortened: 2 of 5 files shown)


fitzgen:
> Should we factor out object loading support for the examples/ somewhere?

That would be great. I don't know if we can add modules to `examples/` or not, but I feel like the obj loading doesn't belong in the library proper (especially as it is currently implemented).

Maybe we should factor it out into a new crate and do it Right (unless some other crate abstracting obj files already exists that seems good enough).

> Should parsing DW_AT_stmt_list be handled inside gimli somewhere? (probably)

Yes, it is on my TODO list to make a method on `UnitHeader` that takes the `DebugLine` and returns the `LineNumberProgramHeader` (probably as an `Option` because it might not exist and I'm also unsure whether partial units have them). First, I need to write more unit tests for line number program execution, however. Feel free to make that method yourself, if you want, as you'll probably be able to beat me to the punch.

Cargo.toml
```diff
@@ -15,6 +15,7 @@ version = "0.6.0"
 [dependencies]
 byteorder = "0.5.3"
 leb128 = "0.2.1"
+getopts = "0.2"
```
fitzgen: This should be under `[dev-dependencies]`, which are only included for tests and examples.

http://doc.crates.io/specifying-dependencies.html#development-dependencies

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+    let mut opts = getopts::Options::new();
+    opts.optopt("e", "exe", "Set the input file name (default is a.out)", "<executable>");
+
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => m,
+        Err(f) => { panic!(f.to_string()); }
+    };
```
fitzgen: This is pretty much equivalent to:

```rust
let matches = opts.parse(&args[1..]).unwrap();
```

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+    }
+}
+
+// All cross platform / object file format compatibility stuff should
+// be contained in the `obj` module. Each supported platform / object
+// file format should implement the `obj` module with an identical
+// interface, but with the `pub type File` changing as needed. Hooray
+// duck typing!
```
fitzgen: "hooray"... yeah we need to figure something out here

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => m,
+        Err(f) => { panic!(f.to_string()); }
+    };
+
+    let file_path = match matches.opt_str("e") {
+        Some(f) => f,
+        None => "a.out".to_string()
+    };
```
fitzgen:
```rust
let file_path = matches.opt_str("e").unwrap_or("a.out".to_string());
```

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+        do_work::<gimli::LittleEndian>(&file, &matches);
+    } else {
+        do_work::<gimli::BigEndian>(&file, &matches);
+    }
+}
+
+fn do_work<Endian>(file: &obj::File, matches: &getopts::Matches)
+    where Endian: gimli::Endianity
+{
+    let addrs: Vec<u64> = matches.free.iter().map(|x| u64::from_str_radix(x,16).expect("Failed to parse")).collect();
```
fitzgen: In `add2line -e foo bar baz bang`, is `matches.free` equal to the vec of `bar`, `baz`, and `bang`?

Might be worth allowing a `0x` prefix on these addresses -- not sure if `u64::from_str_radix` will allow them.

Super nit: I think "Failed to parse address" is a bit more helpful of a message.

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let aranges = obj::get_section(file, ".debug_aranges").expect("Can't addr2line with no aranges");
+    let aranges = gimli::DebugAranges::<Endian>::new(aranges);
+    let mut aranges = aranges.aranges();
+
+    let mut dies: Vec<Option<gimli::DebugInfoOffset>> = Vec::new();
+    dies.resize(addrs.len(), None); // XXXkhuey better way to do this?
```
fitzgen:
```rust
let mut dies = Vec::with_capacity(addrs.len());
```

fitzgen: Ah wait, here we go:

```rust
let mut dies = (0..addrs.len()).map(|_| None);
```

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+    while let Some(arange) = aranges.next_arange().expect("Should parse arange OK") {
+        let start = arange.start();
+        let end = start + arange.len();
+
+        for (i, addr) in addrs.iter().enumerate() {
+            match (*addr >= start, *addr < end) {
+                (true, true) => {
+                    dies[i] = Some(arange.debug_info_offset());
+                },
+                _ => continue,
```
fitzgen: I think a plain ol' `if` is nicer here:

```rust
if *addr >= start && *addr < end {
    dies[i] = ...;
}
```

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+    let debug_info = obj::get_section(file, ".debug_info").expect("Can't addr2line with no debug_info");
+    let debug_info = gimli::DebugInfo::<Endian>::new(debug_info);
+    let debug_abbrev = obj::get_section(&file, ".debug_abbrev")
+        .expect("Does not have .debug_abbrev section");
+    let debug_abbrev = gimli::DebugAbbrev::<Endian>::new(debug_abbrev);
+    let debug_line = obj::get_section(file, ".debug_line").expect("Can't addr2line with no debug_line");
+    let debug_line = gimli::DebugLine::<Endian>::new(&debug_line);
+
+    for (index, die) in dies.iter().enumerate() {
+        let addr = addrs[index];
```
fitzgen: If you want:

```rust
for (die, addr) in dies.iter().zip(addrs.iter()) {
    ...
}
```

examples/addr2line.rs
```diff
@@ -0,0 +1,198 @@
...
+        match *die {
+            None => println!("Found nothing"),
+            Some(d) => match debug_info.header_from_offset(d) {
+                Err(_) => continue,
+                Ok(h) => {
```
fitzgen: I would flatten this match statement so there isn't so much rightward drift:

```rust
match *die {
    None => ...,
    Some(Err(_)) => ...,
    Some(Ok(h)) => ...,
}

Also, not sure how to best handle errors here. Doing the same as when we get back `None` and printing a message seems like an improvement over silently swallowing.

File: examples/addr2line.rs [hunk 10]
```diff
@@ -0,0 +1,198 @@
...
+
+    let file = obj::open(&file_path);
+    if obj::is_little_endian(&file) {
+        do_work::<gimli::LittleEndian>(&file, &matches);
+    } else {
+        do_work::<gimli::BigEndian>(&file, &matches);
+    }
+}
+
+fn do_work<Endian>(file: &obj::File, matches: &getopts::Matches)
```
fitzgen: 1. This function is yuge, let's deconstruct it into its pieces

2. I'm not a fan of the name -- I think something like `symbolicate` is a bit better

Cargo.toml
khuey: Done.

examples/addr2line.rs
khuey: Done.

examples/addr2line.rs
khuey: Done.

examples/addr2line.rs
khuey: Yes.

We need some sort of parsing function.  Ideally one that could handle at a minimum decimal numbers and hex with a 0x prefix.  The built in from_str methods don't :/

examples/addr2line.rs
khuey: Nice!

examples/addr2line.rs
khuey: Done.

examples/addr2line.rs
khuey: Nice.

examples/addr2line.rs
khuey: What would this actually look like?  The Err/Ok aren't inside the Some here.

examples/addr2line.rs
khuey: Yeah, that's a good idea.  I moved one big piece out.  I tried to move out executing the state machine but ran into rust-lang/rust#21906

khuey:
This is ready for another pass, @fitzgen.

examples/addr2line.rs
fitzgen: Ah, I missed that. You could do

```rust
let dies = dies.map(|opt_d| opt_d.map(|d| debug_info.header_from_offset(d)));
```

before the iteration. Whether or not that is an improvement... /me shrugs

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+
+fn parse_uint_from_hex_string(string: &str) -> u64
+{
+    if string.len() > 2 {
+        if string.starts_with("0x") {
+            return u64::from_str_radix(&string[2..], 16).expect("Failed to parse address");
+        }
+    }
+
+    return u64::from_str_radix(string, 16).expect("Failed to parse address");
```
fitzgen: Nit: rust style is to leave out `return` and the semicolon when in tail position.

It's not just functions either. Rust's block "statements" will evaluate to their last expression, if you drop the semicolon, eg:

```rust
let foo = {
    bar();
    1
};
assert_eq(foo, 1);
```

Same for `if`s, etc.

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+fn entries_for_addresses<Endian>(file: &obj::File, addrs: &Vec<u64>) -> Vec<Option<gimli::DebugInfoOffset>>
```
fitzgen: Nit: rename to `offsets_for_addresses` and use `offsets` instead of `dies` below. That would match the types and is less of a surprise to me, the reader.

fitzgen: Or I guess `cu_offsets_for_addresses`

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+    dies
+}
+
+fn offset_for_entry<Endian>(abbrevs: &gimli::DebugAbbrev<Endian>, header: &gimli::UnitHeader<Endian>)
+                            -> Option<gimli::DebugLineOffset>
+    where Endian: gimli::Endianity
+{
+    let abbrev = abbrevs.abbreviations(header.debug_abbrev_offset()).expect("Fail");
+    let mut entries = header.entries(&abbrev);
+    let (_, entry) = entries.next_dfs().expect("Fail").unwrap();
```
fitzgen: Nit:

```rust
let (_, entry) = entries.next_dfs()
    .expect("Should parse first entry OK")
    .expect("And first entry should exist");
```

The whole point of using `expect("...")` rather than `unwrap()` is to get better error messages when things go wrong, so we should write messages that give us some context or idea of what went wrong if/when things fail.

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+    match entry.attr_value(gimli::DW_AT_stmt_list) {
+        Some(gimli::AttributeValue::DebugLineRef(offset)) => Some(offset),
+        _ => None,
+    }
+}
+
+fn display_file<Endian>(header: &gimli::LineNumberProgramHeader<Endian>, row: &gimli::LineNumberRow<Endian>)
+    where Endian: gimli::Endianity
+{
+    let file = row.file().expect("WTF");
```
fitzgen: Ditto. If you can't think of useful error messages, then just use `unwrap()` instead of `expect("...")`.

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+    println!("{}/{}:{}",
+             header.include_directories()[file.directory_index() as usize - 1].to_string_lossy(),
+             file.path_name().to_string_lossy(),
```
fitzgen: I was meaning to add a nicer API in the future for getting a `FileEntry`'s directory that would encapsulate this indexing and the potential off by ones. If you want to add it now, go ahead, if not I can still do it later and update this code at that time.

fitzgen: Imagining something with a signature like:

```rust
impl FileEntry {
    pub fn directory(&self, header: &LineNumberProgramHeader) -> &'input ffi::CStr;
}
```

Although we need to handle the `index == 0` case, which is supposed to mean the compilation's current directory, which we get from the CU DIE... Maybe it should take that as input, too...

----------

Anyways, fine if you don't want to deal with all that now, but it did just make me catch that you're not handling the case where the `directory_index == 0`, so you should at least stub that out right now such that it won't panic.

examples/addr2line.rs
```diff
@@ -0,0 +1,208 @@
...
+                Err(_) => println!("Couldn't get DIE header"),
+                Ok(h) => {
+                    let offset = offset_for_entry(&debug_abbrev, &h).expect("No offset into .debug_lines!?");
+                    let header =
+                        gimli::LineNumberProgramHeader::new(debug_line, offset, h.address_size());
+                    if let Ok(header) = header {
+                        let mut state_machine = gimli::StateMachine::new(&header);
+                        match state_machine.run_to_address(addr) {
+                            Err(_) => println!("Failed to run line number program!"),
+                            Ok(None) => println!("Failed to find matching line for {}", *addr),
```
fitzgen: Thanks!

src/line.rs
```diff
@@ -267,6 +267,26 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
...
+
+    /// Parse and execute opcodes until we reach a row matching `addr`, the end of the program,
+    /// or an error.
+    pub fn run_to_address(&mut self, addr: &u64)
+                          -> parser::ParseResult<Option<&LineNumberRow<'input, 'header, Endian>>> {
+        loop {
+            match self.next_row() {
+                Ok(Some(row)) => {
+                    if row.address() == *addr {
+                        // Can't return 'row' directly here because of rust-lang/rust#21906.
```
fitzgen: Yep. Non-lexical borrows will help a lot when they come, but in the meantime you'll get used to restructuring your programs to avoid these kinds of issues / play better with the borrow checker, and only need these weird control flows when absolutely necessary. I don't really have anything concrete to say, just an intuition that all the higher order functions and iterator combinators and things help with this stuff.

src/line.rs
```diff
@@ -267,6 +267,26 @@ impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
...
+                        break;
+                    }
+                },
+                Ok(None) => return Ok(None),
+                Err(err) => return Err(err),
+            };
+        }
+
+        return Ok(Some(&self.row));
```
fitzgen: Nit: drop the `return` and `;` here, too.

src/parser.rs
```diff
@@ -1243,17 +1260,52 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev,
...
+    pub fn attr_value_raw(&self, name: constants::DwAt) -> Option<AttributeValue<'input>> {
+        self.attr_value_internal(name, |_, x| x)
+    }
+
+    /// Find the first attribute in this entry which has the given name,
+    /// and return its value, after running `prettify_attr_value` on it.
+    /// Returns `Ok(None)` if no attribute is found.
+    pub fn attr_value(&self, name: constants::DwAt) -> Option<AttributeValue<'input>> {
+        self.attr_value_internal(name,
+                                 DebuggingInformationEntry::<'input, 'abbrev, 'unit, Endian>::prettify_attr_value)
```
fitzgen: Nit: instead of making the `_internal` variant take a callback, do

```rust
self.attr_value_raw(...).map(|val| self.prettify_attr_value(name, val));
```

and then you can get rid of the `_internal` variant completely by folding it into the `_raw` version.

tests/parse_self.rs
```diff
@@ -71,17 +70,7 @@ fn test_parse_self_debug_line() {
...
-            // DW_FORM_sec_offset, so instead we get DW_FORM_data4
-            // values. Because of this, we have to turn the data into an offset
-            // manually.
-            let offset = match value {
-                AttributeValue::Data(data) => LittleEndian::read_u32(data),
-                otherwise => panic!("unexpected value form: {:?}", otherwise),
-            };
-            let offset = DebugLineOffset(offset as u64);
-
+        if let Some(AttributeValue::DebugLineRef(offset)) = unit_entry.attr_value(DW_AT_stmt_list) {
```
fitzgen: Nice.

fitzgen:
Ok, I left a bunch of comments, but they are all very nit picky. Consider it an r=me with comments. I'll merge as soon you update.

Thanks!

examples/addr2line.rs
khuey: Yeah, I know this, but it's a hard impulse to break ;)  Thanks for the reminder.

khuey:
If github isn't sending you emails when I overwrite commits, this is ready to merge @fitzgen.

fitzgen:
Huh, I'm not getting those emails.

fitzgen:
Thanks again!

khuey:
np, I'm learning a lot from your reviews :)

PR #49, opened 2016-08-16
Make StateMachine::new consume its LineNumberProgramHeader

### LINKED ISSUE #40: LineNumberProgramHeader can't reliably iterate twice
`LineNumberProgramHeader::file_names` has interior mutability, which is incompatible with multiple `StateMachine`s using it.

fitzgen:
Because executing a line number program can define new file entries, it is
incorrect to pass the same header to multiple `StateMachine`s, which is possible
when the `StateMachine` doesn't consume the header.

Fixes #40

@philipc, mind taking a look?

Unfortunately, this ended up making `OpcodesIter::next_opcode` take a `&header` param instead of storing the reference in the `OpcodesIter` struct itself, but I'm not sure how we could work around that.

This doesn't have any effect on the benches.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -1,7 +1,6 @@
 use constants;
 use endianity::{Endianity, EndianBuf};
 use parser;
-use std::cell::{Ref, RefCell};
 use std::ffi;
 use std::fmt;
 use std::marker::PhantomData;
@@ -48,28 +47,29 @@ impl<'input, Endian> DebugLine<'input, Endian>
 /// to expand the byte-coded instruction stream into a matrix of line number
 /// information." -- Section 6.2.1
 #[derive(Debug)]
-pub struct StateMachine<'input, 'header, Endian>
-    where Endian: 'header + Endianity,
-          'input: 'header
+pub struct StateMachine<'input, Endian>
+    where Endian: Endianity
 {
-    row: LineNumberRow<'input, 'header, Endian>,
-    opcodes: OpcodesIter<'header, 'input, Endian>,
+    header: LineNumberProgramHeader<'input, Endian>,
+    registers: StateMachineRegisters,
+    opcodes: OpcodesIter<'input, Endian>,
 }
 
-impl<'input, 'header, Endian> StateMachine<'input, 'header, Endian>
+impl<'input, Endian> StateMachine<'input, Endian>
     where Endian: Endianity
 {
     /// Construct a new `StateMachine` for executing line programs and
     /// generating the line information matrix.
-    pub fn new(header: &'header LineNumberProgramHeader<'input, Endian>) -> Self {
+    pub fn new(header: LineNumberProgramHeader<'input, Endian>) -> Self {
         let mut registers = StateMachineRegisters::default();
         registers.reset(header.default_is_stmt());
         let opcodes = OpcodesIter {
-            header: header,
             input: header.program_buf.0,
+            endian: PhantomData,
         };
         StateMachine {
-            row: LineNumberRow::new(header, registers),
+            header: header,
+            registers: registers,
             opcodes: opcodes,
         }
     }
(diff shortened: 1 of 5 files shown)


philipc:
Storing a copy of the header in `LineNumberRow` isn't good. If the user does want to clone the row then that'll be expensive. I think it will be better to store the header in `StateMachine`. The problem with that is `LineNumberRow` can no longer contain a reference to the header (because that would be a self-ref in the `StateMachine`). But `LineNumberRow` doesn't actually need the entire header, just one filename, so we can change `LineNumberRow` to only have a `&'input [u8]` for the filename that it needs. Not sure if this will fix the `OpcodesIter` problem too (probably not).

You should also be able to remove the `RefCell` usage.

philipc:
Actually the row needs a `FileEntry`, so storing that isn't as nice.

fitzgen:
@philipc ok, I moved the header into `StateMachine`, removed the `RefCell` usage, and I think things got a lot nicer in general. One small hitch is that `LineNumberRow::file` needs to take the header as input now, but I think that's a fine balance (probably need to add a `StateMachine::header` getter, too, so users can call this method).

fitzgen:
> (probably need to add a StateMachine::header getter, too, so users can call this method)

Done.

fitzgen:
Oh, and this is probably going to conflict with #47 but I can wait and rebase on top of that.

philipc:
I like how this is shaping up.

Maybe `LineNumberRow::file` should take the `StateMachine` as input instead of the header so that the user can do `row.file(&state_machine)` instead of `row.file(state_machine.header())`, and also to disallow the incorrect `row.file(&header)`.

fitzgen:
> Maybe LineNumberRow::file should take the StateMachine as input instead of the header so that the user can do row.file(&state_machine) instead of row.file(state_machine.header()), and also to disallow the incorrect row.file(&header).

I don't think it is really any more difficult to pass the incorrect `StateMachine` than the incorrect `header`. I toyed with adding phantom data for `'input` and `Endian` to LineNumberRow, but that only stops the user from using headers from completely different object files, not different headers in the same object file which seems much more likely an error. The phantom lifetime also artificially restricts against keeping them in the heap indefinitely.

Do you still think passing the state machine is the way to go?

philipc:
> I toyed with adding phantom data for 'input

The phantom data for a lifetime will still only be a bound, not an exact matching lifetime, right?

> Do you still think passing the state machine is the way to go?

I don't think it would be bad to do, but I don't feel strongly about it.

philipc:
I looked at what changes would be needed in addr2line for this, and it's not clean. Because `next_row()` is returning a reference, it keeps a mutable borrow for the state machine, so you can't do anything else (like call `state_machine.header()`).

It seems to me that either `next_row()` shouldn't be returning a reference, or the row should be self contained (so you don't need to call additional functions on the state machine while processing it). The former requires copying the row, the latter requires copying the file entry into the row. Or maybe I'm just not good enough at working with the borrow checker yet.

fitzgen:
Ok after a bit of juggling I have a design that plays well with the borrow checker! I'm really happy with how this came out.

@philipc want to take a final look?

src/line.rs
```diff
@@ -1160,6 +1169,23 @@ impl<'input> FileEntry<'input> {
         self.directory_index
     }
 
+    /// Get this file's directory.
+    ///
+    /// If this file's directory index is 0, meaning that it was found in
+    /// the current directory of compilation, return `None`.
+    pub fn directory<Endian>(&self,
+                             header: &LineNumberProgramHeader<'input, Endian>)
```
fitzgen: I'd like this API to take a UnitHeader or something and attempt to parse out the current directory, too, but I think this can be done in a follow up.

philipc:
I think this is good.  I suspect this won't be able to implement `FallibleIterator` now though, due to the lifetime on `next_row`. Providing helper APIs like `run_to_address` can address that though.

fitzgen:
> I think this is good. I suspect this won't be able to implement FallibleIterator now though, due to the lifetime on next_row. Providing helper APIs like run_to_address can address that though.

Yeah, that's too bad, but I think it is the right trade off.

PR #54, opened 2016-08-19
Add parse_address

tromey:
For expression parsing I needed to parse an address based on the DWARF
format.  I saw that this was done in several other spots as well, so I
consolidated these into a convenience function.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -1023,10 +1023,7 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
             return Err(parser::Error::UnknownVersion);
         }
 
-        let (rest, header_length) = try!(match format {
-            parser::Format::Dwarf32 => parser::parse_u32_as_u64(rest),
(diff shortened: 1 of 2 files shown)


khuey:
These get used for offsets too, perhaps just `parse_word`?

philipc:
I think it makes sense to have a common function for this. However, I haven't looked at the specifics of expressions, but the DWARF format is generally only used to determine the size of offsets within a section, not target addresses (which should use `unit.address_size()` instead).

tromey:
> However, I haven't looked at the specifics of expressions, but the DWARF format is generally only used to determine the size of offsets within a section, not target addresses

The particular case is `DW_OP_call_ref`, which takes a section offset.

tromey:
> These get used for offsets too, perhaps just parse_word?
Ok.

fitzgen:
This is a great clean up! Thanks!

PR #55, opened 2016-08-19
Instead of copy-pasting platform specific obj modules, use the object crate

fitzgen:
The object crate is identical, but we don't have copy-paste all over the place
anymore.

Mostly just testing that travis ci still works with the migration to the gimli-rs github org...


fitzgen:
Travis is still working \o/

PR #56, opened 2016-08-20
Add DebuggingInformationEntry:: has_children()

philipc:
Useful in conjunction with `EntriesCursor::next_entry()`.

I thought that ruby-stacktrace's usage of gimli was more complicated than it should be. The best alternative I could come up with was to drop down to using `next_entry()`, but I need `has_children()` to do that. See https://github.com/philipc/ruby-stacktrace/commit/038d997f22e17e2bc59eea3adf6f17cecd4fef50. `next_dfs()` is great for dwarfdump clones, but not sure it makes as much sense if you really want a tree structure. Using `next_entry()` isn't exactly pretty either though.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1189,6 +1189,11 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev, 
         self.abbrev.tag()
     }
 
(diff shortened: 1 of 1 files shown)


fitzgen:
Nice!

PR #57, opened 2016-08-20
Tell rustfmt not to make backup files.

khuey:


fitzgen:
Thanks, those were getting a little annoying.

PR #58, opened 2016-08-21
Add Attribute::normalize()

### LINKED ISSUE #51: Run prettify_attr_value on values returned from Attribute::value()
This is trickier than it should be because `Attribute` is `Endian`-less.  We should probably fix that, many of the subtypes of `AttributeValue` need an `Endianity` to be interpreted correctly.

philipc:
Fixes #51 

I didn't change the meaning of `Attribute::value()` because I thought that should keep matching the value field.

Two questions regarding follow ups to this:
- sections 7.8 to section 7.18 define constants for various attributes. I think it would make sense for `normalize` to return these, but that would mean adding `AttributeValue` variants for them. Maybe `normalize` should return a different enum instead? I don't feel strongly either way.
- currently dwarfdump converts `AttributeValue::DebugStrRef` to `AttributeValue::String`. I don't think that should be done in `normalize` but I'm not sure (it would require passing the debug_str section in).

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1322,45 +1322,29 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev, 
         }
     }
 
-    /// Run some common fixups to present DWARF attributes in more useful forms.
-    fn prettify_attr_value(&self,
-                           name: constants::DwAt,
-                           value: AttributeValue<'input, Endian>)
-                           -> AttributeValue<'input, Endian> {
-        match name {
-            constants::DW_AT_stmt_list => {
-                let offset = DebugLineOffset(match value {
-                    AttributeValue::Data(data) if data.len() == 4 => {
-                        Endian::read_u32(data.into()) as u64
-                    }
-                    AttributeValue::Data(data) if data.len() == 8 => Endian::read_u64(data.into()),
-                    AttributeValue::SecOffset(offset) => offset,
-                    otherwise => return otherwise,
-                });
-                AttributeValue::DebugLineRef(offset)
-            }
-            _ => value,
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


fitzgen:
> sections 7.8 to section 7.18 define constants for various attributes. I think it would make sense for normalize to return these, but that would mean adding AttributeValue variants for them. Maybe normalize should return a different enum instead? I don't feel strongly either way.

I'm weakly leaning towards the former.

> currently dwarfdump converts AttributeValue::DebugStrRef to AttributeValue::String. I don't think that should be done in normalize but I'm not sure (it would require passing the debug_str section in).

I agree -- let it be separate.

src/parser.rs
```diff
@@ -1421,62 +1410,101 @@ pub enum AttributeValue<'input> {
...
         self.value
     }
+
+    /// Get this attribute's normalized value.
+    ///
+    /// Attribute values can potentially be encoded in multiple equivalent forms,
+    /// and may have special meaning depending on the attribute name.  This method
+    /// converts the attribute value to a normalized form based on the attribute
+    /// name.
+    pub fn normalize(&self) -> AttributeValue<'input, Endian> {
```
fitzgen: Perhaps getting nitpicky, but: `normalize` makes me think of an effectful operation to normalize the contained value so that whenever the value is accessed in the future, it is the normalized version. I think `normalized` would be more intuitive for what is actually happening here.

However, I think the default we present should be the cleaned up / pretty / normalized value. We should be handling DWARF's incidental complexity instead of forcing it on our users. In that light, how about renaming what is currently `value` to `raw_value` and what is currently `normalize` to `value`?

fitzgen:
LGTM! Feel free to merge now and push the `value` /  `raw_value` stuff as a follow up, or fix it up locally and merge after that.

PR #60, opened 2016-08-22
Change AttributeValue::Addr to u64

philipc:
I think this is the form that the user will always want. The alternative is to do this conversion in Attribute::value(), but that requires adding another enum variant, and I thought that would be confusing (what to call it?).


fitzgen:
Yep, this is the right way to go IMO.

PR #62, opened 2016-08-23
Dw expr

tromey:
Here's the PR.  DO NOT MERGE THIS, as discussed, since it's missing at least a zillion tests.  Speaking of, I realize now I didn't add tests for the second patch.


src/op.rs
```diff
@@ -0,0 +1,1288 @@
...
+             parse_signed_leb, parse_word, parse_address};
+use endianity::{Endianity, EndianBuf};
+use std::marker::PhantomData;
+#[cfg(test)]
+use endianity::LittleEndian;
+
+#[derive(Debug, Clone, Copy)]
+/// A reference to a DIE, either relative to the current CU or
+/// relative to the section.
+pub enum DIEReference {
```
fitzgen: Nit: `DieReference`

src/op.rs
```diff
@@ -0,0 +1,1288 @@
...
+    pub size_in_bits: Option<u64>,
+    /// If given, the bit offset of the piece.  If `None`, then the
+    /// piece starts at the next byte boundary.
+    pub bit_offset: Option<u64>,
+    /// Where this piece is to be found.
+    pub location: Location<'input>,
+}
+
+// A helper function to handle branch offsets.
+fn compute_pc(pc: &[u8], bytecode: &[u8], offset: i16) -> ParseResult<usize> {
```
fitzgen: It seems slightly strange that the pc comes in as a slice and the new pc is returned as an offset. Would ergonomics be hurt if it took and returned both as a slice or both as offsets? Is there some other consideration here that trumps consistency?

tromey: I think it would be fine.  I'll make this change.

src/op.rs
```diff
@@ -0,0 +1,1288 @@
...
+    /// `DW_OP_push_object_address`
+    pub fn set_object_address(&mut self, value: u64) {
+        self.object_address = Some(value);
+    }
+
+    /// Set the maximum number of iterations to be allowed by the
+    /// expression evaluator.  The default is `None`.  This value can
+    /// be set to avoid denial of service attacks by bad DWARF
+    /// bytecode.
+    pub fn set_max_iterations(&mut self, value: u32) {
```
fitzgen: Nice.

src/op.rs
tromey: I've fixed this.

src/op.rs
```diff
@@ -0,0 +1,1288 @@
...
+            iteration += 1;
+            if let Some(max_iterations) = self.max_iterations {
+                if iteration > max_iterations {
+                    return Err(Error::TooManyIterations);
+                }
+            }
+
+            let mut terminated = false;
+            let mut piece_end = false;
+            let (newpc, operation) = try!(decode(pc, bytecode, self.address_size, self.format));
```
fitzgen: So in a loop or recursive function or any code that ends up getting executed more than once, we end up decoding the same bytecodes repeatedly?

I guess it is possible to jump between bytecodes and operands and have that get interpreted as valid expressions as well?

I think it would be better if we only decode the operations once and then just jump between decoded operations with a map from byte offset -> decoded opcode index.

Even better would be [separating syntactic analysis from execution](https://mitpress.mit.edu/sicp/full-text/sicp/book/node83.html) by "compiling" operations into lambdas so we skip the "which opcode are we interpreting right now?" check on each iteration (jimb calls this "lambda chaining", but I haven't heard that term anywhere else). Maybe rust's discriminated unions already give us most of the benefit of this, compared to scheme where we would need to repeatedly grab + check operands...

I think we can safely do this stuff as follow ups (unless you think the former is easy enough to land with this initial PR, the latter is far in follow up territory IMO).

tromey: > I guess it is possible to jump between bytecodes and operands and have that get interpreted as valid expressions as well?

Yeah, nothing in DWARF prevents this.  This approach could still be done by simply having one operation per byte, with all element initially set to `Nop`.

The reason I didn't is that it's unusual for DWARF expressions to have loops.  It does happen but it is reasonably hard to provoke: https://github.com/gcc-mirror/gcc/blob/master/gcc/dwarf2out.c#L12918 (there are a couple other cases in there),

Anyway, the benefit to caching the decoded form seemed small; but on the other hand it incurs an expense since we'd have to allocate, and since it would mean decoding operations that potentially would not be executed at all.

If we expect to evaluate the same expression multiple times, then even with the above it makes sense to cache.  In this case we'd want to give the caller some control though.

I think it's important that there be some decoding step that emits just a simple enum that can be inspected.  This enables other uses of the decoder as mentioned in the bug.

fitzgen:
Overall, types and everything seem solid to me.

We should really unit test each individual operation. Some kind of helper that you just give `(input_stack, input_pc, bytecode, expected_stack, expected_pc)` and it does all the work would be nice.

src/op.rs
fitzgen: Ok, ignore me for now. I need to think more deeply about how LUL style compilation of unwinding opcodes can work with gimli.

src/op.rs
```diff
@@ -0,0 +1,1288 @@
...
+/// A DWARF expression evaluation.
+pub struct Evaluation<'input, Endian>
+    where Endian: Endianity
+{
+    bytecode: &'input [u8],
+    address_size: u8,
+    format: Format,
+    initial_value: Option<u64>,
+    object_address: Option<u64>,
+    callbacks: &'input mut EvaluationContext<'input>,
```
fitzgen: This will automatically shorten `'input` to the bounds of the `EvaluationContext`'s lifetime. For maximum flexibility, its best to separate the lifetime of the reference from the generic lifetime parameter out like this:

```rust
pub struct Evaluation<'context, 'input, Endian>
    where Endian: 'context + Endianity,
          'input: 'context
{
    ...
}
```

I don't think this is really problematic in practice in this case because we aren't *returning* any string or slice with the `'input` lifetime and obviously `Evaluation` is bounded by both the reference to the context and the generic lifetime parameter. However, if we were returning a string or slice, we would definitely not want to accidentally shorten its lifetime, or it would be a pain for our users (we did this once already).

All that said, I think we should make this change even if it isn't problematic in this case, just for consistency with the other places where we have this kind of situation.

tromey:
> We should really unit test each individual operation

Yeah, I'm working on that.  I'm done testing the decoding operation now, and I'm about to start on the evaluator.

I think it would be good for gimli to use coveralls to provide coverage reports.  I don't see why it couldn't have 100% test coverage.

fitzgen:
> I think it would be good for gimli to use coveralls to provide coverage reports. I don't see why it couldn't have 100% test coverage.

We did for a little bit (when "we" was just me) but it just stopped running... No idea why, but I got the feeling it was on their end, since I didn't change anything. I'll file an issue.

fitzgen:
> I'll file an issue.

https://github.com/gimli-rs/gimli/issues/69

tromey:
Sorry this is taking so long.  I was away a bit.  Anyway you can watch my progress, mostly on tests but also fixing bugs, here if you care: https://github.com/tromey/gimli/tree/dw-expr-in-progress.  There's maybe a dozen opcodes still left to test, plus some corner cases.  Not including stack overflow checks, though I'm not sure I should bother with those.

fitzgen:
No problem!

tromey:
I wrote lots of tests and I believe I addressed all the review comments.  Testing found some bugs, all fixed now.  I think it's ready for an in-depth review.

fitzgen:
LGTM! Thanks @tromey! I just merged something which also added new error variants, so this will need a tiny rebase again -- sorry!

@philipc do you want to review this too before we merge it in?

src/op.rs
```diff
@@ -0,0 +1,2691 @@
...
+            let valbytes: &[u8] = val.into();
+            assert_eq!(valbytes.len(), bytecode.len() - 4);
+        }
+        otherwise => panic!("Unexpected result: {:?}", otherwise),
+    }
+}
+
+// Like parse_u8 but takes and returns an EndianBuf for convenience.
+fn parse_u8e<'input, Endian>(bytes: EndianBuf<'input, Endian>)
+                             -> ParseResult<(EndianBuf<'input, Endian>, u8)>
```
philipc: These parse functions should probably be with the similar functions in parser.rs.  We still need to split up parser.rs a bit though, I'm fine with this for now.

src/op.rs
```diff
@@ -0,0 +1,2691 @@
...
+
+/// Decode a single DWARF expression operation.
+///
+/// This is useful when examining a DWARF expression for reasons other
+/// than direct evaluation.
+///
+/// `bytes` points to a the operation to decode.  It should point into
+/// the same array as `bytecode`, which should be the entire
+/// expression.
+pub fn decode_expression_op<'input, Endian>
```
philipc: I prefer the parse functions to be static members of the struct that they parse. eg `Operation::parse`. I don't feel strongly about that, but we should be consistent.

src/op.rs
```diff
@@ -0,0 +1,2691 @@
...
+    /// expression evaluator.  The default is `None`.  This value can
+    /// be set to avoid denial of service attacks by bad DWARF
+    /// bytecode.
+    pub fn set_max_iterations(&mut self, value: u32) {
+        self.max_iterations = Some(value);
+    }
+
+    /// Evaluate a DWARF expression.
+    pub fn evaluate(&mut self) -> ParseResult<Vec<Piece<'input>>>
+        where Endian: Endianity
```
philipc: This function has complicated control flow. Generally that means it's better to split it up into a few functions. In this case though, that would be difficult due to the amount of state needed. Perhaps the state could be moved into a helper struct? Just something to consider, feel free to ignore/defer.

philipc:
I haven't gone through in fine detail, but it looks good to me.

src/op.rs
tromey: FWIW I considered your suggestion to move these into EndianBuf, but that seemed pretty invasive and so I put it off.  But maybe I should revisit this?

philipc: Not as part of this PR, and I'm still not sure it's a good idea, so that can be considered further once this is merged.

src/op.rs
tromey: I made this change.

tromey:
I think this latest version addresses all the review comments.  I split up the evaluator, but I'm not sure if it is cleaner or not; please comment either way, I kept the old version around as well so it's no trouble to revert if needed.

fitzgen:
\o/

Thanks @tromey !

philipc:
>  I split up the evaluator, but I'm not sure if it is cleaner or not; please comment either way

Thanks. I think this is moving in the right direction. It's still a bit complicated, but I'll need to spend more time understanding the standard before I can meaningfully comment further. So I think this is good as long as you're happy with it.

PR #65, opened 2016-08-23
Fix arange padding

### LINKED ISSUE #64: ArangeParser doesn't calculate first entry offset correctly
The standard says it has to be aligned to a multiple of the tuple size. This accounts for the weirdness in `parse_uN_as_u64`.

philipc:
Fixes #64 

Also cleaned up the unit_length handling for CUs and TUs.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -204,33 +204,6 @@ pub fn parse_u32_as_u64<Endian>(input: EndianBuf<Endian>) -> ParseResult<(Endian
     }
 }
 
-/// Parse a variable length sequence and return it as a `u64`. Currently only supports lengths of
-/// 0, 1, 2, 4, and 8 bytes.
-#[doc(hidden)]
-#[inline]
-#[allow(non_snake_case)]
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -3062,10 +3011,11 @@ fn parse_type_unit_header<Endian>(input: EndianBuf<Endian>)
...
     where Endian: Endianity
 {
-    let (rest, header) = try!(parse_unit_header(input));
-    let (rest, signature) = try!(parse_type_signature(rest));
+    let (after_unit, mut header) = try!(parse_unit_header(input));
+    let (rest, signature) = try!(parse_type_signature(header.entries_buf));
     let (rest, offset) = try!(parse_type_offset(rest, header.format()));
-    Ok((rest, TypeUnitHeader::new(header, signature, offset)))
+    header.entries_buf = rest;
+    Ok((after_unit, TypeUnitHeader::new(header, signature, offset)))
```
fitzgen: This is way nicer now with the split between the stuff after the unit and the rest of the unit. Thanks!

PR #68, opened 2016-08-25
Exec line number program tests

fitzgen:
Maybe a third of the way through writing these tests, but figured we could land this batch right now.

Care to rubber stamp this, @philipc?

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -672,13 +672,13 @@ impl<'statemachine, 'input, Endian> LineNumberRow<'statemachine, 'input, Endian>
     }
 
     /// The source file corresponding to the current machine instruction.
-    pub fn file(&self) -> Option<&'statemachine FileEntry<'input>> {
+    pub fn file(&self) -> parser::ParseResult<&'statemachine FileEntry<'input>> {
         // NB: registers.file starts counting at 1.
-        let file_idx = self.registers.file as usize - 1;
-        if self.header.file_names.len() > file_idx {
-            Some(&self.header.file_names[file_idx])
+        let file = self.registers.file as usize;
+        if 0 < file && file <= self.header.file_names.len() {
+            Ok(&self.header.file_names[file - 1])
         } else {
-            None
+            Err(parser::Error::BadFileIndex)
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/line.rs
```diff
@@ -672,13 +672,13 @@ impl<'statemachine, 'input, Endian> LineNumberRow<'statemachine, 'input, Endian>
...
     /// The source file corresponding to the current machine instruction.
-    pub fn file(&self) -> Option<&'statemachine FileEntry<'input>> {
+    pub fn file(&self) -> parser::ParseResult<&'statemachine FileEntry<'input>> {
         // NB: registers.file starts counting at 1.
-        let file_idx = self.registers.file as usize - 1;
-        if self.header.file_names.len() > file_idx {
-            Some(&self.header.file_names[file_idx])
+        let file = self.registers.file as usize;
+        if 0 < file && file <= self.header.file_names.len() {
+            Ok(&self.header.file_names[file - 1])
```
fitzgen: One of the tests caught an underflow here when `self.registers.file == 0`.

src/line.rs
```diff
@@ -1186,6 +1186,7 @@ impl<'input> FileEntry<'input> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use super::StateMachineRegisters;
```
fitzgen: This explicit import is necessary because the type is not `pub`.

src/line.rs
```diff
@@ -1619,4 +1633,235 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_exec_special_negative_line_advance() {
+        let header = make_test_header(&[]);
+
+        let mut initial_registers = new_registers();
+        initial_registers.line = 10;
+
+        let opcode = Opcode::Special(13);
```
fitzgen: All of these special opcode tests are using the example matrix from the end of section 6.2.5.1.

src/line.rs
```diff
@@ -1619,4 +1633,235 @@ mod tests {
...
+
+        let mut initial_registers = new_registers();
+        initial_registers.line = 2;
+
+        // -3 line advance.
+        let opcode = Opcode::Special(13);
+
+        let mut expected_registers = initial_registers.clone();
+        // Clamp at 0. No idea if this is the best way to handle this situation
+        // or not...
```
philipc: I expect most other DWARF implementations will wrap. I don't really care what we do until we encounter a case that needs some specific behavior to work.

philipc:
:heavy_check_mark: LGTM

fitzgen:
Thanks!

PR #73, opened 2016-08-27
Make everything a FallibleIterator

### LINKED ISSUE #44: Make iteration pattern consistent
Currently there are two patterns of iteration used across the library:

- implementing the `Iterator` trait with `Item=ParseResult` (eg, [`DebugInfo::units`](http://fitzgen.github.io/gimli/gimli/struct.DebugInfo.html#method.units) / [`UnitHeadersIter`](http://fitzgen.github.io/gimli/gimli/struct.UnitHeadersIter.html), and [`DebugTypes::units`](http://fitzgen.github.io/gimli/gimli/struct.DebugTypes.html#method.units) / [`TypeUnitHeadersIter`](http://fitzgen.github.io/gimli/gimli/struct.TypeUnitHeadersIter.html)
- having a `BlahIter` type that does __not__ implement `Iterator`, and instead has a `next` method that returns a `ParseResult<Option<Blah>>`.

From https://github.com/fitzgen/gimli/pull/37#issuecomment-239650372 it seems that the preferred iteration method is the second one.

I'm mostly interested in consistency, but will say that while I get the argument in the linked comment, I still prefer using `Iterator<Item=ParseResult<Blah>>`. In addition to the more ergonomic `for _ in` syntax, it also gets you the ability to do things like:

```
let units: Vec<_> = try!(debug_info.units().collect());
```

which is much more straightforward than

```
let aranges = vec![];
let mut iter = debug_aranges.aranges();
while let Some(arange) = try!(iter.next_arange()) {
    aranges.push(arange);
}
```

The  potential for infinite loops on error can be fixed by having a `done` flag in the iterator struct, which gets set on the first error, or on successful completion of iteration. This is done in the tar crate, for example: https://github.com/alexcrichton/tar-rs/blob/master/src/archive.rs#L442.

fitzgen:
* Change `fn next_foo(&mut self) -> ParseResult<Option<T>>` to
  `fn next(...) -> ...`.

* Add a `FallibleIterator` implementation for that method. By still having a
  non-trait-impl `next` method we don't force API consumers to
  `use FallibleIterator;` unless they want the helper methods.

* Change existing `Iterator<Item=ParseResult<T>>` implementations over to
  `FallibleIterator<Item=T>` implementations.

Fixes #44.

Look good @philipc ?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,13 +51,60 @@
 //! }
 //! ```
 //!
-//! See the
-//! [`examples/dwarfdump.rs`](https://github.com/gimli-rs/gimli/blob/master/examples/dwarfdump.rs)
-//! program for a complete example program.
+//! See the [example
+//! programs](https://github.com/gimli-rs/gimli/blob/master/examples/) for
+//! complete examples.
+//!
+//! ## Using with `FallibleIterator`
+//!
+//! The standard library's `Iterator` trait and related APIs do not play well
+//! with iterators where the `next` operation is fallible. One can make the
+//! `Iterator`'s associated `Item` type be a `Result<T, E>`, however the
+//! provided methods cannot gracefully handle the case when an `Err` is
+//! returned.
+//!
+//! This situation led to the
+//! [`fallible-iterator`](https://crates.io/crates/fallible-iterator) crate's
+//! existence. You can read more of the rationale for its existence in its
+//! docs. The crate provides the helpers you have come to expect (eg `map`,
+//! `filter`, etc) for iterators that can fail.
+//!
+//! `gimli`'s many lazy parsing iterators are a perfect match for the
+//! `fallible-iterator` crate's `FallibleIterator` trait because parsing is not
+//! done eagerly. Parse errors later in the input might only be discovered after
+//! having iterated through many items.
+//!
+//! To use `gimli` iterators with `FallibleIterator`, import the crate and trait
+//! into your code:
+//!
+//! ```
+//! // Add the `fallinle-iterator` crate. Don't forget to add it to your
+//! // `Cargo.toml`, too!
+//! extern crate fallible_iterator;
+//! extern crate gimli;
+//!
+//! // Use the `FallibleIterator` trait so its methods are in scope!
+//! use fallible_iterator::FallibleIterator;
+//! use gimli::{DebugAranges, LittleEndian, ParseResult};
+//!
+//! fn find_sum_of_address_range_lengths(aranges: DebugAranges<LittleEndian>)
+//!     -> ParseResult<u64>
+//! {
+//!     // `DebugAranges::items` returns a `FallibleIterator`!
+//!     aranges.items()
+//!         // `map` is provided by `FallibleIterator`!
+//!         .map(|arange| arange.len())
+//!         // `fold` is provided by `FallibleIterator`!
+//!         .fold(0, |sum, len| sum + len)
+//! }
+//!
+//! # fn main() {}
+//! ```
 
 #![deny(missing_docs)]
 
 extern crate byteorder;
+extern crate fallible_iterator;
 extern crate leb128;
 
 mod constants;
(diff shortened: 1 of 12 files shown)


src/lib.rs
```diff
@@ -51,13 +51,60 @@
 //! }
 //! ```
 //!
-//! See the
-//! [`examples/dwarfdump.rs`](https://github.com/gimli-rs/gimli/blob/master/examples/dwarfdump.rs)
-//! program for a complete example program.
+//! See the [example
+//! programs](https://github.com/gimli-rs/gimli/blob/master/examples/) for a
```
philipc: Nit: delete 'a'

src/lib.rs
```diff
@@ -51,13 +51,60 @@
...
+//! into your code:
+//!
+//! ```
+//! // Add the `fallinle-iterator` crate. Don't forget to add it to your
+//! // `Cargo.toml`, too!
+//! extern crate fallible_iterator;
+//! extern crate gimli;
+//!
+//! // Use the `FallibleIterator` trait so its methods are in scope!
+//! use fallible_iterator::FallibleIterator;
```
philipc: Not sure, but maybe it is better if we re-export types used in gimli's public API. I'm entirely clear on how cargo handles multiple versions of the same crate. So then the user can do `use gimli::FallibleIterator`. I don't think there is any downside to doing this, because I think they can still add the extern crate if they want to.

src/lib.rs
```diff
@@ -84,6 +131,7 @@ mod line;
 pub use line::*;
 
 mod lookup;
+pub use lookup::{DebugLookup, LookupEntryIter};
```
philipc: Can we avoid exposing these? They seem like an implementation detail.

src/parser.rs
```diff
@@ -353,37 +357,46 @@ impl<'input, Endian> DebugInfo<'input, Endian>
...
-    type Item = ParseResult<UnitHeader<'input, Endian>>;
-
-    fn next(&mut self) -> Option<Self::Item> {
+    /// Advance the iterator to the next unit header.
+    pub fn next(&mut self) -> ParseResult<Option<UnitHeader<'input, Endian>>> {
         if self.input.is_empty() {
-            None
+            Ok(None)
         } else {
             match parse_unit_header(self.input) {
```
philipc: I don't feel strongly about it, but we could change this to `try!(parse_unit_header(self.input)`. That would have the effect that we don't update `self.input` on error though.

philipc:
> Add a FallibleIterator implementation for that method. By still having a
non-trait-impl next method we don't force API consumers to
use FallibleIterator; unless they want the helper methods.

Good idea.

src/lib.rs
fitzgen: I asked about exactly this in https://github.com/sfackler/rust-fallible-iterator/issues/1, and the recommendation was that since the whole point of `FallibleIterator` is to be public to allow cross-crate usage, it *is* a part of the public API and should be public.

Re-exporting was mentioned as an option, but I think it would be an annoyance when combining `FallibleIterator`s across crates with `zip`, for example.

src/lib.rs
fitzgen: Yeah. I added them because it felt weird not having them when browsing the rustdocs, but then I also fixed up the docs for the type aliases to be better, so I don't think its necessary.

src/parser.rs
fitzgen: I think having the `self.input` reset is nice. I tried rewriting this in terms of `map` and `map_err` and it was clear that it bought nothing over `match`.

fitzgen:
Ok, fixed up the parts other than re-exporting and the `try!`. I am happy to change the `FallibleIterator` thing to a re-export in a follow up, if you feel strongly about it (we should probably do this with `ByteOrder`, at least or even drop that dependency entirely).

philipc:
> we should probably do this with ByteOrder

Users would only need `ByteOrder` if they wanted to do their own implementation of `Endianity`, but that is unlikely.

PR #74, opened 2016-08-28
Use `cache: cargo` in .travis.yml to speed up CI runs

fitzgen:
Doing a PR just to check that this does indeed work...


fitzgen:
I'm seeing an item for uploading the cached directory, so it seems to work.

PR #75, opened 2016-08-28
Add the workaround[0] to keep coveralls.io working until travis-cargo…

fitzgen:
… fixes the root issue

[0] https://github.com/huonw/travis-cargo/issues/58#issuecomment-229273382

Testing the fix with PR...


fitzgen:
Sweet :)

PR #76, opened 2016-08-28
.debug_aranges fixes

philipc:
- fix handling of zero tuples
- change our dwarfdump clone to closer match the real dwarfdump

I'm not 100% sure that the zero tuple handling is correct yet. Both dwarfdump and binutils-readelf display these zero tuples. Both llvm-readelf and elfutils-readelf stop reading at the zero tuple. I haven't been able to find anything that actually uses `.debug_aranges` though, so I can't see how clients of the libraries expect to be able to handle them. As far as I can tell, not even the real addr2line uses `.debug_aranges`, although it's a bit hard to follow the code so I'm not sure.

### DIFF
--- a/src/aranges.rs
+++ b/src/aranges.rs
@@ -99,7 +99,7 @@ pub struct ArangeParser<'input, Endian>
 }
 
 impl<'input, Endian> LookupParser<'input, Endian> for ArangeParser<'input, Endian>
-    where Endian: Endianity
+    where Endian: 'input + Endianity
 {
     type Header = ArangeHeader;
     type Entry = ArangeEntry;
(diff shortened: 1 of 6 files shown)


fitzgen:
> I haven't been able to find anything that actually uses .debug_aranges though, so I can't see how clients of the libraries expect to be able to handle them. As far as I can tell, not even the real addr2line uses .debug_aranges, although it's a bit hard to follow the code so I'm not sure.

@tromey told me that he made gdb give up on using it and generate their own tables in an initial pass, because (IIRC) compilers often emit crappy .debug_aranges (or was it something more fundamental about .debug_aranges?).

src/aranges.rs
```diff
@@ -236,3 +252,154 @@ pub type DebugAranges<'input, Endian> = DebugLookup<'input, Endian, ArangeParser
 pub type ArangeEntryIter<'input, Endian> = LookupEntryIter<'input,
                                                            Endian,
                                                            ArangeParser<'input, Endian>>;
+
+#[cfg(test)]
+mod tests {
```
fitzgen: Thanks for adding these unit tests!

fitzgen:
LGTM 💯

Issue #72, opened 2016-08-26
Our `addr2line` fails to symbolicate many addresses that `atos` happily will on OSX


fitzgen:
Discovered in https://github.com/gimli-rs/gimli/issues/53#issuecomment-242655836.

Grab a bunch of addresses in an executable:

```
export ADDRS=$(gobjdump -C -x --disassemble -l $EXEFILE | grep callq -C 4 | grep -v '\-\-' | cut -f1 -d':' | grep -v Users | head -n 100)
```

Run them through `atos` and every single one gets symbolicated:

```
$ atos -o $EXEFILE $ADDRS
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
_$LT$gimli..constants..DwChildren$u20$as$u20$core..fmt..Debug$GT$::fmt::h70fefef2193310dc (in gimli-2ffa6594baf9fafb) (constants.rs:59)
[...]

Issue #53, opened 2016-08-19
addr2line is slower than the binutils version


khuey:
By about 60-70% on querying a handful of symbols from libxul.  Two things jump out at me.
1. We spend a lot of time in memset according to perf record (not memcpy, memset)
2. We're chunking writes in 8KB segments, while binutils does 8MB segments.  So we make 120000 read syscalls instead of binutil's ~200 to load libxul.

fitzgen:
The `obj` modules and interface are far from optimized... Very lowest common denominator. We should probably throw some mmap in there too. I think it is reaching the point where that stuff should be another crate that is done Better and For Real and then our example programs can pull it in as a dev dependency.

fitzgen:
@khuey: I moved the `obj` module out to its own crate: https://github.com/gimli-rs/object

Feel free to go to town optimizing it now and reworking it as needed.

philipc:
It looks like mmap and/or xmas-elf solved this?

fitzgen:
> It looks like mmap and/or xmas-elf solved this?

Hopefully?

I was having a lot of trouble running addr2line so I didn't benchmark... I was doing `nm some-executable-with-dwarf` to get symbols and their addresses out and then trying `addr2line -e some-executable-with-dwarf some-symbols-address` to try and get the symbol and line number back out. This doesn't seem to work with `addr2line`, nor our example clone, but works fine with `atos`. What am I doing wrong?

@khuey can you share the command you were using when benchmarking so I can try and repro?

khuey:
I just used `objdump` to find some addresses to lookup.  Running `objdump -C -x --disassemble -l libxul.so | grep callq -C4` should find you some addresses you can look at.

Then run `addr2line -e libxul.so <address>` and it should work.

khuey:
You should try both a single address (to measure the overhead of starting up) and a lot of them (to measure the actual overhead of each lookup).  For the latter case I don't think it's that important that the addresses are all actually valid, as long as some of them are.

fitzgen:
Ok the binutils port of `addr2line` to OSX is garbage... Can't run it successfully without it trying to free a pointer it didn't malloc.

Will bench against `atos` while I'm on OSX.

fitzgen:
Looks like we pretty much blow `atos` out of the water \o/

```
test bench_atos_100_addrs              ... bench: 129,760,722 ns/iter (+/- 3,294,593)
[...]

PR #77, opened 2016-08-29
addr2line: use DW_AT_low_pc/DW_AT_high_pc instead of .debug_aranges

philipc:
.debug_aranges seems to be poorly supported by compilers and is incomplete.

This is only a partial fix for #72. I'll file issues for the other problems. This is also very slow now, so #53 is still required.

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -23,10 +23,12 @@ fn main() {
         .expect("Should create a mmap for file");
     let file = object::File::parse(unsafe { file.as_slice() });
 
+    let addrs = matches.free.iter().map(|x| parse_uint_from_hex_string(x)).collect();
+
     if file.is_little_endian() {
-        symbolicate::<gimli::LittleEndian>(&file, &matches);
+        symbolicate::<gimli::LittleEndian>(&file, addrs);
     } else {
-        symbolicate::<gimli::BigEndian>(&file, &matches);
+        symbolicate::<gimli::BigEndian>(&file, addrs);
     }
 }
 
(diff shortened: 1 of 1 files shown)


khuey:
> .debug_aranges seems to be poorly supported by compilers and is incomplete.

Well that sucks :/

examples/addr2line.rs
```diff
@@ -111,29 +69,87 @@ fn symbolicate<Endian>(file: &object::File, matches: &getopts::Matches)
...
+    for addr in addrs {
+        find_address(debug_line, &units, addr);
+    }
+}
+
+fn find_address<Endian>(debug_line: gimli::DebugLine<Endian>, units: &[Unit], addr: u64)
+    where Endian: gimli::Endianity
+{
+    for unit in units {
+        if unit.contains_address(addr) {
```
fitzgen: Possible follow up: investigate sorting units by their addresses and doing a binary search here.

fitzgen:
LGTM.

However, for the future, I think we should consider whether we want to implement a higher level, "semantic" API on top of the raw parser/DIE traversal code. Something that has a bunch of `Cow` and can become owned if the API consumer wishes. Basically a nicer type for each DW_TAG_whatever that has (internally cacheable?) methods to get the data that type is expected to have. Something like this (total sketch without lifetimes and all that):

```rust
struct Subprogram {
    ...
}

impl Subprogram {
    pub fn new(die: DebuggingInformationEntry) -> Subprogram {
        assert!(die.tag() == DW_TAG_subprogram);
        Subprogram { ... }
    }

    pub fn name(&self) -> Option<&ffi::CStr> { ... }

    pub fn linkage_name(&self) -> Option<&ffi::CStr> { ... }

    pub fn calling_convention(&self) -> Option<DwCc> { ... }

    pub fn is_pure(&self) -> Option<bool> { ... }

    pub fn is_recursive(&self) -> Option<bool> { ... }

    // Etc...
}
```

fitzgen:
I'm imagining an API like that making the `Unit` type here obsolete.

philipc:
Yeah we definitely need higher level APIs. These should provide the option to cache stuff, and then provide fast access to the cache (eg we should be able to cache abbreviations and line number information for a CU, and use binary search for the line number info). We also may want to provide the option to lazily parse abbreviations. eg if all we want is the CU DIE, and the CU abbreviation is first, then no need to parse them all. I want to finish validating the low level parsing before doing that though.

PR #81, opened 2016-08-29
dwarfdump attribute fixes

philipc:
This handles most of the simple attribute display differences between gimli-dwarfdump and libdwarf-dwarfdump.


fitzgen:
LGTM!

PR #83, opened 2016-08-29
Add `NativeEndian` as a type alias for `BigEndian` or `LittleEndian` depending on target platform

fitzgen:
It's a small convenience.

r? @philipc

### DIFF
--- a/src/endianity.rs
+++ b/src/endianity.rs
@@ -82,6 +82,12 @@ impl byteorder::ByteOrder for BigEndian {
 
 impl Endianity for BigEndian {}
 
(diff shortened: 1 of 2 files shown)


philipc:
LGTM.

I think we still need something that is runtime selectable as well though. https://github.com/gimli-rs/gimli/blob/master/examples/dwarfdump.rs#L79-L83 is fine for simple utilities, but we need some way to stop the endianity bubbling up through all the APIs.

fitzgen:
Thanks! Filed an issue for dynamic endianity.

PR #86, opened 2016-08-30
dwarfdump fixes

philipc:
Fix various things I found while running gimli-dwarfdump on the libdwarf regression tests.

Also filled out more of `Attribute::value`. Heavy macro usage, but I think it is more understandable that way. No tests for this yet... I'm reluctant to add them because I don't think they would provide any value in this case.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -217,7 +217,7 @@ fn dump_attr_value<Endian>(attr: gimli::Attribute<Endian>, debug_str: gimli::Deb
             println!("<0x{:08x}>", offset);
         }
         gimli::AttributeValue::DebugInfoRef(gimli::DebugInfoOffset(offset)) => {
-            println!("0x{:08x}", offset);
+            println!("<GOFF=0x{:08x}>", offset);
         }
         gimli::AttributeValue::DebugLineRef(gimli::DebugLineOffset(offset)) => {
             println!("0x{:08x}", offset);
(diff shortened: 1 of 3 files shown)


examples/dwarfdump.rs
```diff
@@ -156,6 +156,7 @@ fn dump_entries<Endian>(mut entries: gimli::EntriesCursor<Endian>,
     let mut depth = 0;
     while let Some((delta_depth, entry)) = entries.next_dfs().expect("Should parse next dfs") {
         depth += delta_depth;
+        assert!(depth >= 0);
```
fitzgen: You sure you don't want this to just be `debug_assert!`?

src/parser.rs
```diff
@@ -2213,6 +2227,26 @@ fn test_parse_attribute_refaddr_64() {
...
+
+#[test]
+fn test_parse_attribute_refaddr8_version2() {
+    let buf = [0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x99, 0x99];
+    let mut unit = test_parse_attribute_unit::<LittleEndian>(8, Format::Dwarf32);
+    unit.version = 2;
+    let form = constants::DW_FORM_ref_addr;
+    let value = AttributeValue::DebugInfoRef(DebugInfoOffset(0x0807060504030201));
+    test_parse_attribute(&buf, 8, &unit, form, value);
+}
```
fitzgen: Nice!

src/parser.rs
```diff
@@ -1491,228 +1491,356 @@ impl<'input, Endian> Attribute<'input, Endian>
...
-            // DW_AT_sibling: reference
-            // DW_AT_location: exprloc, loclistptr
-            constants::DW_AT_location => {
+        macro_rules! address {
+            () => ();
+        }
+        macro_rules! block {
+            () => ();
+        }
+        macro_rules! constant {
```
fitzgen: All these macros are nice, but it did take me a little bit to figure out what was going on, and how all the `() => ()` are nops/placeholders for future prettifying. I think a comment at the top of the method giving a short overview of the approach would be nice for readers.

fitzgen:
Couple optional nitpicks, but I'll land it and we can handle those in follow ups. Thanks!

examples/dwarfdump.rs
philipc: Possibly it should print a nice error instead and clamp the depth to 0. You definitely don't want to let it go negative, even for release builds.

src/parser.rs
philipc: Yep, I'll fix that up in a few days.

PR #87, opened 2016-08-31
Parse .debug_frame and CFI info

fitzgen:
This adds code to parse the CFI entries (CommonInformationEntry and FrameDescriptionEntry) in the .debug_frame section, but does not expose that functionality yet and doesn't parse instructions either. Nonetheless, it is well tested and worth landing now, IMO.

I tried out using `test-assembler` for these unit tests, and they made writing test cases really nice, especially with the custom trait mixin methods.

Still TODO as follow ups:

* Consolidate length and CIE id/pointer parsing into one `parse_cfi_entry_common` method so that we don't repeat ourselves

* Add `parse_cfi_entry` which uses the method above and then continues parsing either a CIE or FDE depending on which kind of entry we're at. This will let us avoid repeatedly parsing the same length and CIE id/pointer.

* Make some kind of iterator over this data. I can imagine both caching CIEs or re-parsing them on demand, which was why I made FDE::parse take a lambda. Further investigation is needed here.

* Actually, you know, decoding these operations into `RegisterRule`s...

r? @philipc

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1,0 +1,874 @@
+use endianity::{Endianity, EndianBuf};
+use parser::{Error, Format, ParseResult, parse_address, parse_initial_length,
+             parse_null_terminated_string, parse_signed_leb, parse_u8, parse_unsigned_leb,
+             parse_word};
+use std::marker::PhantomData;
+use std::str;
+
+/// An offset into the `.debug_frame` section.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct DebugFrameOffset(pub u64);
+
+/// The `DebugFrame` struct contains the source location to instruction mapping
+/// found in the `.debug_frame` section.
+#[derive(Debug, Clone, Copy)]
+pub struct DebugFrame<'input, Endian>
+    where Endian: Endianity
+{
+    debug_frame_section: EndianBuf<'input, Endian>,
+}
+
+impl<'input, Endian> DebugFrame<'input, Endian>
+    where Endian: Endianity
+{
+    /// Construct a new `DebugFrame` instance from the data in the
+    /// `.debug_frame` section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_frame` section and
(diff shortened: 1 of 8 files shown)


src/cfi.rs
```diff
@@ -0,0 +1,938 @@
...
+            .mark(&start)
+            // CIE Id
+            .L32(0xffffffff)
+            .cie_header(4,
+                        augmentation,
+                        4,
+                        0,
+                        1,
+                        2,
+                        3)
```
philipc: This is an improvement over byte arrays, but it is still too verbose and repetitive.  How about changing `cie_header` to take the expected `CommonInformationEntry` instead (we'll still have to fill in the length afterwards).

src/cfi.rs
```diff
@@ -0,0 +1,938 @@
...
+            // Extra augmented data
+            .L32(1)
+            .L32(2)
+            .L32(3)
+            .L32(4)
```
philipc: Where do these 16 bytes get used/tested?

philipc:
Looks mostly fine.  I think the `test-assembler` usage is definitely an improvement, especially for length calculations.

src/cfi.rs
fitzgen: > Where do these 16 bytes get used/tested?

They don't! This tests that we skip past to the rest of the input correctly, and that we do *NOT* try to read any of the attributes not listed here:

```
If a reader encounters an augmentation string that is unexpected, then only the following fields can be read:
• CIE: length, CIE_id, version, augmentation
• FDE: length, CIE_pointer, initial_location, address_range
```

src/cfi.rs
fitzgen: Nice idea, I'll play with cleaning this up and see what I come up with.

src/cfi.rs
philipc: Do we need to provide the user with a way to access these bytes though?

fitzgen: Maybe? I'm of the feeling that we should cross that bridge when we get to it.

PR #88, opened 2016-09-01
CFI parsing continued...

fitzgen:
This fixes all the TODO follow ups listed in #87 except parsing and executing DW_CFA instructions, which will be its own PR.

See each commit and its commit message for details.

r? @philipc

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -236,7 +236,6 @@ pub struct CommonInformationEntry<'input, Endian>
 impl<'input, Endian> CommonInformationEntry<'input, Endian>
     where Endian: Endianity
 {
-    #[allow(dead_code)]
     fn parse
         (input: EndianBuf<'input, Endian>)
          -> ParseResult<(EndianBuf<'input, Endian>, CommonInformationEntry<'input, Endian>)> {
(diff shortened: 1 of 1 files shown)


src/cfi.rs
```diff
@@ -871,4 +1036,254 @@ mod tests {
...
+            address_range: 9000,
+            instructions: EndianBuf::<BigEndian>::new(&expected_instrs4),
+        };
+
+        let section = section.fde(Endian::Big, &cie1_location, &mut fde1)
+            .fde(Endian::Big, &cie2_location, &mut fde2);
+
+        // TODO: the fact that we have to explicitly set this seems like a bug
+        // in test-assembler... Need to file a bug there.
+        section.start().set_const(0);
```
fitzgen: Filed https://github.com/luser/rust-test-assembler/issues/4

src/cfi.rs
```diff
@@ -119,15 +240,16 @@ impl<'input, Endian> CommonInformationEntry<'input, Endian>
     fn parse
```
philipc: Nit: this is no longer dead code.

src/cfi.rs
```diff
@@ -233,18 +387,23 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
     #[allow(dead_code)]
```
philipc: This is still dead code. I don't think it hurts to have it, but I don't think this will ever be used outside of the tests? Also, maybe that means the tests shouldn't be using it?

philipc:
LGTM. Nice commits.

src/cfi.rs
fitzgen: I imagine if/when we add an `fde_from_offset` method, this will no longer be dead code. However, I need to read the standard in more depth to see if we even want that method. If not, I'll port all tests over to using `PartialFrameDescriptionEntry`.

fitzgen:
Thanks!

PR #92, opened 2016-09-05
Include all tests in coverage

philipc:


philipc:
Cargo only creates temporary executables for doc tests, so they can't be included.

I haven't added benches/examples yet. They shouldn't be hard to add, but they're not actually tests, so not sure they should be included.

I can't browse the files on coveralls.io anymore with this change, but I can't see what might have caused that.

We're currently doing coverage for stable, beta and nightly. Is that useful or should we change it to stable only?

fitzgen:
> I can't browse the files on coveralls.io anymore with this change, but I can't see what might have caused that.

In order to make them browse-able initially, I had to set "src" as the root path somewhere in the coveralls settings. I guess this broke that because now there are also files whose path starts with "tests".

fitzgen:
Thanks for tackling this @philipc :)

fitzgen:
> > I can't browse the files on coveralls.io anymore with this change, but I can't see what might have caused that.
> 
> In order to make them browse-able initially, I had to set "src" as the root path somewhere in the coveralls settings. I guess this broke that because now there are also files whose path starts with "tests".

Ok should be fixed now, eg: https://coveralls.io/builds/7752904/source?filename=src%2Fline.rs

PR #94, opened 2016-09-06
Parse call frame instructions

fitzgen:
* Defines a CallFrameInstruction type to represent parsed call frame
  instructions
* Exposes a FallibleIterator to lazily parse instructions
* Adds methods to FDEs and CIEs to iterate and parse their instructions

r? @philipc

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1,8 +1,10 @@
+use constants;
 use endianity::{Endianity, EndianBuf};
 use fallible_iterator::FallibleIterator;
 use parser::{Error, Format, ParseResult, parse_address, parse_initial_length,
-             parse_null_terminated_string, parse_signed_leb, parse_u8, parse_unsigned_leb,
-             parse_word};
+             parse_length_uleb_value, parse_null_terminated_string, parse_signed_leb,
+             parse_signed_lebe, parse_u8, parse_u8e, parse_u16, parse_u32, parse_unsigned_leb,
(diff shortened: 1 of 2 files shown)


src/cfi.rs
```diff
@@ -475,6 +499,520 @@ pub enum RegisterRule<'input, Endian>
...
+        }
+
+        if high_bits == constants::DW_CFA_restore.0 {
+            let register = instruction & CFI_INSTRUCTION_LOW_BITS_MASK;
+            return Ok((rest, CallFrameInstruction::Restore { register: register as u64 }));
+        }
+
+        debug_assert!(high_bits == 0);
+
+        if instruction == constants::DW_CFA_nop.0 {
```
philipc: Nit: how about `let instruction = DwCfa(instruction);` and then drop the `.0`.

Is there a reason you used `if` statements instead of `match`?

philipc:
LGTM

src/cfi.rs
fitzgen: > Nit: how about let instruction = DwCfa(instruction); and then drop the .0.

Done

> Is there a reason you used if statements instead of match?

I wanted to avoid rightward drift since the high bit masking would lead to two matches, each of which is double indentation, so quad indentation. I ended up making the second larger set (when the high bits aren't set) a match.

fitzgen:
Thanks!

Issue #34, opened 2016-08-12
Split up parser.rs into smaller modules


fitzgen:
Would generally like to have stuff related to each section in its own module, eg `.debug_line` stuff is in `src/line.rs`, `.debug_abbrev` stuff is in `src/abbrev.rs`, and `.debug_str` stuff is in `src/str.rs` (as of a minute ago).

The catch is that the meaty bits of `parser.rs` is DIE tree things that apply equally to both `.debug_info` and `.debug_types`. Not sure how to best break this down into smaller pieces or where exactly boundaries should lie. Make an `entries.rs` or `die.rs` for this shared stuff and then make `info.rs` and `types.rs` relatiely little?

Any thoughts or opinions @philipc?

philipc:
I did the same thing in rust-dwarf a few days ago (but haven't pushed it yet). I split it up as `abbrev.rs`, `die.rs`, and `unit.rs` (and str/line/etc in future). `die.rs` contained the cursor/entry, and `unit.rs` contained info and types. I'm not sure it's the best split, but if you put all the common unit stuff in `die.rs`, then it feels odd to have other unit stuff in separate files. If you really wanted to split it up more you could do info/types/unit/die, but I think it's probably better for info/types/unit to be all together.

BTW, I'm going to keep doing bits in rust-dwarf, as a playground for ideas, and also because I'm not sure that the write support is going to integrate well with gimli, so I don't want to throw that away yet.

philipc:
I think this is done.

PR #95, opened 2016-09-06
Create unit.rs

philipc:
This is a minimal first step for #34. I renamed parser.rs to unit.rs, then moved the common parsing functions back to parser.rs. We can split out more of unit.rs to other files as needed. Look at the individual commits if you want to view this online.

Note: this will conflict with #94


fitzgen:
LGTM -- needs rebase since I just merged #94.

Aside: any opinion on what percentage of coverage regression we want to permit? -.2% causing a CI "failure" is a bit harsh... Maybe set it to a full 1%? 2%?

fitzgen:
Also: thanks for doing this!

philipc:
> Aside: any opinion on what percentage of coverage regression we want to permit? -.2% causing a CI "failure" is a bit harsh... Maybe set it to a full 1%? 2%?

I prefer to use judgement rather than hard rules, so I see the failed coveralls check as something to warn you to investigate, rather than being a reason to deny the PR. Feel free to bump it up a bit if it's too noisy though.

PR #96, opened 2016-09-06
Add SectionData/SectionOffset abstractions

philipc:
Reduces the amount of copy/paste for sections and offsets, while keeping mostly the same API. The main difference in the API is the need to use `new` to create offsets, but that is generally only used in tests.

@fitzgen  I haven't finished converting everything. Let me know if you agree with the direction this is heading; I'm not too sure about it myself. The main problem I can see with this approach is that the documentation doesn't pick up all the extra functions for the type aliases, such as `DebugInfo::units`, so we will need to manually do what we did for `DebugPubNames` and `DebugPubTypes` (or fix rustdoc; other people have reported similar problems with type aliases so it's a needed feature).

Also a general doc comment: I think we need to start exposing modules, because the main page is too long (or at least expose the `constants` module).

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -246,8 +246,8 @@ fn dump_attr_value<Endian>(attr: gimli::Attribute<Endian>,
         gimli::AttributeValue::UnitRef(gimli::UnitOffset(offset)) => {
             println!("<0x{:08x}>", offset);
         }
-        gimli::AttributeValue::DebugInfoRef(gimli::DebugInfoOffset(offset)) => {
-            println!("<GOFF=0x{:08x}>", offset);
+        gimli::AttributeValue::DebugInfoRef(offset) => {
+            println!("<GOFF=0x{:08x}>", offset.0);
         }
         gimli::AttributeValue::DebugLineRef(gimli::DebugLineOffset(offset)) => {
             println!("0x{:08x}", offset);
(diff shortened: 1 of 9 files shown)


fitzgen:
> Also a general doc comment: I think we need to start exposing modules, because the main page is too long (or at least expose the constants module).

+1 for the `constants` module -- that could definitely be encapsulated as one.

Unsure about others, however. It feels tidy from our (library authors) side, but as a library consumer, I always forget what submodules some type is in and it is generally easier when a flat hierarchy is exposed. Everything that comes from a foreign crate is equally "other" from this perspective.

fitzgen:
> The main problem I can see with this approach is that the documentation doesn't pick up all the extra functions for the type aliases, such as DebugInfo::units, so we will need to manually do what we did for DebugPubNames and DebugPubTypes (or fix rustdoc; other people have reported similar problems with type aliases so it's a needed feature).

The other thing we can do is newtypes (ie `struct NewType(OtherType)`) that wrap the helper rather than using type aliases. It involves a bit more boilerplate, though, as we need to re-define all the public methods.

fitzgen:
Since the section types are exposed publicly in `src/lib.rs`, can't the docs show the underlying methods? Do you just mean without the extra click through?

fitzgen:
Since the section types are exposed publicly in `src/lib.rs`, can't the docs show the underlying methods? Do you just mean without the extra click through?

fitzgen:
Since the section types are exposed publicly in `src/lib.rs`, can't the docs show the underlying methods? Do you just mean without the extra click through?

fitzgen:
Since the section types are exposed publicly in `src/lib.rs`, can't the docs show the underlying methods? Do you just mean without the extra click through?

fitzgen:
(Sorry for multiple comments, @github's JS is super buggy for me today)

fitzgen:
Overall, I like this change. I wasn't feeling much pain from a bit of repetition before, but this is nice.

fitzgen:
Oh, also I just remembered: I added some more documentation in `src/lib.rs` with links to various section types. Please update the links as the types change here -- thanks!

philipc:
> Since the section types are exposed publicly in src/lib.rs, can't the docs show the underlying methods? Do you just mean without the extra click through?

I mean the methods specific to each section type (eg `DebugInfo::units`). rustdoc doesn't show these anywhere.

philipc:
> The other thing we can do is newtypes (ie struct NewType(OtherType)) that wrap the helper rather than using type aliases. It involves a bit more boilerplate, though, as we need to re-define all the public methods.

That would probably solve it, but given that the amount of repetition this is solving is low, I think that would complicate things for little net benefit.

I'm going to hold off on this change for now. I'll try to reduce the amount of duplicate offset parsing in another PR.

PR #98, opened 2016-09-08
Parse .debug_ranges

### LINKED ISSUE #78: addr2line: use DW_AT_ranges
Related to #72 and #77.

We should use the `DW_AT_ranges` attribute to determine which addresses are contained in a compilation unit.

philipc:
Fixes #78

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -1,8 +1,10 @@
+extern crate fallible_iterator;
 extern crate gimli;
 extern crate getopts;
 extern crate memmap;
 extern crate object;
 
+use fallible_iterator::FallibleIterator;
 use object::Object;
 use std::env;
 use std::fs;
@@ -67,12 +69,14 @@ fn symbolicate<Endian>(file: &object::File, addrs: Vec<u64>)
     let debug_abbrev = gimli::DebugAbbrev::<Endian>::new(debug_abbrev);
     let debug_line = file.get_section(".debug_line")
         .expect("Can't addr2line without .debug_line");
-    let debug_line = gimli::DebugLine::<Endian>::new(&debug_line);
+    let debug_line = gimli::DebugLine::<Endian>::new(debug_line);
+    let debug_ranges = file.get_section(".debug_ranges").unwrap_or(&[]);
+    let debug_ranges = gimli::DebugRanges::<Endian>::new(debug_ranges);
 
     let mut units = Vec::new();
     let mut headers = debug_info.units();
     while let Some(header) = headers.next().expect("Couldn't get DIE header") {
-        if let Some(unit) = Unit::parse(&debug_abbrev, &header) {
+        if let Some(unit) = Unit::parse(&debug_abbrev, &debug_ranges, &header) {
             units.push(unit);
         }
     }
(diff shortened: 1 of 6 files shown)


examples/addr2line.rs
```diff
@@ -98,48 +102,97 @@ fn find_address<Endian>(debug_line: gimli::DebugLine<Endian>, units: &[Unit], ad
...
+        };
+
+        let high_pc = match entry.attr_value(gimli::DW_AT_high_pc) {
+            Some(gimli::AttributeValue::Addr(addr)) => addr,
+            Some(gimli::AttributeValue::Udata(size)) => low_pc.wrapping_add(size),
+            None => low_pc.wrapping_add(1),
+            _ => return None,
+        };
+
+        assert!(low_pc < high_pc);
```
fitzgen: `debug_assert!` ?

examples/addr2line.rs
```diff
@@ -98,48 +102,97 @@ fn find_address<Endian>(debug_line: gimli::DebugLine<Endian>, units: &[Unit], ad
...
+        Some(gimli::Range {
+            begin: low_pc,
+            end: high_pc,
+        })
+    }
+
     fn contains_address(&self, address: u64) -> bool {
-        self.high_pc == 0 || address >= self.low_pc && address <= self.high_pc
+        self.ranges.iter().find(|range| address >= range.begin && address < range.end).is_some()
```
fitzgen: Nit: `self.ranges.iter().any(|range| address >= range.begin && address < range.end)`

src/constants.rs
```diff
@@ -658,6 +658,6 @@ dw!(DwOp(u8) {
     DW_OP_implicit_value = 0x9e,
     DW_OP_stack_value = 0x9f,
 
-    // GNU extensions that are supported by our expression evaluator.
+// GNU extensions that are supported by our expression evaluator.
```
fitzgen: Unintentional?

src/lib.rs
```diff
@@ -80,6 +80,8 @@
 //!   * [`DebugPubTypes`](./type.DebugPubTypes.html): The `.debug_pubtypes`
 //!   section.
 //!
+//!   * [`DebugRanges`](./struct.DebugRanges.html): The `.debug_ranges` section.
+//!
```
fitzgen: Thanks :)

src/ranges.rs
```diff
@@ -0,0 +1,379 @@
...
+            0x00, 0x06, 0x01, 0x00, 0x00, 0x06, 0x01, 0x00,
+            0x00, 0x08, 0x01, 0x00, 0x00, 0x09, 0x01, 0x00,
+            // A range that starts at 0.
+            0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+            // A range that ends at -1.
+            0xff, 0xff, 0xff, 0xff, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0xff, 0xff, 0xff, 0xff,
+            // A range end.
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0xff,
```
fitzgen: Don't want to use `test-assembler` for constructing these?

fitzgen:
LGTM, but I didn't read up on the exact form of `.debug_ranges` in the standard.

examples/addr2line.rs
philipc: No, this depends on the input file. Added a comment.

examples/addr2line.rs
philipc: Done.

src/constants.rs
philipc: Yes, this is due to running rustfmt. Will fix that in a separate commit.

src/ranges.rs
philipc: Habit :) Done.

philipc:
Thanks for the review.

I also noticed that coveralls isn't showing the new file.

src/ranges.rs
```diff
@@ -0,0 +1,453 @@
+use endianity::{Endianity, EndianBuf};
```
tromey: Other files have `#![deny(missing_docs)]` at the top, which seems like a good idea.

philipc: It's a good idea, but I think this should be done as a crate wide setting, and it seems that the `#![deny(missing_docs)]` in `lib.rs` is enough.

PR #99, opened 2016-09-09
Remove unnecessary `#![deny(missing_docs)]` attributes

fitzgen:
Only the one in lib.rs is needed.

r? @philipc


philipc:
LGTM

fitzgen:
> Coverage decreased (-1.5%)

This... makes no sense. I don't have much faith in kcov and rust combined.

PR #100, opened 2016-09-09
Rename gimli::ParseResult to gimli::Result

fitzgen:
Follows conventions established by std::io::Result and std::fmt::Result, etc.

r? @philipc 

heads up @tromey

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1,10 +1,9 @@
 use constants;
 use endianity::{Endianity, EndianBuf};
 use fallible_iterator::FallibleIterator;
-use parser::{Error, Format, ParseResult, parse_address, parse_initial_length,
-             parse_length_uleb_value, parse_null_terminated_string, parse_signed_leb,
-             parse_signed_lebe, parse_u8, parse_u8e, parse_u16, parse_u32, parse_unsigned_leb,
-             parse_unsigned_lebe, parse_word};
+use parser::{Error, Format, Result, parse_address, parse_initial_length, parse_length_uleb_value,
+             parse_null_terminated_string, parse_signed_leb, parse_signed_lebe, parse_u8,
+             parse_u8e, parse_u16, parse_u32, parse_unsigned_leb, parse_unsigned_lebe, parse_word};
 use std::marker::PhantomData;
 use std::str;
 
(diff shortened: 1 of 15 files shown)


src/cfi.rs
```diff
@@ -104,7 +103,7 @@ impl<'input, Endian> FallibleIterator for CfiEntriesIter<'input, Endian>
     type Item = CieOrFde<'input, Endian>;
     type Error = Error;
 
-    fn next(&mut self) -> Result<Option<Self::Item>, Self::Error> {
+    fn next(&mut self) -> ::std::result::Result<Option<Self::Item>, Self::Error> {
```
philipc: We could change these to `Result<Option<Self::Item>>`. Not sure if that's clearer though.

src/pubnames.rs
```diff
@@ -136,7 +136,7 @@ pub type DebugPubNames<'input, Endian> = DebugLookup<'input,
 ///
 /// Provides:
 ///
-/// * `next_entry(self: &mut) -> ParseResult<Option<PubNamesEntry>>`
+/// * `next_entry(self: &mut) -> gimli::Result<Option<PubNamesEntry>>`
```
philipc: I think this method has been renamed to `next` by the way.

src/pubtypes.rs
```diff
@@ -134,7 +134,7 @@ pub type DebugPubTypes<'input, Endian> = DebugLookup<'input,
 ///
 /// Provides:
 ///
-/// * `next_entry(self: &mut) -> ParseResult<Option<PubTypesEntry>>`
+/// * `next_entry(self: &mut) -> gimli::Result<Option<PubTypesEntry>>`
```
philipc: And this too.

src/constants.rs
```diff
@@ -658,6 +658,6 @@ dw!(DwOp(u8) {
     DW_OP_implicit_value = 0x9e,
     DW_OP_stack_value = 0x9f,
 
-    // GNU extensions that are supported by our expression evaluator.
+// GNU extensions that are supported by our expression evaluator.
```
philipc: Yeah, I haven't fixed this yet, and not sure the best way (rustfmt_skip?)

philipc:
LGTM

src/constants.rs
fitzgen: I didn't realize it was rustfmt's doing when I made that comment, although I guess I should have realized. Given that, I don't think it's a big deal.

src/cfi.rs
fitzgen: The original motivation was that I'd kind of like this to fail loudly if we change any of the `next` signatures, but I suppose the type checker will do that either way. 

/me shrugs

src/pubnames.rs
fitzgen: Done in a follow up.

PR #101, opened 2016-09-09
addr2line improvements

### LINKED ISSUE #79: addr2line doesn't report lines that start before the requested address
Related to #72 and #77.

`line::StateMachine::run_to_address` requires an exact address match. We should change it to stop at the first row that contains the address. This will need to take sequences into account.

philipc:
Handle addresses within rows, and display compilation unit directory. As usual, some of this logic needs moving into the library eventually.

Needed a change to `LineNumberRow`. Open to suggestions for other ways to fix this.

Fixes #79

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -42,11 +42,12 @@ fn parse_uint_from_hex_string(string: &str) -> u64 {
     }
 }
 
-fn display_file<Endian>(row: gimli::LineNumberRow<Endian>)
+fn display_file<Endian>(header: &gimli::LineNumberProgramHeader<Endian>,
+                        row: &gimli::LineNumberRow)
     where Endian: gimli::Endianity
 {
-    let file = row.file().unwrap();
-    if let Some(directory) = file.directory(row.header()) {
+    let file = row.file(header).unwrap();
+    if let Some(directory) = file.directory(header) {
         println!("{}/{}:{}",
                  directory.to_string_lossy(),
                  file.path_name().to_string_lossy(),
@@ -92,8 +93,8 @@ fn find_address<Endian>(debug_line: gimli::DebugLine<Endian>, units: &[Unit], ad
     for unit in units {
         if unit.contains_address(addr) {
             if let Ok(mut lines) = unit.lines(debug_line) {
-                if let Ok(Some(row)) = lines.run_to_address(&addr) {
-                    display_file(row);
+                if let Ok(Some((header, row))) = lines.run_to_address(&addr) {
+                    display_file(header, row);
                     return;
                 };
             }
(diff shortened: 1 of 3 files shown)


fitzgen:
> The problem before was that the struct contained a reference to
> StateMachineRegisters, but the user couldn't clone that because it is
> private.

Well, they could clone *the reference* just fine (indirectly through `LineNumberRow`'s `Clone` impl), but it would still be a reference and still restrict the lifetime of the cloned result. The goal is to make somethig that can be owned, right? So it can be put in a vec or whatever?

fitzgen:
Regarding first commit:

It seems like everything we do, we really want an owned and borrowed version (and maybe the `ToOwned` trait and a `Cow` on top of them)... But that's a lot of boilerplate-y stuff...

Gah. I say we hold off until we have more clients of the library and things solidify a bit.

examples/addr2line.rs
```diff
@@ -90,13 +90,24 @@ fn symbolicate<Endian>(file: &object::File, addrs: Vec<u64>)
...
+    let mut current = None;
     for unit in units {
         if unit.contains_address(addr) {
             if let Ok(mut lines) = unit.lines(debug_line) {
-                if let Ok(Some((header, row))) = lines.run_to_address(&addr) {
-                    display_file(header, row);
-                    return;
-                };
+                while let Ok(Some((header, row))) = lines.next_row() {
+                    if row.address() > addr {
```
fitzgen: I *think* we should be able to add the end address to each row, so we could give the row a `contains(&self, address: u64)` method. If we did this, we wouldn't need to repeatedly make owned copies in `current`.

Actually, the next row's start address isn't immediately available, and we would have to fully evaluate the next row to get that in the worst case.

Nevermind...

fitzgen:
LGTM

philipc:
> It seems like everything we do, we really want an owned and borrowed version (and maybe the ToOwned trait and a Cow on top of them)... But that's a lot of boilerplate-y stuff...

I think that having separate owned and borrowed versions should only be needed for dynamically sized types. In this case, the size of a row should be fixed. Cow won't help because that would mean the iterator would need to copy the row before modifying it to create the next row, and we're trying to avoid copies unless the caller wants them.

The constraint we need is that the lifetime of the value of the row (not the reference to the row) must not depend on the lifetime of the iterator.

Now we could have an `OwnedLineNumberRow` that contains a `StateMachineRegisters` value, and create that by copying the value of the registers from the borrowed version, but that still doesn't solve the problem that the header field was a reference with the iterator lifetime too, and I'm pretty sure we don't want to copy a header value for every row. I also don't see what benefit this would have over the change in this PR.

We could go back to always copying the registers, instead of returning a reference from the iterator. I think that would be a simpler API, but it would be bad for performance, because the registers aren't small. Note that this copy would be in addition to any move that the user wanted (such as when putting in a Vec), because I don't think rust can combine the copy and move (I don't think even placement new would help, because of the `Result`). On the other hand, if the normal use case is to read the rows once and put them in a Vec, then maybe performance here doesn't matter.

PR #105, opened 2016-09-13
Fix bugs in expression evaluator

tromey:
I realized that some operations weren't using wrapping arithmetic.
While fixing these I realized that the reason these weren't causing
assertions was that several tests were using DW_OP_skip and simply
jumping over the remaining tests.  Fixing this in turn revealed a few
more bugs.

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -1518,10 +1518,10 @@ impl<'context, 'input, Endian> Evaluation<'context, 'input, Endian>
             Operation::Div => {
                 let v1 = try!(self.pop_signed());
                 let v2 = try!(self.pop_signed());
-                if v2 == 0 {
+                if v1 == 0 {
                     return Err(Error::DivisionByZero);
                 }
-                self.push((v2 / v1) as u64);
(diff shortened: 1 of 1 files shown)


src/op.rs
```diff
@@ -1566,17 +1566,39 @@ impl<'context, 'input, Endian> Evaluation<'context, 'input, Endian>
             Operation::Shl => {
                 let v1 = try!(self.pop());
                 let v2 = try!(self.pop());
-                self.push(v2 << v1);
+                // Because wrapping_shl takes a u32, not a u64, we do
+                // the check by hand.
+                if v1 >= 32 {
```
philipc: Should this be either 64 or address_size * 8? Same for the other shifts.

tromey: Oh duh, it should be 64.  I'll add a test for that.
It doesn't have to match the target's type here, just whatever native type we're using in the implementation.

fitzgen:
LGTM feel free to merge yourself :)

PR #106, opened 2016-09-13
Parse and dump .debug_loc

### LINKED ISSUE #104: Implement support for .debug_loc
Section 2.6.2 Location Lists

philipc:
Fixes #104

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -91,14 +91,26 @@ fn dump_file<Endian>(file: object::File, flags: &Flags)
 {
     let debug_abbrev = file.get_section(".debug_abbrev").unwrap_or(&[]);
     let debug_abbrev = gimli::DebugAbbrev::<Endian>::new(debug_abbrev);
+    let debug_loc = file.get_section(".debug_loc").unwrap_or(&[]);
+    let debug_loc = gimli::DebugLoc::<Endian>::new(debug_loc);
     let debug_ranges = file.get_section(".debug_ranges").unwrap_or(&[]);
     let debug_ranges = gimli::DebugRanges::<Endian>::new(debug_ranges);
     let debug_str = file.get_section(".debug_str").unwrap_or(&[]);
     let debug_str = gimli::DebugStr::<Endian>::new(debug_str);
 
     if flags.info {
-        dump_info(&file, debug_abbrev, debug_ranges, debug_str, flags);
-        dump_types(&file, debug_abbrev, debug_ranges, debug_str, flags);
+        dump_info(&file,
+                  debug_abbrev,
+                  debug_loc,
+                  debug_ranges,
+                  debug_str,
+                  flags);
+        dump_types(&file,
+                   debug_abbrev,
+                   debug_loc,
+                   debug_ranges,
+                   debug_str,
+                   flags);
     }
     if flags.line {
         dump_line(&file, debug_abbrev);
@@ -110,6 +122,7 @@ fn dump_file<Endian>(file: object::File, flags: &Flags)
 
 fn dump_info<Endian>(file: &object::File,
                      debug_abbrev: gimli::DebugAbbrev<Endian>,
+                     debug_loc: gimli::DebugLoc<Endian>,
                      debug_ranges: gimli::DebugRanges<Endian>,
                      debug_str: gimli::DebugStr<Endian>,
                      flags: &Flags)
@@ -129,6 +142,7 @@ fn dump_info<Endian>(file: &object::File,
             dump_entries(unit.entries(&abbrevs),
                          unit.address_size(),
                          unit.format(),
+                         debug_loc,
                          debug_ranges,
                          debug_str,
                          flags);
@@ -138,6 +152,7 @@ fn dump_info<Endian>(file: &object::File,
 
 fn dump_types<Endian>(file: &object::File,
                       debug_abbrev: gimli::DebugAbbrev<Endian>,
+                      debug_loc: gimli::DebugLoc<Endian>,
                       debug_ranges: gimli::DebugRanges<Endian>,
                       debug_str: gimli::DebugStr<Endian>,
                       flags: &Flags)
@@ -157,6 +172,7 @@ fn dump_types<Endian>(file: &object::File,
             dump_entries(unit.entries(&abbrevs),
                          unit.address_size(),
                          unit.format(),
+                         debug_loc,
                          debug_ranges,
                          debug_str,
                          flags);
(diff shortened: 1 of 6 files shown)


src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+    ///
+    /// Can be [used with
+    /// `FallibleIterator`](./index.html#using-with-fallibleiterator).
+    pub fn locations(&self,
+                     offset: DebugLocOffset,
+                     address_size: u8,
+                     base_address: u64)
+                     -> Result<LocationListIter<Endian>> {
+        let offset = offset.0 as usize;
+        if self.debug_loc_section.len() < offset {
```
fitzgen: This should be `<=` right?

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+    /// such as handling base addresses.
+    ///
+    /// Can be [used with
+    /// `FallibleIterator`](./index.html#using-with-fallibleiterator).
+    pub fn raw_locations(&self,
+                         offset: DebugLocOffset,
+                         address_size: u8)
+                         -> Result<RawLocationListIter<Endian>> {
+        let offset = offset.0 as usize;
+        if self.debug_loc_section.len() < offset {
```
fitzgen: ditto?

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+        let input = self.debug_loc_section.range_from(offset..);
+        Ok(RawLocationListIter::new(input, address_size))
+    }
+}
+
+/// A raw iterator over a location list.
+///
+/// This iterator does not perform any processing of the location entries,
+/// such as handling base addresses.
+#[derive(Debug)]
```
fitzgen: Derive `Clone` too?

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+{
+    type Item = LocationListEntry<'input, Endian>;
+    type Error = Error;
+
+    fn next(&mut self) -> ::std::result::Result<Option<Self::Item>, Self::Error> {
+        RawLocationListIter::next(self)
+    }
+}
+
+/// An iterator over a location list.
```
fitzgen: It would be nice to explain the base address processing in this doc comment.

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+{
+    raw: RawLocationListIter<'input, Endian>,
+    base_address: u64,
+}
+
+impl<'input, Endian> LocationListIter<'input, Endian>
+    where Endian: Endianity
+{
+    /// Construct a `LocationListIter`.
+    pub fn new(input: EndianBuf<'input, Endian>,
```
fitzgen: Should this be `pub`? A lot of our iterators don't have public constructors, but its something that I've always been on the fence with. On the one hand, it means people can only get valid iterators (barring bugs in our library) but on the other hand it makes mocking more difficult.

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+/// A location list entry from the `.debug_loc` section.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LocationListEntry<'input, Endian>
+    where Endian: Endianity
+{
+    /// The address range that this location is valid for.
+    pub range: Range,
+
+    /// The data containing a single location description.
+    pub data: EndianBuf<'input, Endian>,
```
fitzgen: I slightly prefer keeping members private and having getter methods. It allows the most flexibility on our end and isn't really any less ergonomic for users. The one thing is that it disallows modifying the members, which is desirable IMO.

Is there a reason you preferred public members here?

src/loc.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+impl<'input, Endian> LocationListEntry<'input, Endian>
+    where Endian: Endianity
+{
+    /// Parse an address range entry from `.debug_ranges` or `.debug_loc`.
+    #[doc(hidden)]
+    #[inline]
+    pub fn parse(input: EndianBuf<Endian>,
```
fitzgen: Why does this need to be `pub` and `doc(hidden)`? Is it used outside this module?

src/ranges.rs
```diff
@@ -240,6 +227,34 @@ impl Range {
...
+        debug_assert!(!self.is_base_address(address_size));
+        let mask = !0 >> (64 - address_size * 8);
+        self.begin = base_address.wrapping_add(self.begin) & mask;
+        self.end = base_address.wrapping_add(self.end) & mask;
+    }
+
+    /// Parse an address range entry from `.debug_ranges` or `.debug_loc`.
+    #[doc(hidden)]
+    #[inline]
+    pub fn parse<Endian>(input: EndianBuf<Endian>,
```
fitzgen: ditto

examples/dwarfdump.rs
```diff
@@ -433,6 +468,54 @@ fn dump_op<Endian>(dwop: gimli::DwOp, op: gimli::Operation<Endian>, data: &[u8])
...
+        print!("\t\t\t[{:2}]", i);
+        if location.range.is_end() {
+            println!("<end-of-list>");
+        } else if location.range.is_base_address(address_size) {
+            println!("<new base address 0x{:08x}>", location.range.end);
+            base_address = location.range.end;
+        } else {
+            let mut range = location.range;
+            range.add_base_address(base_address, address_size);
+            // This messed up formatting matches libdwarf-dwarfdump.
```
fitzgen: 😆

fitzgen:
Overall, LGTM, just a few questions where I was unsure why things were a certain way

src/loc.rs
philipc: Not sure. It won't panic, but the iterator will immediately return None. I'm also not sure if we should make the iterator more strict and require the list end entry, in which case it wouldn't matter that this was <.

We currently use < nearly everywhere for this type of check (and we're missing the check in some places, eg debug_line, debug_abbrev).

src/loc.rs
philipc: Wouldn't hurt, not sure it would be useful though. Should we derive `Debug` and `Clone` for every struct? We're a bit inconsistent currently.

src/loc.rs
philipc: Will do. I didn't think it was important because all the raw processing is only there so that we can clone the dwarfdump output. The user should never actually need to use the raw ranges or need to care about whether the base address processing was done.

src/loc.rs
philipc: Will change. I'm happy to be conservative about exposing things until there is a need.

src/loc.rs
philipc: I slightly prefer to use plain old data structs unless there are invariants that need to be maintained. The getters/setters seem like unnecessary boilerplate. You can disallow modifying members using `mut`. I'm happy to change it for consistency, but we do have other plain old data structs in use already.

src/loc.rs
philipc: Will fix. Copy/paste from range.rs.

src/ranges.rs
philipc: Use in loc.rs.

src/loc.rs
philipc: Also, I don't see structs in enums as being any different from this case. eg https://github.com/gimli-rs/gimli/pull/107/files#diff-40fe1838d9c931774e10ee89393a864eR1158

philipc:
Will do follow ups for any further changes that are needed.

src/loc.rs
fitzgen: Fair point, WFM

src/loc.rs
fitzgen: Yeah I think we should. There might be some pragmas we could use to help the compiler enforce this

PR #107, opened 2016-09-14
Implement support for evaluating call frame unwinding programs

fitzgen:
Continuing work on #7... Phew!

This commit adds support for evaluating `CallFrameInstruction`s, yielding what
the standard calls the "virtual unwind table". The virtual unwind table is
represented here as a streaming iterator over its rows.

To minimize the number of allocations required when repeatedly running many CFI
programs (for example, when unwinding the whole stack), the context in which a
CFI program runs has been split out into its own re-usable type. Furthermore,
because the `DW_CFA_restore` instruction allows a register's rule to be restored
to the value defined after evaluating the CIE's initial instructions but before
evaluating any of the FDE's instructions, we need to keep track of the initial
register rules and if we evaluated them or not yet. To shelter library users
from tracking this state manually, we push it into the type system via
new-types (`InitializedUnwindContext` and `UninitializedUnwindContext`), and
only these new-types are exposed publicly.

r? @philipc @tromey -- I think this is big (and important) enough to get both of your
eyes on it.

Still TODO in follow ups:

* Be robust against overflow / casting outside of target type's range
* Support `.eh_frame` augmentation
* Benches
* Self-parse integration test (requires the `.eh_frame` augmentation)
* Output this info in dwarfdump.rs
* A single, orchestrating `DebugFrame::unwind_info_for_address(&self, ctx: &mut UninitializedUnwindContext, address: u64) -> UnwindTableRow` method (for folks who don't have any kind of CIE/FDE caching or anything like that)

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -444,13 +444,668 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
         &self.cie
     }
 
-    /// Iterate over this FDE's instructions. Does not include the CIE's initial
-    /// instructions.
+    /// Iterate over this FDE's instructions.
+    ///
+    /// Will not include the CIE's initial instructions, if you want those do
+    /// `fde.cie().instructions()` first.
     ///
     /// Can be [used with
     /// `FallibleIterator`](./index.html#using-with-fallibleiterator).
     pub fn instructions(&self) -> CallFrameInstructionIter<'input, Endian> {
         CallFrameInstructionIter { input: self.instructions }
+    }
+}
+
+/// Common context needed when evaluating the call frame unwinding information.
+///
+/// To avoid re-allocating the context multiple times when evaluating multiple
+/// CFI programs, it can be reused. At first, a context is uninitialized
+/// (`UninitializedUnwindContext`). It can be initialized by providing the
+/// `CommonInformationEntry` for the CFI program about to be evaluated and
+/// calling `UninitializedUnwindContext::initialize`. The result is an
+/// `InitializedUnwindContext`, which can be used to evaluate and run a
+/// `FrameDescriptionEntry`'s CFI program. When the CFI program is complete, the
+/// context can be de-initialized by calling `InitializedUnwindContext::reset`.
+///
+/// ```
+/// use gimli::{UninitializedUnwindContext, UnwindTable};
+///
+/// # fn foo<'a>(some_fde: gimli::FrameDescriptionEntry<'a, gimli::LittleEndian>) -> gimli::Result<()> {
+/// // An uninitialized context.
+/// let ctx = UninitializedUnwindContext::new();
+///
+/// // Initialize the context by evaluating the CIE's initial instruction program.
+/// let mut ctx = try!(ctx.initialize(some_fde.cie()));
+///
+/// {
+///     // The initialized context can now be used to generate the unwind table.
+///     let mut table = UnwindTable::new(&mut ctx, &some_fde);
+///     while let Some(row) = try!(table.next_row()) {
+///         // Do stuff with each row...
+/// #       let _ = row;
+///     }
+/// }
+///
+/// // Reset the context to the uninitialized state and re-use it with other CFI
+/// // programs.
+/// let ctx = ctx.reset();
+/// # let _ = ctx;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// In general, the states will flow from one to the other in accordance to the
+/// following diagram:
+///
+/// ```text
+///         +-------+
+///         | Start |
+///         +-------+
+///             |
+///             |
+/// UninitializedUnwindContext::new()
+///             |
+///             |
+///             V
+/// +----------------------------+
+/// | UninitializedUnwindContext |<---------------.
+/// +----------------------------+                |
+///             |                                 |
+///             |                                 |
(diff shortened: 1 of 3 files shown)


src/cfi.rs
```diff
@@ -454,12 +456,734 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
...
+    /// The rule at index `i` of the slice is the rule for register `i`. The
+    /// slice is **not** guaranteed to have length equal to the number of
+    /// registers on the target architecture. It is only as long as the largest
+    /// numbered register for which a recovery rule is defined in this row. All
+    /// others are implied to be `Undefined`.
+    ///
+    /// The register number mapping is architecture dependent. For example, in
+    /// the x86-64 ABI the register number mapping is defined in Figure 3.36:
+    ///
+    /// > Figure 3.36: DWARF Register Number Mapping
```
fitzgen: In practice, I think library consumbers need these register number mappings to make heads and tails of the resulting unwind table rows.

Should we make a `register_number_mapping` module exposing constant structs like `register_number_mapping::x86_64`?

src/cfi.rs
```diff
@@ -454,12 +456,734 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
...
+#[derive(Clone, Debug)]
+enum InitialRules<'input, Endian>
+    where Endian: Endianity
+{
+    // Don't access data in this vec! We just keep it around to reuse the
+    // allocation.
+    Uninitialized(Vec<RegisterRule<'input, Endian>>),
+    // This is safe to access, though.
+    Initialized(Vec<RegisterRule<'input, Endian>>),
+}
```
philipc: I'm not sure that this enum is carrying its weight. It's not necessary to make the module API safe to use (`InitializedUnwindContext` and `UninitializedUnwindContext` are enough for that), and it's creating a lot of complexity regarding `UnwindContext::initial_rules`.

src/cfi.rs
```diff
@@ -454,12 +456,734 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
...
+    /// `InitializedUnwindContext`.
+    pub fn initialize(mut self,
+                      cie: &CommonInformationEntry<'input, Endian>)
+                      -> Result<InitializedUnwindContext<'input, Endian>> {
+        assert!(self.0.is_uninitialized());
+
+        {
+            let mut table = UnwindTable::new_internal(&mut self.0, cie, None);
+            while let Some(_) = try!(table.next_row()) {}
+        }
```
philipc: So we're evaluating the CIE initial_instructions for every FDE? It seems like avoiding that would be an important optimization if you want to evaluate instructions for multiple FDEs, and the current API doesn't let the user avoid it.

philipc:
Continuing from my other two comments... can we make the API usage something like this:
````
let ctx = UnwindContext::new(&cie);
let mut table = UnwindTable::new(&ctx, &fde);
````

So `UnwindContext` is always initialized, and reusable for multiple `UnwindTable`s, but the borrow checker ensures only one table can use it at a time.

src/cfi.rs
philipc: Need it somewhere, not sure if it belongs in gimli or not. I'd wait until we have somewhere to use it and see where it fits in there.

fitzgen:
> Continuing from my other two comments... can we make the API usage something like this:
> 
>     let ctx = UnwindContext::new(&cie);
>     let mut table = UnwindTable::new(&ctx, &fde);
> 
> So UnwindContext is always initialized, and reusable for multiple UnwindTables, but the borrow checker ensures only one table can use it at a time.

I assume you meant `&mut ctx` as the first arg to `UnwindTable::new`? Would doing a `reset` involve passing a new CIE to reset the context to?

> So we're evaluating the CIE initial_instructions for every FDE? It seems like avoiding that would be an important optimization if you want to evaluate instructions for multiple FDEs, and the current API doesn't let the user avoid it.

Yes, I thought about how to do this, and the best I could think of was to `Clone` the initialized context after running the CIE's initial instructions. There is no way to reset to the initialized state without keeping a hidden copy of the stack, which itself is equivalent to just cloning.

> I'm not sure that this enum is carrying its weight. It's not necessary to make the module API safe to use (InitializedUnwindContext and UninitializedUnwindContext are enough for that), and it's creating a lot of complexity regarding UnwindContext::initial_rules.

Agreed, this was a stretch.

fitzgen:
(BTW, I'm at Strange Loop right now, so don't expect quick replies for the next few days)

fitzgen:
> Ok, I wasn't sure if step 1 of section 6.4.3 required that the stack be kept in addition to the registers. 

Yeah it isn't 100% clear...

> However, I'm also not sure how much the CIE is shared between FDEs. Is the same CIE likely to be used for more than one function? If not, then there won't be any reuse possible anyway.

From what I have seen, there is a fair bit of sharing, but it is like two or three instructions and some nop padding. I haven't seen any intense CFI programs in the wild (although I'm sure some do exist).

> I didn't think that through very well :) My goal was to avoid things be in an uninitialized state, but now I don't think that's going to be possible if you want to reuse stuff in a loop... it would have to be declared outside the loop and be uninitialized on the first time through. So let's go with how you have it and see how usable it is in practice.

SGTM

PR #109, opened 2016-09-15
Display file attribute values in dwarfdump

philipc:
Also some other small fixes.

### DIFF
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -165,11 +165,8 @@ impl Unit {
             _ => return None,
         };
 
-        let comp_dir = match entry.attr_value(gimli::DW_AT_comp_dir) {
-                Some(gimli::AttributeValue::String(dir)) => Some(dir),
-                Some(gimli::AttributeValue::DebugStrRef(offset)) => debug_str.get_str(offset).ok(),
-                _ => None,
-            }
+        let comp_dir = entry.attr(gimli::DW_AT_comp_dir)
+            .and_then(|attr| attr.string_value(debug_str))
             .map(|dir| dir.to_string_lossy().into_owned());
 
         Some(Unit {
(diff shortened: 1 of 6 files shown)


examples/dwarfdump.rs
```diff
@@ -134,24 +134,26 @@ fn dump_info<Endian>(file: &object::File,
...
+            dump_entries(offset,
+                         unit.entries(&abbrevs),
                          unit.address_size(),
                          unit.format(),
                          debug_line,
                          debug_loc,
                          debug_ranges,
                          debug_str,
                          flags);
+            offset += unit.length_including_self();
```
fitzgen: I think the offset stuff should be moved within the unit eventually, but doesn't need to be now.

examples/addr2line.rs
```diff
@@ -165,10 +165,9 @@ impl Unit {
...
 
-        let comp_dir = match entry.attr_value(gimli::DW_AT_comp_dir) {
-                Some(gimli::AttributeValue::String(dir)) => Some(dir),
-                Some(gimli::AttributeValue::DebugStrRef(offset)) => debug_str.get_str(offset).ok(),
-                _ => None,
+        let comp_dir = match entry.attr(gimli::DW_AT_comp_dir) {
+                Some(attr) => attr.string_value(debug_str),
+                None => None,
             }
             .map(|dir| dir.to_string_lossy().into_owned());
```
philipc: Thanks, I'm still not familiar with these chaining functions.

src/unit.rs
```diff
@@ -1529,6 +1529,17 @@ impl<'input, Endian> Attribute<'input, Endian>
             _ => return None,
         })
     }
+
+    /// Try to convert this attribute's value to a string.
```
philipc: I slightly reworded the short description, and used your text as the body. I wanted to keep the short description similar to the other functions.

examples/dwarfdump.rs
philipc: Ok. Doing it here is a bit ugly, but I wasn't sure if the offset stuff was going to be useful outside of this dwarfdump clone.

Issue #48, opened 2016-08-16
Re-evaluate deriving Copy


fitzgen:
It is easy to accidentally make copies when you didn't intend to, which leads to pretty surprising behavior until you figure out what's up. We should go through each place we derive Copy and re-evaluate whether that impl is providing value or not.

cc @kamalmarhubi

philipc:
Related to this, we should should ensure we derive Debug and Clone where appropriate. These can be useful to users of the library, and it's easy to miss adding them if we're not using them internally.

philipc:
According to the rust docs, we should normally derive Copy where possible: https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy. In what case was this a problem?

fitzgen:
Let's close this for now.

@kamalmarhubi: if you can provide more details (see previous comment), we can revisit.

PR #110, opened 2016-09-18
Add `#![deny(missing_debug_implementations)]` to the crate

fitzgen:
Also, fill in the missing Debug implementations. Also, fill in `Clone` implementations where possible.

Unsure whether the `Debug` trait bound on `EvaluationContext` is the right way to handle this... Open to alternatives.

See issue #48.

### DIFF
--- a/src/aranges.rs
+++ b/src/aranges.rs
@@ -90,6 +90,7 @@ impl Ord for ArangeEntry {
     }
 }
 
+#[derive(Clone, Debug)]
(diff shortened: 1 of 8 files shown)


fitzgen:
Unfortunately, there is no `#![deny(missing_clone_implementations)]`.

philipc:
> Unsure whether the Debug trait bound on EvaluationContext is the right way to handle this... Open to alternatives

There are a few traits in libstd that do the same, so seems ok to me.

PR #111, opened 2016-09-19
line: handle comp_dir and name

### LINKED ISSUE #59: Extend `FileEntry::directory` to return the compilation's current directory as `&CStr` instead of `None`
@fitzen suggested:

> impl FileEntry {
>     pub fn directory(&self, header: &LineNumberProgramHeader) -> &'input ffi::CStr;
> }

> Although we need to handle the index == 0 case, which is supposed to mean the compilation's current directory, which we get from the CU DIE... Maybe it should take that as input, too...

philipc:
Handle the interpretation of 0 indices for directories and files.  We
do this by storing the compilation unit directory and file references
inside the header struct.  This will be compatible with DWARF 5, which
stores copies of these values in the .debug_line section.

Fixes #59

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -118,7 +118,7 @@ fn bench_parsing_line_number_program_opcodes(b: &mut test::Bencher) {
     let debug_line = DebugLine::<LittleEndian>::new(&debug_line);
 
     b.iter(|| {
-        let header = debug_line.header(OFFSET, ADDRESS_SIZE)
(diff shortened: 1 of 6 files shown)


philipc:
I think the higher level semantic type for the compilation unit should do the attribute parsing, and have a `line_rows` method that passes the attribute values to the line module (ie the same as what the `Unit` struct in addr2line.rs does).

PR #118, opened 2016-09-20
Add the `DebugFrame::unwind_info_for_address` method

### LINKED ISSUE #117: Add an easy-mode function to get the unwind info for a given address
`DebugFrame::unwind_info_for_address(...)`

fitzgen:
The `unwind_info_for_address` method provides sane default behavior for quickly
and easily getting the unwind information for a given address for folks who
don't want to do anything too crazy like cache parsed CIEs and FDEs or CFI
evaluation results.

Fixes #117.

r? @philipc

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -63,6 +63,78 @@ impl<'input, Endian> DebugFrame<'input, Endian>
         let (_, entry) = try!(CommonInformationEntry::parse(input));
         Ok(entry)
     }
+
+    /// Find the frame unwind information for the given address.
+    ///
+    /// If found, the unwind information is returned along with the reset
+    /// context in the form `Ok((unwind_info, context))`. If not found,
+    /// `Err(gimli::Error::NoUnwindInfoForAddress)` is returned. If parsing or
+    /// CFI evaluation fails, the error is returned.
+    ///
+    /// ```
+    /// use gimli::{DebugFrame, NativeEndian, UninitializedUnwindContext};
+    ///
+    /// # fn foo() -> gimli::Result<()> {
+    /// # let read_debug_frame_section = || unimplemented!();
+    /// // Get the `.debug_frame` section from the object file.
+    /// let debug_frame = DebugFrame::<NativeEndian>::new(read_debug_frame_section());
+    ///
+    /// # let get_frame_pc = || unimplemented!();
+    /// // Get the address of the PC for a frame you'd like to unwind.
+    /// let address = get_frame_pc();
+    ///
+    /// // This context is reusable, which cuts down on heap allocations.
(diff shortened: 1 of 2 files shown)


src/cfi.rs
```diff
@@ -63,6 +63,78 @@ impl<'input, Endian> DebugFrame<'input, Endian>
...
+    /// let debug_frame = DebugFrame::<NativeEndian>::new(read_debug_frame_section());
+    ///
+    /// # let get_frame_pc = || unimplemented!();
+    /// // Get the address of the PC for a frame you'd like to unwind.
+    /// let address = get_frame_pc();
+    ///
+    /// // This context is reusable, which cuts down on heap allocations.
+    /// let ctx = UninitializedUnwindContext::new();
+    ///
+    /// let (unwind_info, ctx) = try!(debug_frame.unwind_info_for_address(ctx, address));
```
philipc: Does the borrow checker allow the following syntax? I'm worried all the moves are going to hurt; we can reevaluate once there is benchmark.
```
let mut ctx = UninitializedUnwindContext::new();
let unwind_info = try!(debug_frame.unwind_info_for_address(&mut ctx, address));
```
Also, I like that you're using `try!` in this example. We should update the other examples to do this.

src/cfi.rs
```diff
@@ -63,6 +63,78 @@ impl<'input, Endian> DebugFrame<'input, Endian>
...
+         address: u64)
+         -> Result<(UnwindTableRow<'input, Endian>, UninitializedUnwindContext<'input, Endian>)> {
+        let mut target_fde = None;
+
+        let mut entries = self.entries();
+        while let Some(entry) = try!(entries.next()) {
+            match entry {
+                CieOrFde::Cie(_) => continue,
+                CieOrFde::Fde(partial) => {
+                    let fde = try!(partial.parse(|offset| self.cie_from_offset(offset)));
```
philipc: This is needlessly parsing some CIEs.  Maybe the FDE should only store the offset, and we parse the CIE once we know if we want to evaluate it.

fitzgen: Unfortunately, you need to parse the CIE to figure out the FDE's range of addresses :(

We could build up a map of offset -> CIE while iterating entries, because it seems likely that the CIE will come before the FDE, but I'll leave that to when we have benches and can properly investigate.

src/cfi.rs
fitzgen: It wouldn't because uninitialized -> initialized is consuming `self`, so we can't have a reference.

Agreed regarding `try!` -- we also probably don't need actual tests in the doc examples like how a lot of the unit stuff has (it was convenient at the time) because there's so much noise and we really just want to show how to use the APIs.

PR #120, opened 2016-09-20
Unit offset fixes

### LINKED ISSUE #63: UnitHeader::size_of_header is invalid for type units
This breaks `is_valid_offset`, `range`, etc.

Need to store something in `UnitHeader` (eg the header size, or the offset of the first entry), and change `size_of_header` to take `&self`.

philipc:
Fixes #63, plus a few other fixes related to offsets.

Not sure about the addition of `CompilationUnitHeader`. It's simply to get the right type for unit.offset().


src/unit.rs
```diff
@@ -127,8 +130,9 @@ impl<'input, Endian> CompilationUnitHeadersIter<'input, Endian>
         if self.input.is_empty() {
             Ok(None)
         } else {
-            match CompilationUnitHeader::parse(self.input) {
+            match CompilationUnitHeader::parse(self.input, self.offset) {
                 Ok((rest, header)) => {
+                    self.offset.0 += self.input.len() as u64 - rest.len() as u64;
```
fitzgen: Did you happen to bench before and after adding offsets into units? I expect it should fall within the noise, and I think we want this change either way, but I am a little curious.

examples/dwarfdump.rs
```diff
@@ -138,13 +138,12 @@ fn dump_info<Endian>(file: &object::File,
...
         let debug_info = gimli::DebugInfo::<Endian>::new(&debug_info);
 
-        let mut offset = 0;
         let mut iter = debug_info.units();
         while let Some(unit) = iter.next().expect("Should parse compilation unit") {
             let abbrevs = unit.abbreviations(debug_abbrev)
                 .expect("Error parsing abbreviations");
 
-            dump_entries(offset,
+            dump_entries(unit.offset().0,
```
fitzgen: Nice.

fitzgen:
> Not sure about the addition of CompilationUnitHeader. It's simply to get the right type for unit.offset().

I like it 👍

src/unit.rs
philipc: I wouldn't expect this to affect the bench. I did run the bench, but any change wasn't large enough to be able to measure reliably.

PR #125, opened 2016-09-21
Changes offsets to usize

### LINKED ISSUE #119: Should the offset newtypes be usize?
The values of DebugInfoOffset etc nearly always need to be used as usize, not u64. Perhaps we should change them to usize?

Note that `offset.0 as usize` will truncate if usize is smaller than u64, so if we validate this when parsing then I think overall this will be safer. I can't see any way to support offsets that are larger than the host's usize anyway, so I think failing when parsing is all we can do.

philipc:
Fixes #119

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -3,13 +3,14 @@ use endianity::{Endianity, EndianBuf};
 use fallible_iterator::FallibleIterator;
 use parser::{Error, Format, Result, parse_address, parse_initial_length, parse_length_uleb_value,
              parse_null_terminated_string, parse_signed_leb, parse_signed_lebe, parse_u8,
-             parse_u8e, parse_u16, parse_u32, parse_unsigned_leb, parse_unsigned_lebe, parse_word};
+             parse_u8e, parse_u16, parse_u32, parse_unsigned_leb, parse_unsigned_lebe, parse_word,
+             u64_to_offset};
 use std::marker::PhantomData;
 use std::str;
 
 /// An offset into the `.debug_frame` section.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub struct DebugFrameOffset(pub u64);
+pub struct DebugFrameOffset(pub usize);
 
 /// The `DebugFrame` struct contains the source location to instruction mapping
 /// found in the `.debug_frame` section.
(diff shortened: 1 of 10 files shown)


src/parser.rs
```diff
@@ -575,3 +493,89 @@ pub fn parse_length_uleb_value<Endian>(input: EndianBuf<Endian>)
     let (rest, len) = try!(parse_unsigned_leb(input.into()));
     take(len as usize, EndianBuf::new(rest))
 }
+
+#[cfg(test)]
+mod tests {
```
fitzgen: Thanks!

src/parser.rs
```diff
@@ -496,12 +496,16 @@ pub fn parse_length_uleb_value<Endian>(input: EndianBuf<Endian>)
 
 #[cfg(test)]
 mod tests {
+    extern crate test_assembler;
```
fitzgen: \o/

src/cfi.rs
```diff
@@ -471,23 +473,22 @@ impl<'input, Endian> FrameDescriptionEntry<'input, Endian>
         where F: FnMut(DebugFrameOffset) -> Result<CommonInformationEntry<'input, Endian>>
     {
         let (rest_rest, (length, format, cie_pointer, rest)) = try!(parse_cfi_entry_common(input));
+        if is_cie_id(format, cie_pointer) {
+            return Err(Error::NotCiePointer);
```
fitzgen: Good catch!

src/parser.rs
```diff
@@ -348,6 +353,40 @@ pub fn parse_u32_as_u64<Endian>(input: EndianBuf<Endian>) -> Result<(EndianBuf<E
...
 
+/// Convert a `u64` to a `usize` and return it.
+#[doc(hidden)]
+#[inline]
+pub fn u64_to_offset(offset64: u64) -> Result<usize> {
+    let offset = offset64 as usize;
+    if offset as u64 == offset64 {
+        Ok(offset)
+    } else {
+        Err(Error::UnsupportedOffset)
```
fitzgen: We should have a test for the error branch. I don't see one in this commit, maybe in a follow up I haven't got to yet?

If not, it can be a follow up, if you'd prefer.

philipc: I'll try to add an i386 test in a followup.

PR #129, opened 2016-09-22
Fix offset tests for i686

philipc:
Note: I couldn't see any way to do a travis i686 build using travis-cargo. Might need to change to custom build scripts if we want that.


fitzgen:
We could maybe leverage generics to make a checked T -> U conversion, and could therefore test u64 -> u32 directly. I think generics are problematic with numbers, however...

fitzgen:
Thanks for following up!

PR #131, opened 2016-09-27
Run rustfmt in travis-ci

philipc:
This lets me know when I forgot to run rustfmt, or if I need to update rustfmt.

### DIFF
--- a/format
+++ b/format
@@ -29,5 +29,5 @@ echo "$0: using rustfmt=$RUSTFMT"
 cd $(dirname $0)
 RUST_FILES=$(git ls-files | grep '\.rs')
 
-$RUSTFMT -v --write-mode replace $RUST_FILES 2>&1 \
+$RUSTFMT -v --write-mode overwrite $RUST_FILES 2>&1 \
     | xargs -I '{}' echo "rustfmt: {}"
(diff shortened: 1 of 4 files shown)


fitzgen:
I'm not sure that we need to actually block PRs on forgetting to run `rustfmt` -- is it not ok to just let it be fixed the next time we run `rustfmt`?

FWIW, I run `rustfmt` as part of my  compile-and-test command (`cargo test && cargo doc && ./format`) so things should get caught pretty quickly.

Aside: what version of `rustfmt` do you have? Mine is 0.6.0, and the changes here seem to suggest a different format than mine is applying for the CIE functions.

philipc:
I wasn't intending to block PRs, just wanted something to let me know it needs to be run, so that I don't get unrelated formatting changes mixed in with my current work. I'm fine with dropping that commit though.

I updated my rustfmt to the latest version (0.6.2), since this PR caused travis to install the newest version every time. I don't remember what my previous version was, but it wanted to reformat `parse_cfi_entry` too.

fitzgen:
Ok, I've updated to rustfmt 0.6.2.

I'd prefer we not run rustfmt on travis because it makes the pass/fail noisier: if we aren't going to block PRs on rustfmt cleanliness, then we need to manually check the travis results to see if it was a real failure or just rustfmt.

Other two commits LGTM, though.

Between the u64 -> usize commit and this formatting, my `.eh_frame` branch is going to be fun to rebase :P

PR #134, opened 2016-09-30
Unit fixes

### LINKED ISSUE #132: Fix AttributeValue::value() for DW_AT_data_member_location
In DWARF version 2/3, `DW_FORM_data4`/`DW_FORM_data8` for a `DW_AT_data_member_location` attribute means a `.debug_loc` offset. But in DWARF version 4, it means a constant address offset.  So `AttributeValue::value()` needs to interpret these forms based on the DWARF version. However, we don't currently have access to the DWARF version in this function.

philipc:
Two fixes in unit.rs.  Also converted most of the unit.rs tests to test-assembler to help in writing tests for those fixes.

Fixes #132

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -166,6 +166,12 @@ pub struct CompilationUnitHeader<'input, Endian>
 impl<'input, Endian> CompilationUnitHeader<'input, Endian>
     where Endian: Endianity
 {
+    /// Return the serialized size of the compilation unit header for the given
+    /// DWARF format.
+    pub fn size_of_header(format: Format) -> usize {
+        UnitHeader::<Endian>::size_of_header(format)
+    }
+
     /// Get the offset of this compilation unit within the .debug_info section.
     pub fn offset(&self) -> DebugInfoOffset {
         self.offset
@@ -398,7 +404,7 @@ impl<'input, Endian> UnitHeader<'input, Endian>
         }
     }
 
-    /// Return the serialized size of the compilation unit header for the given
+    /// Return the serialized size of the common unit header for the given
     /// DWARF format.
     pub fn size_of_header(format: Format) -> usize {
         let unit_length_size = Self::size_of_unit_length(format);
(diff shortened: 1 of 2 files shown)


fitzgen:
Travis CI had some breakage with rust when they updated their rustup.sh. I think this PR caught that. It's since been fixed, and I've restarted the build.

src/unit.rs
```diff
@@ -158,90 +154,6 @@ impl<'input, Endian> FallibleIterator for CompilationUnitHeadersIter<'input, End
     }
 }
 
-#[test]
```
fitzgen: Nice, thanks

src/unit.rs
```diff
@@ -2403,21 +2403,144 @@ fn parse_type_unit_header<Endian>(input: EndianBuf<Endian>,
...
+                Format::Dwarf32 => self.L32(offset as u32),
+                Format::Dwarf64 => self.L64(offset as u64),
+            }
+        }
+
+        fn uleb(self, val: u64) -> Self {
+            let mut buf = Vec::new();
+            let written = leb128::write::unsigned(&mut buf, val).unwrap();
+            self.append_bytes(&buf[0..written])
+        }
```
fitzgen: My `.eh_frame` branch will introduce `test_util.rs` which has a trait with `uleb` and `sleb` methods. Once I get that merged, we will be able to make this trait simply require that one, and we can remove the duplication.

src/unit.rs
```diff
@@ -2540,86 +2668,41 @@ mod tests {
...
+                debug_abbrev_offset: DebugAbbrevOffset(0x08070605),
+                address_size: 4,
+                format: Format::Dwarf32,
+                entries_buf: EndianBuf::new(expected_rest),
+            },
+            offset: DebugInfoOffset(0),
+        };
+        let section = Section::with_endian(Endian::Little)
+            .comp_unit(&mut unit64)
+            .comp_unit(&mut unit32);
```
fitzgen: Awesome, thanks so much for doing all this!

PR #135, opened 2016-10-02
Extend constant tables with latest and extension values.

borman:
Used definitions from LLVM trunk:
- http://llvm.org/viewvc/llvm-project/llvm/trunk/include/llvm/Support/Dwarf.h?revision=278940&view=markup
- http://llvm.org/viewvc/llvm-project/llvm/trunk/include/llvm/Support/Dwarf.def?revision=278940&view=markup


fitzgen:
Thanks!

PR #139, opened 2016-10-06
Dwarfdump improvements

philipc:
We're pretty close to libdwarf for `.debug_info` now. Most of the differences are in error handling, which I probably won't try to make the same.


fitzgen:
Great! :)

PR #141, opened 2016-10-06
next_sibling() improvements

philipc:


src/unit.rs
```diff
@@ -557,7 +557,6 @@ pub struct DebuggingInformationEntry<'input, 'abbrev, 'unit, Endian>
     offset: UnitOffset,
     attrs_slice: &'input [u8],
     after_attrs: Cell<Option<&'input [u8]>>,
-    code: u64,
```
fitzgen: Nice.

src/unit.rs
```diff
@@ -2095,36 +2096,47 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
...
-                    return Ok(self.current());
+        if self.current().is_none() {
+            // We're already at the null for the end of the sibling list.
+            return Ok(None);
+        }
+
+        // Loop until we find an entry at the current level.
+        let mut depth = 0;
+        loop {
+            if self.current().map(|entry| entry.has_children()) == Some(true) {
```
fitzgen: `blah.unwrap_or(false)` can replace `blah == Some(true)` but I don't feel strongly about it.

philipc:
> Do you have #[bench] numbers for checking has_children() before iterating attrs? I guess we don't really have any benches that exercise these code paths heavily -- it might would be nice to add some eventually.

It doesn't show in the current benchmarks, because none of them call `next_sibling` . I am working on a tree-like iterator, and the `has_children` check made a significant difference there, because I'm calling `next_sibling` for leaf entries too. The `has_children` check is basically free, because we need to check it anyway to see if the depth is going to change, so it won't hurt the non-leaf entries either.

PR #143, opened 2016-10-07
Optimize for sequential abbreviation codes

philipc:
Normally abbreviations will have sequential codes starting at 1. Optimize for this by storing them in a Vec.

    Before:
    test bench_parsing_debug_abbrev                ... bench:      19,021 ns/iter (+/- 794)
    test bench_parsing_debug_info                  ... bench:   2,737,211 ns/iter (+/- 92,580)

    After:
    test bench_parsing_debug_abbrev                ... bench:      14,844 ns/iter (+/- 628)
    test bench_parsing_debug_info                  ... bench:   2,326,573 ns/iter (+/- 29,075)

This is a bit of a trade-off if the codes aren't sequential, but I think the cost is small relative to the cost of hashing, and the potential gain is large.

### DIFF
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -357,81 +357,78 @@ impl AttributeSpecification {
 }
 
 #[cfg(test)]
-mod tests {
+pub mod tests {
+    extern crate test_assembler;
+
     use super::*;
     use constants;
-    use parser::Error;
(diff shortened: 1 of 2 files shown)


philipc:
I probably should add some tests for all the edge cases in this too...

PR #145, opened 2016-10-07
Add a CONTRIBUTING.md file

fitzgen:
r? @philipc 

Anything I'm missing here that you'd like to see added?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,107 @@
+# Contributing to `gimli`
+
+Hi! We'd love to have your contributions! If you want help or mentorship, reach
+out to us in a GitHub issue, or ping `fitzgen` in `#rust` on `irc.mozilla.org`.
+
+* [Code of Conduct](#coc)
+* [Filing an Issue](#issues)
+* [Building `gimli`](#building)
+* [Testing `gimli`](#testing)
+  * [Test Coverage](#coverage)
+  * [Using `test-assembler`](#test-assembler)
+* [Benchmarking](#benchmarking)
+* [Style](#style)
+
+## <a id="coc"></a> Code of Conduct
+
+We abide by the
+[Rust Code of Conduct](https://www.rust-lang.org/en-US/conduct.html) and ask
+that you do as well.
+
+## <a id="issues"></a> Filing an Issue
+
+Think you've found a bug? File an issue! To help us understand and reproduce the
+issue, provide us with:
+
+* The (preferably minimal) test case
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
...
+
+The ideal we aim to reach is having our unit tests exercise every branch in
+`gimli`. We allow an exception for branches which propagate errors inside a
+`try!(..)` invocation, but we *do* want to exercise the original error paths.
+
+Pull requests adding new code should ensure that this ideal is met.
+
+At the time of writing we have 94% test coverage according to our coveralls.io
+continuous integration. That number should generally stay the same or go up ;)
+This is a bit subjective, because -.001% is just noise and doesn't matter.
```
philipc: If you have `kcov` installed under linux, then you can generate the coverage results using `./coverage` in the root of the repository, and view them at target/kcov/index.html. Otherwise you can create a pull request and view the coverage results on coveralls.io.

CONTRIBUTING.md
```diff
@@ -0,0 +1,107 @@
...
+
+The benchmarks require nightly `rustc`, so use `rustup`:
+
+```
+$ rustup run nightly cargo bench
+```
+
+We aim to be the fastest DWARF library. Period.
+
+Please provide before and after benchmark results with your pull requests. You
```
philipc: But only needed if the PR affects performance?

fitzgen: Ideally, we would have a CI bot to run before and after benches for us. That would be super sweet.

I don't think it hurts to encourage more benching than is strictly necessary, especially for newcomers before they've gotten a feel for how we do things here.

CONTRIBUTING.md
fitzgen: Will add, thanks!

CONTRIBUTING.md
fitzgen: Also, should we link to your dwarf-bench here? Is that something you'd like to continue building in the future? I'd like to do some more in depth cross-library benchmarking sometime (*handwaves*) in the future.

philipc: dwarf-bench isn't useful to link to yet. It's hard to use and the results probably aren't that meaningful yet. For cross-library comparisons, I think it's going to be more useful to benchmark things that solve a well-defined problem with a verifiable result, so that we know if we're comparing the same thing. So current status is that it was useful to get ballpark figures, but it's going to need a rethink to get anything more.

PR #148, opened 2016-10-10
Use a `Vec<(number, rule)>` to keep track of register rules

fitzgen:
What we were doing was constructing an array where each index I contained the
recovery rule for register I. We usually have very few register rules: only one
or two. The result was that we would allocate more space than we needed and fill
it with undefined register rules. Avoiding that is a nice little speedup, as
shown in the `cargo benchcmp` results below:

```
 name                                                       before ns/iter  after ns/iter  diff ns/iter   diff %
 cfi::eval_longest_fde_instructions_new_ctx_everytime       539             419                    -120  -22.26%
 cfi::eval_longest_fde_instructions_same_ctx                504             396                    -108  -21.43%
 cfi::iterate_entries_evaluate_every_fde                    2,247,613       1,735,106          -512,507  -22.80%
```

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1768,6 +1768,10 @@ impl<'input, Endian> RegisterRuleMap<'input, Endian>
     fn clear(&mut self) {
         self.rules.clear();
     }
+
+    fn iter<'me>(&'me self) -> RegisterRuleIter<'me, 'input, Endian> {
+        RegisterRuleIter(self.rules.iter())
+    }
 }
 
 impl<'a, 'input, Endian> FromIterator<&'a (u64, RegisterRule<'input, Endian>)> for RegisterRuleMap<'input, Endian>
@@ -1809,6 +1813,25 @@ impl<'input, Endian> PartialEq for RegisterRuleMap<'input, Endian>
 }
 
 impl<'input, Endian> Eq for RegisterRuleMap<'input, Endian> where Endian: Endianity {}
+
+/// An unordered iterator for register rules.
+#[derive(Debug, Clone)]
+pub struct RegisterRuleIter<'iter, 'input, Endian>(::std::slice::Iter<'iter,
+                                                                        (u64,
+                                                                         RegisterRule<'input,
+                                                                                      Endian>)>)
+    where Endian: 'iter + Endianity,
+          'input: 'iter;
+
+impl<'iter, 'input, Endian> Iterator for RegisterRuleIter<'iter, 'input, Endian>
+    where Endian: Endianity
+{
+    type Item = &'iter (u64, RegisterRule<'input, Endian>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.next()
+    }
+}
 
 /// A row in the virtual unwind table that describes how to find the values of
 /// the registers in the *previous* frame for a range of PC addresses.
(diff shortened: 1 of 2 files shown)


src/cfi.rs
```diff
@@ -1720,6 +1712,87 @@ impl<'input, 'cie, 'fde, 'ctx, Endian, Section> UnwindTable<'input,
...
+    where Endian: Endianity
+{
+    fn get(&self, register: u64) -> Option<&RegisterRule<'input, Endian>> {
+        self.rules.iter()
+            .find(|rule| {
+                rule.0 == register
+            })
+            .map(|rule| {
+                &rule.1
+            })
```
philipc: rustfmt will probably want to format this differently.

src/cfi.rs
```diff
@@ -1825,8 +1892,8 @@ impl<'input, Endian> UnwindTableRow<'input, Endian>
     /// >   <tr><td>Vector Mask Registers 0–7</td>        <td>118-125</td> <td>%k0–%k7</td></tr>
     /// >   <tr><td>Reserved</td>                         <td>126-129</td> <td></td></tr>
     /// > </table>
-    pub fn registers(&self) -> &[RegisterRule<'input, Endian>] {
```
philipc: Not sure what the common use case for this data is, but I suspect there will be cases where the user needs to handle every rule. So maybe we should still provide an API that returns the array. Otherwise I think iterating over the rules will be `O(n^2)`.

src/cfi.rs
```diff
@@ -1720,6 +1712,87 @@ impl<'input, 'cie, 'fde, 'ctx, Endian, Section> UnwindTable<'input,
...
+        }
+        rules
+    }
+}
+impl<'input, Endian> PartialEq for RegisterRuleMap<'input, Endian>
+    where Endian: Endianity
+{
+    fn eq(&self, rhs: &Self) -> bool {
+        let undef = RegisterRule::Undefined;
+        let undef = &undef;
```
philipc: Nit: you could combine these two lines.

src/cfi.rs
```diff
@@ -1720,6 +1712,87 @@ impl<'input, 'cie, 'fde, 'ctx, Endian, Section> UnwindTable<'input,
...
+    fn from_iter<T>(iter: T) -> RegisterRuleMap<'input, Endian>
+        where T: IntoIterator<Item=&'a (u64, RegisterRule<'input, Endian>)>
+    {
+        let mut rules = RegisterRuleMap::default();
+        for &(reg, ref rule) in iter {
+            rules.set(reg, rule.clone());
+        }
+        rules
+    }
+}
```
philipc: Nit: add a blank line.

src/cfi.rs
```diff
@@ -4885,4 +4941,31 @@ mod tests {
...
+                                                   (2, RegisterRule::Offset(1))].iter().collect();
+        let map4: RegisterRuleMap<LittleEndian> = [(3, RegisterRule::Offset(1)),
+                                                   (0, RegisterRule::SameValue)].iter().collect();
+        assert!(map3 != map4);
+        assert!(map4 != map3);
+
+        // One has undefined, other implicitly has undefined.
+        let mut map5 = RegisterRuleMap::<LittleEndian>::default();
+        map5.set(0, RegisterRule::SameValue);
+        map5.set(0, RegisterRule::Undefined);
```
philipc: Not sure I like that `set` treats undefined differently based on whether there was already a value. I can see why you've done it though. If we're going to do that, I think we should hide the difference from the user, so we should at least make `get` return `None` if the rule is undefined, whether implicit or not.

src/cfi.rs
fitzgen: Ah yes, a stepping debugger will certainly want to iterate all registers.

src/cfi.rs
fitzgen: Good catch. I think it makes sense to discard the option and return `RegisterRule::Undefined` if we don't have a rule: that's what we end up doing by the time we get to a public API anyways.

fitzgen: Also, `swap_remove` could allow us to never keep `Undefined` in the vec.

PR #150, opened 2016-10-15
Implement DW_OP_implicit_pointer and DW_OP_entry_value

philipc:
- move op.rs tests into submodule
- convert operation parsing tests to use `test_assembler`
- implement DW_OP_implicit_pointer and DW_OP_entry_value

The implementation for the new operations is minimal: it offloads all the real work to the caller. We probably want to add more support in future to make it easier for users to handle these though.

I didn't convert the operation evaluation tests to use `test_assembler` because @tromey had already written some code that does basically the same thing, so there wouldn't be much benefit to doing so (other than consistency).

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -193,6 +193,15 @@ pub enum Operation<'input, Endian>
     },
     /// Represents `DW_OP_stack_value`.
     StackValue,
+    /// Represents `DW_OP_implicit_pointer`. The object is a pointer to
+    /// a value which has no actual location, such as an implicit value or
+    /// a stack value.
+    ImplicitPointer {
+        /// The `.debug_info` offset of the value that this is an implicit pointer into.
+        value: DebugInfoOffset,
+        /// The byte offset into the value that the implicit pointer points to.
+        byte_offset: i64,
+    },
 }
 
 /// A single location of a piece of the result of a DWARF expression.
@@ -220,6 +229,13 @@ pub enum Location<'input> {
     Bytes {
         /// The value.
         value: &'input [u8],
+    },
+    /// The piece is a pointer to a value which has no actual location.
+    ImplicitPointer {
+        /// The `.debug_info` offset of the value that this is an implicit pointer into.
+        value: DebugInfoOffset,
+        /// The byte offset into the value that the implicit pointer points to.
+        byte_offset: i64,
     },
 }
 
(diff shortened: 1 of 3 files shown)


src/op.rs
```diff
@@ -221,6 +230,13 @@ pub enum Location<'input> {
         /// The value.
         value: &'input [u8],
     },
+    /// The piece is a pointer to a value which has no action location.
```
tromey: I think you probably meant s/action/actual/

src/op.rs
```diff
@@ -771,6 +787,16 @@ impl<'input, Endian> Operation<'input, Endian>
                 Ok((newbytes, Operation::ImplicitValue { data: data.into() }))
             }
             constants::DW_OP_stack_value => Ok((bytes, Operation::StackValue)),
+            constants::DW_OP_implicit_pointer |
+            constants::DW_OP_GNU_implicit_pointer => {
+                let (newbytes, value) = try!(parse_offset(bytes, format));
+                let (newbytes, byte_offset) = try!(parse_signed_lebe(newbytes));
+                Ok((newbytes,
+                    Operation::ImplicitPointer {
+                    value: DebugInfoOffset(value),
```
tromey: The indentation looks off here but I can't tell if that is github or just needing rustfmt.

src/op.rs
```diff
@@ -2485,13 +2518,30 @@ mod tests {
 
         check_eval_with_context(&program, Ok(&result), 8, Format::Dwarf64, context);
 
+        // Test `evaluate_entry_value` callback.
+        let program = [
+            Op(DW_OP_entry_value), Sleb(8), U64(0x12345678),
```
tromey: Pedantically I think this should be Uleb(8)

src/op.rs
philipc: Yep, thanks.

src/op.rs
philipc: Neither :) Looks like a rustfmt bug, where it's trying to block indent the struct members, but forgets that the struct is also visually indented for the enclosing tuple.

src/op.rs
philipc: Fixed, thanks.

philipc:
> This one is too big to see in github, but if it's just a pure move of the tests, then I think that's a good idea.

Yeah it is just a move.

PR #153, opened 2016-10-30
[DO NOT MERGE] Cursor rework

philipc:
These changes are still incomplete/buggy, but have a look through if you want.

My goal is to change `EntriesCursor::cached_current` to no longer be an `Option`. I think this simplifies the code, and it shows ~25% improvement for `bench_parsing_debug_info`. This simplification will also follow through to the `EntriesTree` API.

The meaning of this change is that cursor will always have a valid current value. The main implications of this are:
- the root entry is parsed in `UnitHeader::entries()`
- `next_dfs()` doesn't update `cached_current` when it reaches EOF
- `next_entry()` is no longer meaningful, because `cached_current` can no longer be a null entry. I'm not concerned about this though, because `EntriesTree` provides the behaviour that I previously wanted `next_entry` for.

I don't know what behaviour `next_sibling()` should have. Currently we leave the cursor pointing at a null entry after the last sibling, but that will no longer be possible. If it's left pointing at the last sibling, then there's not really any further meaningful operations that you can perform on the cursor, but I don't think it makes sense to advance to the next valid entry either. For my purposes, `EntriesTree` replaces `next_sibling()` for full tree traversals, so it's probably ok if `next_sibling()` can only be used for navigating one level of the tree.

Any opinions on whether this change is for the better?

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -56,7 +56,7 @@ fn bench_parsing_debug_info(b: &mut test::Bencher) {
             let abbrevs = unit.abbreviations(debug_abbrev)
                 .expect("Should parse abbreviations");
 
-            let mut cursor = unit.entries(&abbrevs);
+            let mut cursor = unit.entries(&abbrevs).expect("Should parse root entry");
             while let Some((_, entry)) = cursor.next_dfs().expect("Should parse next dfs") {
                 let mut attrs = entry.attrs();
                 while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
@@ -90,8 +90,7 @@ fn bench_parsing_debug_info_tree(b: &mut test::Bencher) {
 
 fn parse_debug_info_tree(mut iter: EntriesTreeIter<LittleEndian>) {
     {
-        let entry = iter.entry().expect("Should have current entry");
-        let mut attrs = entry.attrs();
+        let mut attrs = iter.entry().attrs();
         while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
             test::black_box(&attr);
         }
--- a/examples/addr2line.rs
+++ b/examples/addr2line.rs
@@ -137,10 +137,8 @@ impl<'input> Unit<'input> {
         where Endian: gimli::Endianity
     {
         let abbrev = header.abbreviations(*debug_abbrev).expect("Fail");
-        let mut entries = header.entries(&abbrev);
-        let (_, entry) = entries.next_dfs()
-            .expect("Should parse first entry OK")
-            .expect("And first entry should exist!");
+        let entries = header.entries(&abbrev).expect("Should parse first entry OK");
+        let entry = entries.current();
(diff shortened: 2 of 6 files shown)


fitzgen:
I'm tentatively positive &mdash; ~25% is nothing to sneeze at!

> I don't know what behaviour next_sibling() should have. Currently we leave the cursor pointing at a null entry after the last sibling, but that will no longer be possible. If it's left pointing at the last sibling, then there's not really any further meaningful operations that you can perform on the cursor, but I don't think it makes sense to advance to the next valid entry either. For my purposes, EntriesTree replaces next_sibling() for full tree traversals, so it's probably ok if next_sibling() can only be used for navigating one level of the tree.

So you're saying that we will no longer be able to advance a cursor to the last sibling, and then down that sibling's children? Or something else?

philipc:
> So you're saying that we will no longer be able to advance a cursor to the last sibling, and then down that sibling's children? Or something else?

No, I'm talking about when you've already called `next_sibling` to advance the cursor past the last sibling. The problem is that this change means the cursor can no longer point to the null entry that terminates the sibling list, so where should the cursor end up? If you leave it pointing at the last sibling, then subsequently calling `next_dfs` may return one of the last sibling's children, which is probably not what you want. If you move it to point to the next non-null entry, then the caller will have lost the information about the depth change that `next_dfs` normally gives you.

Unless you have better ideas for how to handle this, I think the change in this PR is a negative.

I'm also suspicious of how meaningful our benchmarks are for guiding optimizations. For example, commenting out `bench_parsing_debug_info_tree` can affect the result of `bench_parsing_debug_info`. I assume this is because LLVM changes its inlining decision based on how many times a function is called. Some of the 25% improvement I stated above was due to this.

I'm going to continue looking for ways to simplify this code though.

I've had a look at how gdb handles `DW_AT_sibling`, and it has a `skip_one_die` function than can end up pointing to a null entry. gdb does its DIE navigation using two basic functions:
- `peek_die_abbrev` which reads the abbreviation code
- `read_attribute` which reads one attribute

and two slightly higher level functions:
- `skip_children` which calls `peek_die_abbrev` and `skip_one_die` in a loop
- `skip_one_die` which calls `read_attribute` (handling any `DW_AT_sibling`) then `skip_children`

I like the simplicity of this, but it does put some of the complexity on the caller. eg, the caller needs to keep track of the nesting, and shouldn't call `skip_children` immediately after `peek_die_abbrev`.

fitzgen:
> No, I'm talking about when you've already called next_sibling to advance the cursor past the last sibling.

Ah, ok. I see.

> Unless you have better ideas for how to handle this, I think the change in this PR is a negative.

I don't. Introducing a boolean flag or putting the guts in an Option isn't an improvement.

> I'm also suspicious of how meaningful our benchmarks are for guiding optimizations.

:-/ Not sure what to do about this other than sprinkle more `#[inline]` where we are pretty sure inlining should happen (for some hand wavy, cross-your-fingers-its-correct definition of "pretty sure").

PR #156, opened 2016-11-11
Change entries_tree() to take an offset

philipc:
Also adds an `entries_from()` method that takes an offset, since we probably don't want to change the `entries()` API at this stage. I'm open to better names.


philipc:
Thanks, changed it to entries_at_offset.

PR #157, opened 2016-11-29
Derive Copy for Operation/Location/Piece

philipc:


philipc:
The linux nightly failure is a cargo bug, and the osx failures seem to be because travis already has python installed now. I'll try to fix that later.

PR #159, opened 2016-11-30
Allow user specified Error type for EvaluationContext

### LINKED ISSUE #158: Evaluation and EvaluationContext should allow a custom Error type
Users of `Evaluation` are currently constrained to picking an error value from `gimli::Error` for any errors that occur in the `EvaluationContext` callbacks. Instead, we should make these generic for any error type that `gimli::Error` can be converted to.

philipc:
Without this, I can't return meaningful errors from the callbacks.

Also use static dispatch for EvaluationContext callbacks.

Fixes #158

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -825,9 +825,9 @@ impl<'input, Endian> Operation<'input, Endian>
 
 /// A DWARF expression evaluation.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


philipc:
Same test failures as yesterday.

tromey:
This all looked great to me.  Thanks.

PR #160, opened 2016-11-30
Fix OSX Python install in CI

fitzgen:
I had to do this same fix in another project to keep OSX builds passing in CI.

r? @philipc


fitzgen:
The nightly linux failure is https://github.com/rust-lang/cargo/issues/3340, which I thought would be fixed by today's nightly, but maybe tomorrow's.

fitzgen:
Going to go ahead and merge this.

> delays in osx builds.

Yeah, pretty terrible...

PR #162, opened 2016-12-07
Make CFI evaluation signal safe

fitzgen:
See individual commits for details.

r? @philipc

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1,10 +1,12 @@
+use arrayvec::ArrayVec;
 use constants;
 use endianity::{Endianity, EndianBuf};
 use fallible_iterator::FallibleIterator;
 use parser::{Error, Format, Pointer, Result, parse_address, parse_encoded_pointer,
              parse_initial_length, parse_length_uleb_value, parse_null_terminated_string,
              parse_pointer_encoding, parse_signed_lebe, parse_u8, parse_u8e, parse_u16, parse_u32,
-             parse_u32_as_u64, parse_u64, parse_unsigned_lebe, u64_to_offset};
+             parse_u32_as_u64, parse_u64, parse_unsigned_lebe, parse_unsigned_leb_as_u8e,
+             u64_to_offset};
 use std::cell::RefCell;
 use std::fmt::Debug;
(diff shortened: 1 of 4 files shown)


src/cfi.rs
```diff
@@ -1271,9 +1282,18 @@ impl<'input, Endian, Section> UninitializedUnwindContext<'input, Endian, Section
 {
     /// Construct a new call frame unwinding context.
     pub fn new() -> UninitializedUnwindContext<'input, Endian, Section> {
-        UninitializedUnwindContext(UnwindContext::new())
+        UninitializedUnwindContext(Box::new(UnwindContext::new()))
```
philipc: Is this the only place that allocates now?

fitzgen:
Thanks for the review :)

> Did you measure how much difference the u8 register made? I think it's fine to do anyway, but I suspect it hasn't gained much due to alignment.

I didn't measure in isolation, unfortunately. I was just trying to get everything down to a somewhat reasonable size. I think we may want to make the whole storage generic in the future in case people want more stack size or something like that.

I was dismayed that from a glance that most places wouldn't benefit, but I think a few did.

/me shrugs

> We may be able to reduce RegisterRule size by changing Expression and ValExpression to store only offsets, and keep the common buffer pointer elsewhere.

Nice idea -- I'll file an issue to investigate further.

> FYI, libunwind has its own mempool routines that it uses to allocate the equivalent of UnwindTableRow. It also sets the number of registers based on architecture.

Interesting that they both pre-allocate memory and use mmap. I didn't think that mmap was signal safe, but perhaps I am wrong, or they don't do it during unwinding proper.

But yeah, I think down the line we want to be generic across storage for the register rules, cfa and stack (maybe make `UnwindContext` into a trait?).

src/cfi.rs
fitzgen: For CFI stuff, yes it should be.

philipc:
> I didn't think that mmap was signal safe, but perhaps I am wrong.

I think that system calls are ok in general because they are atomic from the application's point of view (they return EINTR if interrupted). You have to be sure that the libc function is only a system call and not something more though.

PR #164, opened 2016-12-08
Derive `Clone` for `StateMachine` and `OpcodesIter`

fitzgen:
r? @jonhoo


jonhoo:
I assume we don't have @bors running?
In any case, r+, LGTM

PR #165, opened 2016-12-09
Change DebuggingInformationEntry::attr() to preserve the Result

philipc:
Not sure how much of an improvement this is. It's more correct, but harder to use and the `Result` would have been picked up on `EntriesCursor::next` anyway. It matches the documentation now at least.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -792,27 +792,31 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev, 
 
     /// Find the first attribute in this entry which has the given name,
     /// and return it. Returns `Ok(None)` if no attribute is found.
-    pub fn attr(&self, name: constants::DwAt) -> Option<Attribute<'input, Endian>> {
+    pub fn attr(&self, name: constants::DwAt) -> Result<Option<Attribute<'input, Endian>>> {
         let mut attrs = self.attrs();
-        while let Ok(Some(attr)) = attrs.next() {
+        while let Some(attr) = try!(attrs.next()) {
             if attr.name() == name {
-                return Some(attr);
+                return Ok(Some(attr));
             }
         }
-        None
+        Ok(None)
     }
 
     /// Find the first attribute in this entry which has the given name,
     /// and return its raw value. Returns `Ok(None)` if no attribute is found.
-    pub fn attr_value_raw(&self, name: constants::DwAt) -> Option<AttributeValue<'input, Endian>> {
-        self.attr(name).map(|attr| attr.raw_value())
+    pub fn attr_value_raw(&self,
+                          name: constants::DwAt)
+                          -> Result<Option<AttributeValue<'input, Endian>>> {
+        self.attr(name).map(|attr| attr.map(|attr| attr.raw_value()))
     }
 
     /// Find the first attribute in this entry which has the given name,
     /// and return its normalized value.  Returns `Ok(None)` if no
     /// attribute is found.
-    pub fn attr_value(&self, name: constants::DwAt) -> Option<AttributeValue<'input, Endian>> {
-        self.attr(name).map(|attr| attr.value())
+    pub fn attr_value(&self,
+                      name: constants::DwAt)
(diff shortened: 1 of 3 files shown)


jonhoo:
While I agree `Result<Option<T>>` is an annoying type to work with, I think you're right that this is the *right* way of doing things. In fact, I feel like this PR should take things further in that it should also start using `Result` for the return type of the methods that call `attr()` (it currently instead inserts a bunch of `expect` calls), so that the errors can be propagated all the way up to callers. It's never great when the internals of a library can panic.

src/unit.rs
```diff
@@ -792,27 +792,31 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev,
 
     /// Find the first attribute in this entry which has the given name,
     /// and return it. Returns `Ok(None)` if no attribute is found.
-    pub fn attr(&self, name: constants::DwAt) -> Option<Attribute<'input, Endian>> {
+    pub fn attr(&self, name: constants::DwAt) -> Result<Option<Attribute<'input, Endian>>> {
         let mut attrs = self.attrs();
-        while let Ok(Some(attr)) = attrs.next() {
+        while let Some(attr) = try!(attrs.next()) {
```
jonhoo: Personal preference, but I would like to see this written as
```rust
while let Some(attr) = attrs.next()? {
```

src/unit.rs
```diff
@@ -3808,6 +3812,7 @@ mod tests {
         where Endian: Endianity
     {
         let value = entry.attr_value(constants::DW_AT_name)
+            .expect("Should have parsed the name attribute")
```
jonhoo: I'm slightly inclined to use `.ok().expect()` here, given that they are both considered fatal errors (and pretty similar ones too). As mentioned in the PR comment though, I think the right thing to do here is return `Result` to the caller.

src/unit.rs
philipc: Agree, but we haven't used ? in gimli yet, so I want to leave that for another PR.

src/unit.rs
```diff
@@ -792,27 +792,31 @@ impl<'input, 'abbrev, 'unit, Endian> DebuggingInformationEntry<'input, 'abbrev,
...
     }
 
     /// Find the first attribute in this entry which has the given name,
     /// and return its raw value. Returns `Ok(None)` if no attribute is found.
-    pub fn attr_value_raw(&self, name: constants::DwAt) -> Option<AttributeValue<'input, Endian>> {
-        self.attr(name).map(|attr| attr.raw_value())
+    pub fn attr_value_raw(&self,
+                          name: constants::DwAt)
+                          -> Result<Option<AttributeValue<'input, Endian>>> {
+        self.attr(name).map(|attr| attr.map(|attr| attr.raw_value()))
```
philipc: An alternative is `Ok(self.attr(name)?.map(|attr| attr.raw_value())`. Not sure if that is better though.

src/unit.rs
philipc: This is in a unit test. We use `expect` there instead of returning `Result`, and having separate expects lets us know exactly which one failed.

philipc:
> In fact, I feel like this PR should take things further in that it should also start using Result for the return type of the methods that call attr() (it currently instead inserts a bunch of expect calls), so that the errors can be propagated all the way up to callers. It's never great when the internals of a library can panic.

Fully agree that the library should never panic, but I think all of these are in tests or example code. I think we should do better in example code, but that change is too big for this PR. tests though should always panic immediately, that's what they are testing.

src/unit.rs
jonhoo: Fair enough.

src/unit.rs
jonhoo: No, keep it as it is. I was more complaining that there isn't a way to make this pretty in Rust than complaining about your code in particular.

src/unit.rs
jonhoo: Good point, I completely missed the context.

jonhoo:
Agreed, I missed the context for some of the changes. LGTM, r+

philipc:
Thanks for the review!

PR #167, opened 2016-12-13
Add some text to expression evaluator documentation

tromey:


jonhoo:
While we're updating this, I'd like to also see a somewhat longer description for `set_max_iterations`. In particular, it is unclear what exactly "iterations" means. Also, the use of `None` in the description doesn't match the interface (which takes a `u32`, not an `Option<u32>`).

tromey:
Thanks for the suggestions!  I'll fix up this patch tomorrow.

tromey:
For the `None` thing, what is going on is that the API is asymmetric -- you can set the max iterations but you can't clear it.  This capability didn't seem very important so I left it out.

jonhoo:
@tromey I think that's fine (not letting the user clear it), but I think it is then not right for the method documentation to talk about `None`.

jonhoo:
LGTM, r+

PR #168, opened 2016-12-13
Add `LineNumberProgramHeader::raw_program_buf`

fitzgen:
r? @jonhoo

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -2188,7 +2188,8 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
                 // down one level.
                 depth += 1;
 
-                let sibling_ptr = try!(self.current().unwrap().attr_value(constants::DW_AT_sibling));
+                let sibling_ptr =
+                    try!(self.current().unwrap().attr_value(constants::DW_AT_sibling));
                 if let Some(AttributeValue::UnitRef(offset)) = sibling_ptr {
                     if self.unit.is_valid_offset(offset) {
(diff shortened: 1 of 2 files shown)


src/unit.rs
```diff
@@ -2188,7 +2188,8 @@ impl<'input, 'abbrev, 'unit, Endian> EntriesCursor<'input, 'abbrev, 'unit, Endia
                 // down one level.
                 depth += 1;
 
-                let sibling_ptr = try!(self.current().unwrap().attr_value(constants::DW_AT_sibling));
+                let sibling_ptr =
+                    try!(self.current().unwrap().attr_value(constants::DW_AT_sibling));
```
fitzgen: Its in a second commit, which just ran `rustfmt`, so it is separate.

jonhoo:
I don't think 20a441f should be included, but r+ for 637f70b

src/unit.rs
jonhoo: Yeah, but I don't feel like it should be a part of this PR. In the larger context it probably doesn't matter, but this change could probably be pushed separately (maybe even without a PR)

fitzgen:
> I don't think 20a441f should be included, but r+ for 637f70b

I think the best alternative to `rustfmt`ing on an as-needed basis in separate commits would be enforcing `rustfmt`ing of PRs in CI. In order to keep the workflow good, this would also require synchronizing on a particular `rsutfmt` version for all of us (perhaps enforced in the `./format` script). Something we should probably do at some point, if we get enough concurrent contributors, but not something we need at the moment, when @philipc and I last discussed the issue.

PR #169, opened 2016-12-13
Link to the DWARF standard in the README

fitzgen:
Care to rubber stamp this, @philipc ?


jonhoo:
Should this also have a link to DWARFv4 given that that's what we really support atm?

fitzgen:
We should in theory support DWARF 2 through 4 since they are mostly backwards compatible and we attempt to handle the differences between each version. 5 is also mostly backwards compatible as well, we just don't understand the semantics of new attributes and tags.

So, I don't know if it is worth pointing to a specific version when we're aiming for support of all modern versions of DWARF... Do you think it is that big of a deal?

jonhoo:
Eh, no, you're probably right. Never mind. LGTM, r+

fitzgen:
Going to just merge now, since Linux is green and OSX is backlogged, and this is just a README change anyways.

PR #170, opened 2016-12-16
Add Endianity::is_big_endian()

philipc:
Allows users to implement endian-specific functionality without needing
to define further traits.

I needed this to correctly handle `DW_AT_bit_offset` attributes for data member entries.


src/endianity.rs
```diff
@@ -13,6 +13,9 @@ use std::ops::{Deref, Index, Range, RangeFrom, RangeTo};
 /// that implement this trait, it is just used as compile-time phantom data.
 pub trait Endianity
     : byteorder::ByteOrder + Debug + Default + Clone + Copy + PartialEq + Eq {
+
+    /// Return true for big endian byte order.
+    fn is_big_endian() -> bool;
```
fitzgen: For completeness's sake, it might make sense to also have an `is_little_endian`?

philipc: Yep, done.

PR #175, opened 2017-01-16
Add a NamedDwarfSection trait to ease loading of DWARF sections.

khuey:
This makes life easier for library consumers because I can skip writing some of the loading boilerplate now.  See the doc comment on the trait.

### DIFF
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -6,6 +6,7 @@ use parser::{Error, Result, Format};
 use parser::{parse_unsigned_leb, parse_u8};
 use unit::UnitHeader;
 use std::collections::hash_map;
+use Section;
 
 /// An offset into the `.debug_abbrev` section.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -49,6 +50,22 @@ impl<'input, Endian> DebugAbbrev<'input, Endian>
     pub fn abbreviations(&self, debug_abbrev_offset: DebugAbbrevOffset) -> Result<Abbreviations> {
         let input: &[u8] = self.debug_abbrev_section.into();
         Abbreviations::parse(&input[debug_abbrev_offset.0..]).map(|(_, abbrevs)| abbrevs)
+    }
+}
+
+impl<'input, Endian> Section<'input> for DebugAbbrev<'input, Endian>
+    where Endian: Endianity
+{
+    fn section_name() -> &'static str {
+        ".debug_abbrev"
+    }
+}
+
+impl<'input, Endian> From<&'input [u8]> for DebugAbbrev<'input, Endian>
+    where Endian: Endianity
+{
+    fn from(v: &'input [u8]) -> Self {
+        Self::new(v)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -206,3 +206,25 @@ pub use unit::{DebugTypes, DebugTypesOffset, DebugTypeSignature, TypeUnitHeaders
                TypeUnitHeader};
 pub use unit::{EntriesCursor, EntriesTree, EntriesTreeIter, DebuggingInformationEntry};
 pub use unit::{AttrsIter, Attribute, AttributeValue};
+
+/// A convenience trait for loading DWARF sections from object files.  To be
+/// used like:
+///
+/// ```
+/// use gimli::{DebugInfo, LittleEndian, Section};
+///
+/// fn load_section<'a, S, F>(loader: F) -> S
+///   where S: Section<'a>, F: FnOnce(&'static str) -> &'a [u8]
+/// {
+///   let data = loader(S::section_name());
+///   S::from(data)
+/// }
+///
+/// let buf = [0x00, 0x01, 0x02, 0x03];
+///
+/// let debug_info: DebugInfo<LittleEndian> = load_section(|_: &'static str| &buf);
+/// ```
+pub trait Section<'input> : From<&'input [u8]> {
+    /// Returns the ELF section name for this type.
+    fn section_name() -> &'static str;
+}
--- a/src/aranges.rs
+++ b/src/aranges.rs
@@ -6,6 +6,7 @@ use unit::{DebugInfoOffset, parse_debug_info_offset};
 use std::cmp::Ordering;
 use std::marker::PhantomData;
 use std::rc::Rc;
+use Section;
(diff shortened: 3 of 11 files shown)


jonhoo:
This is good idea!

Could we maybe also make the different section types implement `From<&[u8]>`, so that we could write code like

```rust
fn new<'a>(file: &'a object::File) -> DebugInfo<'a> {
    DebugInfo::from(file.get_section(DebugInfo::name()))
}
```

That would also remove the need for the `construct` method on the trait. I'm also hesitant about the trait function name `name()`. Could we make it `section_name()`?

In fact, come to think of it, maybe we should ditch the trait and instead just have an enum:
```rust
enum DebugSection {
  DebugInfo,
  // ...
}

impl DebugSection {
  fn section(&self) -> &'static str {
    match *self {
      DebugSection::DebugInfo => ".debug_info",
      // ...
    }
  }
}
```

That would give us code that reads almost exactly as the example above, but with a much less verbose implementation, and without this extraneous trait lying around?

khuey:
Replacing `construct` with `From<&[u8]>` is righteous.  But the suggested enum wouldn't be connected to the gimli type, afaict, so it doesn't really help.

philipc:
Would it make sense to modify the dwarfdump example to use this? If so, can you do that so that I can see its use in practice?

jonhoo:
@khuey I'm not sure I see what the advantage of tying the section name to the gimli type is?

fitzgen:
It means we know we are getting string contents from debug_str rather than
(incorrectly) from debug_info by construction. With an enumerate, we would
need to check every time and handle errors appropriately.

khuey:
> Would it make sense to modify the dwarfdump example to use this? If so, can you do that so that I can see its use in practice?

Not really, no.  Here's some code from what I'm actually working on.

```
#[derive(Default)]
struct MappedDebugInfoInner<'a> {
    debug_info: Option<gimli::DebugInfo<'a, LittleEndian>>,
}

struct MappedDebugInfo<'a> {
    inner: OwningHandle<Box<Mmap>, Box<OwningHandle<Box<object::File<'a>>, Box<MappedDebugInfoInner<'a>>>>>,
}

impl<'a> MappedDebugInfo<'a> {
    fn new(mmap: Box<Mmap>) -> MappedDebugInfo<'a> {
	MappedDebugInfo {
            inner: OwningHandle::new(mmap, |mmap| {
		let mmap: &memmap::Mmap = unsafe { &*mmap };
		Box::new(OwningHandle::new(Box::new(object::File::parse(unsafe { mmap.as_slice() })), |_| {
                    Box::new(MappedDebugInfoInner::default())
		}))
            }),
	}
    }
}

/// `DwarfEngine`                                                                                                            
pub struct DwarfEngine<'a> {
    cached_maps: HashMap<PathBuf, MappedDebugInfo<'a>>,
}

impl<'a> DwarfEngine<'a> {
    /// Create a new `DwarfEngine`                                                                                           
    pub fn new() -> DwarfEngine<'a> {
	DwarfEngine {
            cached_maps: HashMap::new(),
	}
    }

    fn debug_section_for<D>(&'a mut self,
                            path: PathBuf,
                            offset: FieldOffset<MappedDebugInfoInner<'a>, Option<D>>) -> Result<&'a D>
	where D: gimli::NamedDwarfSection<'a>
    {
     	match self.cached_maps.entry(path) {
            Entry::Occupied(o) => {
		let mapped = o.into_mut();
		if offset.apply(mapped.inner.deref()).is_some() {
                    return Ok(offset.apply(mapped.inner.deref()).as_ref().unwrap());
		}

                mapped.inner.deref_mut().update(|file, debug_sections| {
                    let file: &object::File = unsafe { &*file };
                    let slot = offset.apply_mut(debug_sections);
                    *slot = file.get_section(D::section_name()).map(|r| D::from(r));
		});
		// XXX finish me
            },
            _ => panic!(),
        }
    }

    pub fn debug_info_for(&'a mut self, path: PathBuf) -> Result<&'a gimli::DebugInfo<'a, LittleEndian>> {
        self.debug_section_for(path, offset_of!(MappedDebugInfoInner<'a> => debug_info))
    }

    pub fn debug_blah_for(&'a mut self, path: PathBuf) -> Result<&'a gimli::DebugBlah<'a, LittleEndian>> {
        self.debug_section_for(path, offset_of(!MappedDebugInfoInner<'a> => debug_blah))
    }

```

khuey:
> @khuey I'm not sure I see what the advantage of tying the section name to the gimli type is?

It allows us to deduce everything needed from the type of the variable we want to fill.  See the code sample above.

src/lib.rs
```diff
@@ -206,3 +206,15 @@ pub use unit::{DebugTypes, DebugTypesOffset, DebugTypeSignature, TypeUnitHeaders
...
 pub use unit::{EntriesCursor, EntriesTree, EntriesTreeIter, DebuggingInformationEntry};
 pub use unit::{AttrsIter, Attribute, AttributeValue};
+
+/// A convenience trait for loading DWARF sections from object files.  To be
+/// used like:
+///
+/// fn load_stuff<'a, S: NamedDwarfSection<'a>>() -> S {
+///   let data = thing_that_loads_object_sections<'a>::get(S::name());
+///   S::from(data)
+/// }
```
philipc: This needs quotes around it, and once you do that it'll try testing it, so that will require more fixes.

src/lib.rs
```diff
@@ -206,3 +206,15 @@ pub use unit::{DebugTypes, DebugTypesOffset, DebugTypeSignature, TypeUnitHeaders
...
+pub trait NamedDwarfSection<'input> : From<&'input [u8]> {
```
philipc: Does it matter if trait names are generic? If not, I prefer simply `Section` since I think all gimli sections will have names and be for DWARF.

src/abbrev.rs
```diff
@@ -6,6 +6,7 @@ use parser::{Error, Result, Format};
 use parser::{parse_unsigned_leb, parse_u8};
 use unit::UnitHeader;
 use std::collections::hash_map;
+use super::NamedDwarfSection;
```
philipc: This could be just `use NamedDwarfSection;` unless there is some advantage to using a relative path here.

philipc:
LGTM, thanks!

khuey:
ty!

PR #179, opened 2017-01-26
Add API for finding sequences within a line number program and resuming at sequence boundaries

### LINKED ISSUE #178: Make it possible to resume a line number program from a sequence boundary
In order to make it possible to address things like gimli-rs/addr2line#29 while maintaining reasonable performance I'd like to be able to resume the line number program from a sequence boundary.

The correct state of the `LineNumberProgramHeader` and an offset into the line number program are all that are required to restart a line number program from a sequence boundary.  And almost all of the `LineNumberProgramHeader` state is immutable, except for the table of file names.  To solve that I propose to share a table of file names between `LineNumberProgramHeader`s that are cloned from one another.  That gives us an API that looks like (in pseudo-Rust)

```
pub struct LineNumberProgramHeader {
  // existing immutable bits
  file_names: Arc<Mutex<Vec<FileEntry>>>
  valid_files: usize;
}

pub struct PausedLineNumberProgram {
  valid_files: usize,
  rest: // whatever type this is,
}

impl StateMachine {
  fn resume(header: LineNumberProgramHeader, rest: Whatever) -> Self {
    // make default row
    let opcodes = OpcodesIter {
      input: rest,
      endian: PhantomData,
    };
    StateMachine {
      header: header,
      row: row,
      opcodes: opcodes,
    }
  }

  pub fn save(&self) -> PausedLineNumberProgram {
    assert!(end_sequence())
    PausedLineNumberProgram {
      valid_files: self.valid_files,
      rest: rest_of_the_program(),
    }
  }
}

impl LineNumberProgramHeader {
  fn define_file(filename) {
    // Add filename if file_names.len() == valid_files, valid_files++
    // else ignore (debug_assert! matching, perhaps)
  }

  fn get_file(index) {
    // Bound returned filenames from file_names by valid_files
  }

  pub fn resume(&self, lnp: &PausedLineNumberProgram) -> StateMachine {
    let mut header = self.clone();
    header.valid_files = lnp.valid_files;
    StateMachine::resume(header, lnp.rest)
  }
}
```

If you don't like foisting the `Arc<Mutex<>>` onto everyone (and I suspect you won't) we can introduce a trait or something to let the user determine whether or not they need thread safety.

Thoughts?

CC @fitzgen @philipc @jonhoo

khuey:
Fixes #178.

Something like this perhaps?  I haven't actually tried modifying `addr2line` to use it yet.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -4,7 +4,6 @@ use parser;
 use std::ffi;
 use std::fmt;
 use std::marker::PhantomData;
-use std::ops::Deref;
 use Section;
 
 /// An offset into the `.debug_line` section.
@@ -41,7 +40,7 @@ impl<'input, Endian> DebugLine<'input, Endian>
         DebugLine { debug_line_section: EndianBuf(debug_line_section, PhantomData) }
     }
 
-    /// Parse the line number program header at the given `offset` in the
+    /// Parse the line number program whose header is at the given `offset` in the
     /// `.debug_line` section.
     ///
     /// The `address_size` must match the compilation unit that the lines apply to.
@@ -50,7 +49,7 @@ impl<'input, Endian> DebugLine<'input, Endian>
     /// compilation unit.
     ///
     /// ```rust,no_run
-    /// use gimli::{DebugLine, DebugLineOffset, LineNumberProgramHeader, LittleEndian};
+    /// use gimli::{DebugLine, DebugLineOffset, IncompleteLineNumberProgram, LittleEndian};
     ///
     /// # let buf = [];
     /// # let read_debug_line_section_somehow = || &buf;
@@ -62,22 +61,23 @@ impl<'input, Endian> DebugLine<'input, Endian>
     /// let offset = DebugLineOffset(0);
     /// let address_size = 8;
     ///
-    /// let header = debug_line.header(offset, address_size, None, None)
+    /// let program = debug_line.program(offset, address_size, None, None)
     ///     .expect("should have found a header at that offset, and parsed it OK");
     /// ```
-    pub fn header(&self,
-                  offset: DebugLineOffset,
-                  address_size: u8,
-                  comp_dir: Option<&'input ffi::CStr>,
-                  comp_name: Option<&'input ffi::CStr>)
-                  -> parser::Result<LineNumberProgramHeader<'input, Endian>> {
+    pub fn program(&self,
+                   offset: DebugLineOffset,
+                   address_size: u8,
+                   comp_dir: Option<&'input ffi::CStr>,
+                   comp_name: Option<&'input ffi::CStr>)
+                   -> parser::Result<IncompleteLineNumberProgram<'input, Endian>> {
         if self.debug_line_section.len() < offset.0 {
             return Err(parser::Error::UnexpectedEof);
         }
         let input = self.debug_line_section.range_from(offset.0..);
         let (_, header) =
             try!(LineNumberProgramHeader::parse(input, address_size, comp_dir, comp_name));
-        Ok(header)
+        let program = IncompleteLineNumberProgram { header: header };
+        Ok(program)
     }
 }
 
@@ -97,23 +97,37 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
     }
 }
 
-#[derive(Clone, Debug)]
-enum LineNumberProgramHeaderHolder<'input, Endian>
+/// A `LineNumberProgram` provides access to a `LineNumberProgramHeader` and
+/// a way to add files to the files table if necessary. Gimli consumers should
+/// never need to use or see this trait.
+pub trait LineNumberProgram<'input, Endian>
     where Endian: Endianity
 {
-    Complete(LineNumberProgramHeader<'input, Endian>),
-    Incomplete(LineNumberProgramHeader<'input, Endian>),
+    /// Get a reference to the held `LineNumberProgramHeader`.
+    fn header<'a>(&'a self) -> &'a LineNumberProgramHeader<'input, Endian>;
+    /// Add a file to the file table if necessary.
+    fn add_file(&mut self, file: FileEntry<'input>);
 }
 
-impl<'input, Endian> Deref for LineNumberProgramHeaderHolder<'input, Endian>
+impl<'input, Endian> LineNumberProgram<'input, Endian> for IncompleteLineNumberProgram<'input, Endian>
     where Endian: Endianity
 {
-    type Target = LineNumberProgramHeader<'input, Endian>;
-    fn deref(&self) -> &Self::Target {
-        match self {
-            &LineNumberProgramHeaderHolder::Complete(ref lnp) => lnp,
-            &LineNumberProgramHeaderHolder::Incomplete(ref lnp) => lnp,
-        }
+    fn header<'a>(&'a self) -> &'a LineNumberProgramHeader<'input, Endian> {
+        &self.header
+    }
+    fn add_file(&mut self, file: FileEntry<'input>) {
+        self.header.file_names.push(file);
+    }
+}
+
+impl<'program, 'input, Endian> LineNumberProgram<'input, Endian> for &'program CompleteLineNumberProgram<'input, Endian>
+    where Endian: Endianity
+{
+    fn header<'a>(&'a self) -> &'a LineNumberProgramHeader<'input, Endian> {
+        &self.header
+    }
+    fn add_file(&mut self, _: FileEntry<'input>) {
+// Nop. Our file table is already complete.
     }
 }
 
@@ -123,27 +137,52 @@ impl<'input, Endian> Deref for LineNumberProgramHeaderHolder<'input, Endian>
 /// "The hypothetical machine used by a consumer of the line number information
 /// to expand the byte-coded instruction stream into a matrix of line number
 /// information." -- Section 6.2.1
-#[derive(Clone, Debug)]
-pub struct StateMachine<'input, Endian>
-    where Endian: Endianity
+#[derive(Debug, Clone)]
+pub struct StateMachine<'input, Program, Endian>
+    where Program: LineNumberProgram<'input, Endian>,
+          Endian: Endianity
 {
-    header: LineNumberProgramHeaderHolder<'input, Endian>,
+    program: Program,
     row: LineNumberRow,
     opcodes: OpcodesIter<'input, Endian>,
 }
 
-impl<'input, Endian> StateMachine<'input, Endian>
-    where Endian: Endianity
+
+type OneShotStateMachine<'input, Endian> = StateMachine<'input,
+                                                        IncompleteLineNumberProgram<'input,
+                                                                                          Endian>,
+                                                        Endian>;
+
+type ResumedStateMachine<'program, 'input, Endian> =
+    StateMachine<'input, &'program CompleteLineNumberProgram<'input, Endian>, Endian>;
+
+impl<'input, Program, Endian> StateMachine<'input, Program, Endian>
+    where Program: LineNumberProgram<'input, Endian>,
+          Endian: Endianity
 {
-    fn new(header: LineNumberProgramHeaderHolder<'input, Endian>) -> Self {
+    fn new(program: IncompleteLineNumberProgram<'input, Endian>)
+           -> OneShotStateMachine<'input, Endian> {
         let mut row = LineNumberRow::default();
-        row.registers.reset(header.default_is_stmt());
+        row.registers.reset(program.header().default_is_stmt());
         let opcodes = OpcodesIter {
-            input: header.program_buf.0,
+            input: program.header().program_buf.0,
             endian: PhantomData,
         };
         StateMachine {
-            header: header,
+            program: program,
+            row: row,
+            opcodes: opcodes,
+        }
+    }
+
+    fn resume<'program>(program: &'program CompleteLineNumberProgram<'input, Endian>,
+                        sequence: &LineNumberSequence<'input, Endian>)
+                        -> ResumedStateMachine<'program, 'input, Endian> {
+        let mut row = LineNumberRow::default();
+        row.registers.reset(program.header().default_is_stmt());
+        let opcodes = sequence.opcodes.clone();
+        StateMachine {
+            program: program,
             row: row,
             opcodes: opcodes,
         }
@@ -165,9 +204,9 @@ impl<'input, Endian> StateMachine<'input, Endian>
 
     /// Step 2 of section 6.2.5.1
     fn apply_operation_advance(&mut self, operation_advance: u64) {
-        let minimum_instruction_length = self.header.minimum_instruction_length as u64;
+        let minimum_instruction_length = self.header().minimum_instruction_length as u64;
         let maximum_operations_per_instruction =
-            self.header.maximum_operations_per_instruction as u64;
+            self.header().maximum_operations_per_instruction as u64;
 
         let op_index_with_advance = self.row.registers.op_index + operation_advance;
 
@@ -178,7 +217,7 @@ impl<'input, Endian> StateMachine<'input, Endian>
     }
 
     fn adjust_opcode(&self, opcode: u8) -> u8 {
-        opcode - self.header.opcode_base
+        opcode - self.header().opcode_base
     }
 
     /// Section 6.2.5.1
@@ -187,14 +226,14 @@ impl<'input, Endian> StateMachine<'input, Endian>
 
         // Step 1
 
-        let line_base = self.header.line_base as i64;
-        let line_range = self.header.line_range;
+        let line_base = self.header().line_base as i64;
+        let line_range = self.header().line_range;
         let line_increment = line_base + (adjusted_opcode % line_range) as i64;
         self.apply_line_advance(line_increment);
 
         // Step 2
 
-        let operation_advance = adjusted_opcode / self.header.line_range;
+        let operation_advance = adjusted_opcode / self.header().line_range;
         self.apply_operation_advance(operation_advance as u64);
     }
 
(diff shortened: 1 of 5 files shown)


src/line.rs
```diff
@@ -260,7 +294,18 @@ impl<'input, Endian> StateMachine<'input, Endian>
...
+                match self.header {
+                    LineNumberProgramHeaderHolder::Incomplete(ref mut header) => {
+                        header.file_names.push(entry);
+                    },
+                    _ => {},
+                    /*
+                    LineNumberProgramHeaderHolder::Complete(_) => {
+                        // All the filenames are already in the header, we don't
+                        // need to do anything (except have this branch to get
+                        // rustc not to warn about an unused 'Complete').
```
fitzgen: I like the commented out version because

* If we add more variants (however unlikely) then the compiler will yell at us to match them
* The comment is nice

src/line.rs
```diff
@@ -1062,6 +1123,72 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
...
+    /// }
+    ///
+    /// let header = get_line_number_program_header();
+    /// let (complete_header, sequences) = header.sequences().unwrap();
+    /// println!("There are {} sequences in this line number program", sequences.len());
+    /// # }
+    /// ```
+    pub fn sequences(self)
+                     -> parser::Result<(LineNumberProgramHeader<'input, Endian>,
+                                        Vec<LineNumberSequence<'input, Endian>>)> {
```
fitzgen: Do we want to let the caller control allocation (maybe they know when they can reuse a vec?) by accepting an `&mut Vec<LineNumberSequence<'input, Endian>>` parameter, asserting it is initially empty, and then filling it?

I'm not actually 100% sure if this would actually work or not due to the `'input` lifetime.

src/unit.rs
```diff
@@ -881,7 +881,7 @@ pub enum AttributeValue<'input, Endian>
     /// different compilation unit from the current one.
     DebugInfoRef(DebugInfoOffset),
 
-    /// An offset into the `.debug_lines` section.
+    /// An offset into the `.debug_line` section.
```
fitzgen: Thanks!

src/line.rs
philipc: Can we add something like a `next_sequence` method to `StateMachine` instead, so that the caller completely controls allocation.

tests/parse_self.rs
```diff
@@ -90,6 +90,18 @@ fn test_parse_self_debug_line() {
...
                 .expect("Should parse and execute all rows in the line number program") {
             }
+
+            let header = debug_line.header(offset, unit.address_size(), comp_dir, comp_name)
+                .expect("should parse line number program header");
+            let (header, sequences) = header.sequences()
+                .expect("should parse and execute the entire line number program");
+            assert!(sequences.len() > 0); // Should be at least one sequence.
+            for sequence in sequences {
+                let mut resumed = header.clone().resume_from(&sequence);
```
philipc: It would be nice to avoid needing to clone the header to iterate the rows in a sequence. So the caller would need to somehow obtain a `StateMachine` that already has a `Complete` header, and resume on that instead.

src/line.rs
```diff
@@ -1062,6 +1123,72 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
...
+                None => break,
+            };
+
+            // We just finished a sequence.
+            sequences.push(LineNumberSequence {
+                // In theory one could have multiple DW_LNE_end_sequence opcodes
+                // in a row.
+                start: sequence_start_addr.unwrap_or(0),
+                end: sequence_end_addr,
+                rest: state_machine.opcodes.clone(),
```
philipc: This is the opcodes from after the sequence? Also, it would be nice to limit it to only the opcodes for the sequence (so reduce the size of the input buffer to cover only the sequence).

tests/parse_self.rs
khuey: Er, yes, good catch.

src/line.rs
khuey: Er, yes, good catch.

src/line.rs
khuey: The comment about rustc warning is actually wrong.  But sure, I can be explicit here.

tests/parse_self.rs
philipc: Another option would be to change the `Complete` variant to only use a reference.

src/line.rs
khuey: This is the only comment I haven't addressed yet.  Having the caller iterate through the sequences runs the risk of producing `LineNumberSequence`s that can't actually be used yet if the `StateMachine` has not run to completion.

An outparam `Vec` doesn't have that problem, but we'd have to be careful to clear the contents of the `Vec` if we return early due to a parser error condition.

khuey:
Take a look at this one.  I'm pretty happy with it, except that `StateMachine::resume` doesn't validate that the iterator has been run to completion.  We could return an error there (a `parser:Error` seems a little goofy because its an API usage mistake and not a problem with the DWARF though).  There's nothing that stops one from using a `LineNumberSequence` with the wrong `StateMachine` entirely though which is why I'm not too worried about it.

src/line.rs
```diff
@@ -658,6 +693,26 @@ pub struct OpcodesIter<'input, Endian>
 impl<'input, Endian> OpcodesIter<'input, Endian>
     where Endian: Endianity
 {
+    fn remove_trailing(&self, other: &OpcodesIter<'input, Endian>) -> OpcodesIter<'input, Endian> {
+        println!("Us {:?}, {} them {:?}, {}, end {:?}",
+                 self.input.as_ptr(),
+                 self.input.len(),
+                 other.input.as_ptr(),
+                 other.input.len(),
+                 unsafe { self.input.as_ptr().offset(self.input.len() as isize)});
```
khuey: Indeed, thanks.

khuey:
I think I might want to switch the state machine to using references to the header though.  As is you have to clone it to use it in a multithreaded environment, which means cloning the header.  So don't merge this yet while I sleep on it :)

khuey:
Ok, I ended up changing the `LineNumberProgramHeaderHolder` into a trait that is implemented by `IncompleteLineNumberProgramHeaderHolder` and `CompleteLineNumberProgramHeaderHolder`.  `StateMachine` largely became `StateMachineInner`, which is parameterized over a `LineNumberProgramHeaderHolder`, and a new `StateMachine` and `ResumableStateMachine` provide the public facing APIs (the original and the resumable, respectively).

This allows me to make `ResumableStateMachine` implement `Clone` (because the underlying reference is immutable) while before we couldn't really do that.  Now I can do addr2line lookups on multiple threads simultaneously from the sequence data because the sequence array is immutable and I can make a cheap clone of the `ResumableStateMachine`.

philipc:
I like the concept of a trait for `add_file`. We may want to take it further and add a `file` method to that trait too, although the current implementations don't need it. I think the trait and its implementations need a better names though.

I'm not sure about making `StateMachineInner` private. It's good to avoid exposing implementation details, but it's also good to keep the implementation simple. `StateMachine` and `ResumableStateMachine` could be type aliases instead of new-types. That would cut down on a bit of boilerplate. Open to opinions on that.

Changing `StateMachine` to use a mutable reference to the header is a step backwards; it makes #40 a problem again. I can't see a reason to need this change. If we really want this to be a reference then I prefer to make it a borrow, and store the mutable list of file names in `IncompleteLineNumberProgramHeaderHolder` instead.

src/line.rs
```diff
@@ -96,32 +97,102 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
     }
 }
 
+/// Shutup
+pub trait LineNumberProgramHeaderHolder<'input, Endian> : Deref<Target = LineNumberProgramHeader<'input, Endian>>
```
fitzgen: Its generally frowned upon to `Deref` but *also* add methods. Generally a thing should be a magic pointer and only deref, or if it must also have its own methods, then it should use `As[Mut]Ref` instead of `Deref`. The reasoning is that code becomes harder to read with the "fake inheritence" that `Deref` provides, and method resolution is obscured as well.

So I would prefer:

```rust
pub trait LineNumberProgramHeaderHolder<'input, Endian> : AsRef<LineNumberProgramHeader<'input, Endian>>
```

src/line.rs
```diff
@@ -96,32 +97,102 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
...
+    fn deref(&self) -> &Self::Target {
+        self.header
+    }
+}
+
+impl<'header, 'input, Endian> LineNumberProgramHeaderHolder<'input, Endian> for CompleteLineNumberProgramHeaderHolder<'header, 'input, Endian>
+    where Endian: Endianity
+{
+    fn add_file(&mut self, _: FileEntry<'input>) {
+        // Nop
```
fitzgen: Perhaps with debug assertions here?

src/line.rs
```diff
@@ -96,32 +97,102 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
...
 ///
 /// "The hypothetical machine used by a consumer of the line number information
 /// to expand the byte-coded instruction stream into a matrix of line number
 /// information." -- Section 6.2.1
-#[derive(Clone, Debug)]
-pub struct StateMachine<'input, Endian>
-    where Endian: Endianity
+#[derive(Debug)]
+struct StateMachineInner<'input, Holder, Endian>
+    where Holder: LineNumberProgramHeaderHolder<'input, Endian>, Endian: Endianity
```
fitzgen: I suspect `rustfmt` will put `Endian: Endianity` on a new line.

src/line.rs
```diff
@@ -297,43 +358,43 @@ impl<'input, Endian> StateMachine<'input, Endian>
...
+            } else {
+                // Previous opcode was one of:
+                // - Special - specified in Section 6.2.5.1, steps 4-7
+                // - Copy - specified in Section 6.2.5.2
+                // The reset behaviour is the same in both cases.
+                self.row.registers.discriminator = 0;
+                self.row.registers.basic_block = false;
+                self.row.registers.prologue_end = false;
+                self.row.registers.epilogue_begin = false;
+            }
```
fitzgen: Did `rustfmt` add this indentation??

src/line.rs
```diff
@@ -347,13 +408,117 @@ impl<'input, Endian> StateMachine<'input, Endian>
...
+    }
+
+    /// Parse and execute opcodes until we reach a row matching `addr`, the end of the program,
+    /// or an error.
+    pub fn run_to_address
+        (&mut self,
+         addr: &u64)
+         -> parser::Result<Option<(&LineNumberProgramHeader<'input, Endian>, &LineNumberRow)>> {
+        self.inner.run_to_address(addr)
+    }
```
fitzgen: Since both `StateMachine` and `ResumableStateMachine` have {header, next_row, run_to_address} methods, it could be nice to define them in a trait and consolidate the documentation in one place.

fitzgen:
> I'm not sure about making StateMachineInner private. It's good to avoid exposing implementation details, but it's also good to keep the implementation simple. StateMachine and ResumableStateMachine could be type aliases instead of new-types. That would cut down on a bit of boilerplate. Open to opinions on that.

Either way works for me.

> Changing StateMachine to use a mutable reference to the header is a step backwards; it makes #40 a problem again. I can't see a reason to need this change. If we really want this to be a reference then I prefer to make it a borrow, and store the mutable list of file names in IncompleteLineNumberProgramHeaderHolder instead.

+1

khuey:
Using an owned header in one variant and immutable references in the other isn't feasible without using `Rc`/`Arc`.  If we go back to an owned header then `sequences` has to take `self` instead of `&mut self`, and once we build the `ResumableStateMachine` we can't hand back out the completed header.

philipc:
Can you change `sequences` to take `self` and return `CompleteLineNumberProgramHeaderHolder`, and add a method to `CompleteLineNumberProgramHeaderHolder` to return a `ResumableStateMachine`?

khuey:
Yeah, that's a good idea.  I did essentially that, with some of the types renamed.  Lmk what you think of this version.

src/line.rs
```diff
@@ -96,33 +97,113 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
     }
 }
 
+/// A `LineNumberProgramHolder` provides access to a `LineNumberProgramHeader` and
+/// a way to add files to the files table if necessary. Gimli consumers should
+/// never need to use or see this trait.
+pub trait LineNumberProgramHolder<'input, Endian>
```
philipc: Rename this trait to `LineNumberProgram`.

src/line.rs
```diff
@@ -96,33 +97,113 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
...
+    fn header<'a>(&'a self) -> &'a LineNumberProgramHeader<'input, Endian>;
+    /// Add a file to the file table if necessary.
+    fn add_file(&mut self, file: FileEntry<'input>);
+}
+
+/// `IncompleteLineNumberProgramHolder` holds a `LineNumberProgramHeader` that needs
+/// files defined by the line number program header to be added to it. Gimli consumers
+/// should never need to use or see this trait.
+#[derive(Debug)]
+pub struct IncompleteLineNumberProgramHolder<'input, Endian>
```
philipc: Delete this struct, it adds no benefit. Replace its usage with `IncompleteLineNumberProgram`.

src/line.rs
```diff
@@ -96,33 +97,113 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
...
+    fn add_file(&mut self, file: FileEntry<'input>) {
+        self.program.header.file_names.push(file);
+    }
+}
+
+/// `CompleteLineNumberProgramHolder` holds a `LineNumberProgramHeader` that does not need
+/// files defined by the line number program header to be added to it. Gimli consumers
+/// should never need to use or see this trait.
+#[derive(Debug, Clone)]
+pub struct CompleteLineNumberProgramHolder<'program, 'input, Endian>
```
philipc: Delete this struct, it adds no benefit. Replace its usage with `&'program CompleteLineNumberProgram`.

src/line.rs
```diff
@@ -96,33 +97,113 @@ impl<'input, Endian> From<&'input [u8]> for DebugLine<'input, Endian>
...
 /// instructions to source locations.
 ///
 /// "The hypothetical machine used by a consumer of the line number information
 /// to expand the byte-coded instruction stream into a matrix of line number
 /// information." -- Section 6.2.1
-#[derive(Clone, Debug)]
-pub struct StateMachine<'input, Endian>
-    where Endian: Endianity
+#[derive(Debug, Clone)]
+pub struct StateMachine<'input, Holder, Endian>
```
philipc: Rename `Holder` to `Program`.

src/line.rs
```diff
@@ -1183,6 +1288,136 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
...
+{
+    /// Retrieve the `LineNumberProgramHeader` for this program.
+    pub fn header(&self) -> &LineNumberProgramHeader<'input, Endian> {
+        &self.header
+    }
+
+    /// Construct a new `StateMachine` for executing line programs and
+    /// generating the line information matrix.
+    pub fn rows(self) -> OneShotStateMachine<'input, Endian> {
+        StateMachine::<IncompleteLineNumberProgramHolder<_>, _>::new(IncompleteLineNumberProgramHolder {
```
philipc: Change to `OneShotStateMachine::new(self)`

src/line.rs
```diff
@@ -1183,6 +1288,136 @@ impl<'input, Endian> LineNumberProgramHeader<'input, Endian>
...
+    ///     let mut sm = program.resume_from(sequence);
+    /// }
+    /// # }
+    /// ```
+    pub fn resume_from<'program>(&'program self,
+                                 sequence: &LineNumberSequence<'input, Endian>)
+                                 -> ResumedStateMachine<'program, 'input, Endian> {
+        StateMachine::<CompleteLineNumberProgramHolder<_>, _>::resume(CompleteLineNumberProgramHolder {
+            program: self,
+        }, sequence)
```
philipc: Change to `ResumedStateMachine::resume(self, sequence)`

khuey:
Review comments largely addressed.  I'm not sure whether the `LineNumberProgram` trait is actually worth it at this point (instead of implementing `Deref` on `LineNumberProgramHolder` we could just implement a `header()` method).

philipc:
Why did you need to add the enum back in? That wasn't what I intended. The changes I requested should have only been a handful of lines changed. eg when you delete `IncompleteLineNumberProgramHolder`, just replace all its uses with `IncompleteLineNumberProgram`, including the impl of the `LineNumberProgram` trait. The `LineNumberProgram` trait should keep its `add_file` method.

khuey:
I need to hold an `IncompleteLineNumberProgram` and a `&'program CompleteLineNumberProgram`.  `StateMachine` needs to know which one to hold.  We could do that with a type on the `LineNumberProgram` trait, but we also need to handle `add_file`.  We can't pass `&mut self` to it in the `CompleteLineNumberProgram` case because we only have an immutable reference.  Having an intermediate type of some sort solves that problem.

philipc:
But the point of the trait was to enable `StateMachine` to not care which it is. See https://github.com/philipc/gimli/commit/cb924fc9a932f95f8889e47f646e9027cdfc6b3a for what I intended.

khuey:
Ah, I didn't realize you meant to implement `LineNumberProgram` on `&'program CompleteLineNumberProgram`.  That's a neat trick.

Changes are folded in, and this should make `benches` build too.

philipc:
@fitzgen Did you want to take another look before merging?

fitzgen:
Nope, I've been following along without commenting (too many cooks in the kitchen), and it looks good to me. Most recent changes A+++

Ship it!

khuey:
Wait for Travis before merging, I don't have nightly rustc installed so I don't know for sure that `benches` builds yet.

khuey:
Well, Travis seems to be having a bad time today.  `benches` did pass though on the PR.

khuey:
Thanks for the reviews!

PR #181, opened 2017-02-05
Try to only run dwarfdump on executables

philipc:


philipc:
The 'cargo clean' is because it was caching lots of `target/debug/gimli-*`, so we weren't running dwarfdump on the latest. Doesn't seem to have affected build time.

Issue #184, opened 2017-03-06
Add a mechanism for asynchronously evaluating a DWARF expression.


khuey:
The gimli embedding provides the inputs needed to evaluate a DWARF expression by implementing the `EvaluationContext` trait.  In the system I'm building, getting register and memory values requires a database lookup (potentially on another machine if we're `evaluate_entry_value()`ing and subprogram entry was far enough away in time).

Spinning up a thread for evaluating the expression and blocking there works, but what I'd really like is an `Evaluation` that takes an `AsyncEvaluationContext` that returns futures and has an `evaluate` that returns a `Future<Item=Vec<Piece>>`.

I started hacking this up but doing it without duplicating much of the evaluation code is difficult and/or impossible :/

fitzgen:
Yeah, I don't know if Rust provides a way to be generic across synchronous and asynchronous evaluation. I haven't tried, but I suspect it would be tricky.

tromey:
Perhaps the base implementation could be async and we could wrap sync evaluation contexts in something that just returns resolved futures.

khuey:
`futures-rs` has built in support for converting `Result`s to immediately resolvable futures.  Doing it this way does mean introducing an unconditional dependency on `futures-rs` though.

tromey:
Another idea would be to keep the evaluator state in a struct, provide a way to evaluate operations until some contextual data is needed, and then have the evaluator return an enum that holds either a final result (a piece vector) or a contextual request.  This is basically the same effect as futures but without the dependency on `futures-rs`; pushing any asynchrony to the user.

PR #185, opened 2017-03-07
Miscellaneous cleanups

khuey:
While messing around in src/op.rs I cleaned up a couple things.

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -219,6 +219,23 @@ pub enum Operation<'input, Endian>
     },
 }
 
+#[derive(Debug)]
+struct OperationEvaluationResult<'input> {
+    terminated: bool,
+    piece_end: bool,
+    current_location: Location<'input>,
+}
+
+impl<'input> Default for OperationEvaluationResult<'input> {
+    fn default() -> OperationEvaluationResult<'input> {
+        OperationEvaluationResult {
+            terminated: false,
+            piece_end: false,
+            current_location: Location::Empty,
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


tromey:
The first patch is totally fine.

For the second patch, I think the rationale behind a mutable context is that we didn't want to assume a readonly context.  Obviously there's no concrete use for this at this time, but my thought was perhaps some implementation would want to be caching results or something along those lines.

khuey:
The methods on `EvaluationContext` don't take mutable self refs though ...

philipc:
I think it's better to change `callbacks` to be a value, not a reference. Then the caller can choose which type of reference they want.

khuey:
fwiw I don't think the Travis failure is me. Looks like a compiler upgrade broke some existing code?

tromey:
> The methods on EvaluationContext don't take mutable self refs though ...

Yeah.  Probably I was confused somehow, it seems likely.  I don't have any objections to this; and since it seems like we're contemplating changing the API anyway...

fitzgen:
Looks like a change in type inference.

```
---- src/cfi.rs - cfi::CfiEntriesIter (line 551) stdout ----
	error[E0282]: type annotations needed
  --> <anon>:25:1
   |
25 | Ok(())
   | ^^ cannot infer type for `E`
```

khuey:
Something like that, perhaps?

khuey:
I'm going to end up rewriting everything in this PR for #184, so let's not bother what this.

PR #186, opened 2017-03-08
Allow for async evaluation of DWARF expressions

### LINKED ISSUE #184: Add a mechanism for asynchronously evaluating a DWARF expression.

khuey:
Fixes #184, as suggested by @tromey.

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -895,7 +895,8 @@ pub enum EvaluationResult<'input, Endian>
 /// which is either `EvaluationResult::Complete` or a value indicating what
 /// data is needed to resume the `Evaluation`.  The consumer is responsible for
 /// producing that data and resuming the computation with the correct method,
-/// as documented for `EvaluationResult`.
+/// as documented for `EvaluationResult`.  Only once an `EvaluationResult::Complete`
+/// is returned can the consumer call `result()`.
 ///
 /// This design allows the consumer of `Evaluation` to decide how and when to
 /// produce the required data and resume the computation.  The `Evaluation` can
@@ -1000,12 +1001,16 @@ impl<'input, Endian> Evaluation<'input, Endian>
     /// stack before evaluation commences.  If no initial value is
     /// set, and the expression uses an opcode requiring the initial
     /// value, then evaluation will fail with an error.
+    ///
+    /// # Panics
+    /// Panics if `set_initial_value()` has already been called, or if
+    /// `evaluate()` has already been called.
     pub fn set_initial_value(&mut self, value: u64) {
         match self.state {
             EvaluationState::Start(None) => {
                 self.state = EvaluationState::Start(Some(value));
             },
-            _ => panic!(),
+            _ => panic!("`Evaluation::set_initial_value` was called twice, or after evaluation began."),
         };
     }
 
--- a/src/op.rs
+++ b/src/op.rs
@@ -7,7 +7,6 @@ use parser::{Error, Format, parse_u8e, parse_i8e, parse_u16, parse_i16, parse_u3
 use endianity::{Endianity, EndianBuf};
 use unit::{UnitOffset, DebugInfoOffset};
 use std::marker::PhantomData;
-use std::fmt;
 
 /// A reference to a DIE, either relative to the current CU or
 /// relative to the section.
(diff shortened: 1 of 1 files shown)


src/op.rs
```diff
@@ -1198,34 +1266,210 @@ impl<'context, 'input, Endian, Context> Evaluation<'context, 'input, Endian, Con
...
-    /// object.
-    pub fn evaluate(&mut self) -> Result<Vec<Piece<'input>>, Context::ContextError>
-        where Endian: Endianity
+    /// Get the evaluation result.  This `Evaluation` must have previously been
+    /// driven to completion by calling the evaluate and resume_with methods
+    /// until `EvaluationResult::Complete` was returned.
+    pub fn result(self) -> Vec<Piece<'input>> {
+        match self.state {
+            EvaluationState::Complete => self.result,
+            _ => panic!(),
```
fitzgen: We should have a diagnostic message for the panic here:

```rust
_ => panic!("Called `Evaluation::result` on an `Evaluation` that has not been completed")
```

src/op.rs
```diff
@@ -1198,34 +1266,210 @@ impl<'context, 'input, Endian, Context> Evaluation<'context, 'input, Endian, Con
...
     }
 
-    /// Evaluate a DWARF expression.  If successful, the result will
-    /// hold a vector of pieces, each describing a part of the final
-    /// object.
-    pub fn evaluate(&mut self) -> Result<Vec<Piece<'input>>, Context::ContextError>
-        where Endian: Endianity
+    /// Get the evaluation result.  This `Evaluation` must have previously been
+    /// driven to completion by calling the evaluate and resume_with methods
+    /// until `EvaluationResult::Complete` was returned.
```
fitzgen: This should have a section of the docs like the standard library does about panics, for example like in here: https://doc.rust-lang.org/nightly/std/string/struct.String.html#method.insert_str

So it would be something like this:

```markdown
# Panics

If this `Evaluation` has not been driven to completion, calling `result` will panic.
```

src/op.rs
```diff
@@ -1198,34 +1266,210 @@ impl<'context, 'input, Endian, Context> Evaluation<'context, 'input, Endian, Con
...
+    /// more information again.
+    pub fn resume_with_memory(&mut self, value: u64) -> Result<EvaluationResult<'input, Endian>, Error>
+        where Endian: Endianity
+    {
+        match self.state {
+            EvaluationState::Error(err) => return Err(err),
+            EvaluationState::Waiting(OperationEvaluationResult::AwaitingMemory { .. }) => {
+                self.push(value);
+            },
+            _ => panic!(),
```
fitzgen: Ditto with regards to a diagnostic message for the panic and a section in the doc comment.

src/op.rs
```diff
@@ -1198,34 +1266,210 @@ impl<'context, 'input, Endian, Context> Evaluation<'context, 'input, Endian, Con
...
+    /// more information again.
+    pub fn resume_with_register(&mut self, register: u64) -> Result<EvaluationResult<'input, Endian>, Error>
+        where Endian: Endianity
+    {
+        match self.state {
+            EvaluationState::Error(err) => return Err(err),
+            EvaluationState::Waiting(OperationEvaluationResult::AwaitingRegister { offset, .. }) => {
+                self.push(register.wrapping_add(offset));
+            },
+            _ => panic!(),
```
fitzgen: Ditto.

src/op.rs
```diff
@@ -1198,34 +1266,210 @@ impl<'context, 'input, Endian, Context> Evaluation<'context, 'input, Endian, Con
...
+    /// or needs more information again.
+    pub fn resume_with_frame_base(&mut self, frame_base: u64) -> Result<EvaluationResult<'input, Endian>, Error>
+        where Endian: Endianity
+    {
+        match self.state {
+            EvaluationState::Error(err) => return Err(err),
+            EvaluationState::Waiting(OperationEvaluationResult::AwaitingFrameBase { offset }) => {
+                self.push(frame_base + offset);
+            },
+            _ => panic!(),
```
fitzgen: You know the drill... Here and all the others below.

fitzgen:
Also, thanks Kyle!

khuey:
Thanks.  The cset I just pushed should address the documentation comments.

My quick skim of the session types paper makes that seem quite involved, so while I agree that it would be nice to statically check that the consumer follows the API contract, I don't think its worth the effort.

fitzgen:
Thanks!

I just submitted #187 to work around the type inference failures in nightly rustc.

fitzgen:
CI looks good aside from the nightly issue fixed in #187 and the fact that it takes 24h for travis to run OSX builds.

PR #187, opened 2017-03-08
Work around nightly rustc breaking type inference

fitzgen:
It seems there was a type inference regression in Nightly rustc, where returning
an Ok literal at the end of a function whose type signature returns a Result is
failing. I also hit this on some other projects.

r? @tromey (or anyone else)


khuey:
lgtm

fitzgen:
Thanks for the review!

PR #188, opened 2017-03-09
Use EndianBuf throughout Evaluation.

khuey:
Not really sure why this isn't this way to begin with.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -495,7 +495,7 @@ fn dump_exprloc<Endian>(data: gimli::EndianBuf<Endian>, unit: &Unit<Endian>)
     let mut space = false;
     while pc.len() != 0 {
         let dwop = gimli::DwOp(pc[0]);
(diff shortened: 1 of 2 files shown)


tromey:
> Not really sure why this isn't this way to begin with.

I thought the whole gimli API just asked users to pass in an `&[u8]` and that `EndianBuf` was generally only used internally.  E.g., `DebugInfo::new` does this.

I don't mind this change but I think it shouldn't be made to just one part of the API.

khuey:
`DebugInfo::new` input comes from a raw ELF section.  In practice, `Evaluation::new` input comes from an `AttributeValue::ExprLoc` or a `LocationListEntry::data` both of which are `EndianBuf`s.

tromey:
I'd rather @fitzgen decide.

fitzgen:
The elf sections take raw input, yes, because that's how you enter our API. Everything after that should be tracking its endianity in the types.

PR #189, opened 2017-03-15
Specify lifetimes for DebugLoc iterators because lifetime elision is too conservative

khuey:
Lifetime elision binds the iterator to the lifetime of `self` rather than to `'input`, which is too limiting.


philipc:
Thanks!

PR #191, opened 2017-03-20
Bump to version 0.13.0

fitzgen:
r? @philipc


fitzgen:
I'll publish on crates.io and make a tag once this is rubber stamped and merged, assuming there aren't reasons we shouldn't make a release right now that I'm not thinking of.

fitzgen:
New version published on crates.io! Thanks @philipc :)

PR #197, opened 2017-05-09
Get `cargo clippy` clean

fitzgen:
Also `cargo fmt` with the latest `rustfmt`.

r? @philipc or @tromey

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -11,7 +11,7 @@ use std::io::Read;
 use std::path::PathBuf;
 
 pub fn read_section(section: &str) -> Vec<u8> {
-    let mut path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap_or(".".into()));
+    let mut path = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap_or_else(|_| ".".into()));
     path.push("./fixtures/self/");
     path.push(section);
 
@@ -157,7 +157,7 @@ fn bench_parsing_line_number_program_opcodes(b: &mut test::Bencher) {
         let header = program.header();
 
         let mut opcodes = header.opcodes();
-        while let Some(opcode) = opcodes.next_opcode(&header).expect("Should parse opcode") {
+        while let Some(opcode) = opcodes.next_opcode(header).expect("Should parse opcode") {
             test::black_box(opcode);
         }
     });
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -143,7 +143,7 @@ impl Abbreviations {
             match abbrev {
                 None => break,
                 Some(abbrev) => {
-                    if let Err(_) = abbrevs.insert(abbrev) {
+                    if abbrevs.insert(abbrev).is_err() {
                         return Err(Error::DuplicateAbbreviationCode);
                     }
                 }
@@ -175,7 +175,7 @@ impl Abbreviation {
                has_children: constants::DwChildren,
                attributes: Vec<AttributeSpecification>)
                -> Abbreviation {
-        assert!(code != 0);
+        assert_ne!(code, 0);
         Abbreviation {
             code: code,
             tag: tag,
@@ -326,6 +326,7 @@ impl AttributeSpecification {
                 }
             }
 
+            // Variably sized forms.
             constants::DW_FORM_block |
             constants::DW_FORM_block1 |
             constants::DW_FORM_block2 |
@@ -335,7 +336,7 @@ impl AttributeSpecification {
             constants::DW_FORM_string |
             constants::DW_FORM_sdata |
             constants::DW_FORM_udata |
-            constants::DW_FORM_indirect => None,
+            constants::DW_FORM_indirect |
 
             // We don't know the size of unknown forms.
             _ => None,
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -599,7 +599,7 @@ impl<'bases, 'input, Endian, Section> CfiEntriesIter<'bases, 'input, Endian, Sec
         // the last entry.
         self.bases.func.borrow_mut().take();
 
-        match parse_cfi_entry(&self.bases, self.section, self.input) {
+        match parse_cfi_entry(self.bases, self.section, self.input) {
             Err(e) => {
                 self.input = EndianBuf::new(&[]);
                 Err(e)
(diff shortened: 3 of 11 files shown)


benches/bench.rs
```diff
@@ -254,17 +260,18 @@ mod cfi {
                 match entry {
                     CieOrFde::Cie(cie) => {
                         let mut instrs = cie.instructions();
-                        while let Some(i) = instrs.next()
-                            .expect("Can parse next CFI instruction OK") {
+                        while let Some(i) =
+                            instrs.next().expect("Can parse next CFI instruction OK") {
```
jonhoo: Ugh... But if rustfmt wants it...

benches/bench.rs
```diff
@@ -318,10 +327,13 @@ mod cfi {
...
                           -> usize {
-        fde.instructions().fold(0, |count, _| count + 1).expect("fold over instructions OK")
+        fde.instructions()
+            .fold(0, |count, _| count + 1)
+            .expect("fold over instructions OK")
     }
 
-    fn get_fde_with_longest_cfi_instructions(eh_frame: EhFrame<LittleEndian>)
+    fn get_fde_with_longest_cfi_instructions
+        (eh_frame: EhFrame<LittleEndian>)
```
jonhoo: Really? `rustfmt` wants that change?

fitzgen: ¯\_(ツ)_/¯

rustfmt.toml
```diff
@@ -1,2 +1,2 @@
 write_mode = "Overwrite"
-
+closure_block_indent_threshold = 1
```
jonhoo: Ahh, this is why some of the changes are looking so odd to me. Why do you want this here?

src/abbrev.rs
```diff
@@ -650,10 +652,8 @@ pub mod tests {
         let expect = Some(Abbreviation::new(1,
                                             constants::DW_TAG_subprogram,
                                             constants::DW_CHILDREN_no,
-                                            vec![
-                    AttributeSpecification::new(constants::DW_AT_name,
-                                                constants::DW_FORM_string),
-                ]));
+                                            vec![AttributeSpecification::new(constants::DW_AT_name,
+                                                                    constants::DW_FORM_string)]));
```
jonhoo: That's not great, but then again, rustfmt still doesn't deal with macros very well..

src/lib.rs
```diff
@@ -148,6 +148,15 @@
...
 #![deny(missing_debug_implementations)]
 
+// Allow clippy warnings when we aren't building with clippy.
+#![allow(unknown_lints)]
+// False positives with `fallible_iterator`.
+#![allow(should_implement_trait)]
+// Many false positives involving `continue`.
+#![allow(never_loop)]
+// False positives when block expressions are used inside an assertion.
+#![allow(panic_params)]
```
jonhoo: These are a bit unfortunate. Are there upstream bugs filed against clippy for the worst offenders here that we can link to?

src/lookup.rs
```diff
@@ -26,6 +26,7 @@ pub trait LookupParser<'input, Endian>
     /// Parse a header from `input`. Returns a tuple of `input` sliced beyond this header and
     /// all of its entries, `input` sliced to contain just the entries corresponding to this
     /// header (without the header itself), and the parsed representation of the header itself.
+    #[allow(type_complexity)]
```
jonhoo: Clippy is right that a number of these are kind of unfortunate, but probably worth dealing with separately.

src/unit.rs
```diff
@@ -975,6 +976,8 @@ impl<'input, Endian> Attribute<'input, Endian>
     /// name.
     ///
     /// See "Figure 20. Attribute encodings" and "Figure 21. Attribute form encodings".
+    #[allow(cyclomatic_complexity)]
+    #[allow(match_same_arms)]
```
jonhoo: Why is this necessary?

src/unit.rs
```diff
@@ -4146,10 +4172,22 @@ mod tests {
...
+                    .expect("Should parse next sibling")
+                    .is_none());
+        assert!(cursor
+                    .next_sibling()
+                    .expect("Should parse next sibling")
+                    .is_none());
+        assert!(cursor
+                    .next_sibling()
+                    .expect("Should parse next sibling")
+                    .is_none());
```
jonhoo: :(

tests/parse_self.rs
```diff
@@ -39,12 +39,14 @@ fn test_parse_self_debug_info() {
...
+        while cursor
+                  .next_dfs()
+                  .expect("Should parse next dfs")
+                  .is_some() {
             let entry = cursor.current().expect("Should have a current entry");
 
             let mut attrs = entry.attrs();
-            while let Some(_) = attrs.next().expect("Should parse entry's attribute") {
-            }
+            while let Some(_) = attrs.next().expect("Should parse entry's attribute") {}
```
jonhoo: Shouldn't this be `.is_some()`?

src/cfi.rs
```diff
@@ -629,22 +630,23 @@ impl<'bases, 'input, Endian, Section> FallibleIterator
...
-fn parse_cfi_entry_common<'input, Endian, Section>(input: EndianBuf<'input, Endian>)
-                                                   -> Result<Option<(EndianBuf<'input, Endian>,
-                                                                     (u64,
-                                                                      Format,
-                                                                      EndianBuf<'input, Endian>,
-                                                                      u64,
-                                                                      EndianBuf<'input, Endian>))>>
+/// `Ok(Some(tuple))`, where `tuple.0` is the start of the next entry and
+/// `tuple.1` is the parsed CFI entry data.
+fn parse_cfi_entry_common<'input, Endian, Section>
```
jonhoo: Curious that we only call this from one place (at least as far as I can tell from what places had to be modified to use `CfiEntryCommon`). I guess it'll be inlined though, so it's okay.

jonhoo:
Left some inline comments, but generally looks good to me. Biggest questions are:

 - Why the `rustfmt.toml` change?
 - Can we link to the worst clippy bugs so we can remove the `allow()` when they're fixed?

rustfmt.toml
philipc: I personally prefer block indent, so I'm fine with this change, and I also think the official rust style is moving towards preferring block indent too.

However, I expect the reason is that without this, rustfmt seems to be buggy and formats some stuff horribly, particularly in unit.rs. eg
````
             constants::DW_FORM_addr => {
                 return parse_address(input, unit.address_size()).map(|(rest, addr)| {
-                    let attr = Attribute {
+                                                                         let attr = Attribute {
                         name: spec.name(),
                         value: AttributeValue::Addr(addr),
                     };
-                    (rest, attr)
-                });
+                                                                         (rest, attr)
+                                                                     });
             }
````

src/abbrev.rs
philipc: Generally rustfmt ignores macros if it can't understand them. I thought it had special handling for `vec!` though. It's an improvement over what was there before at least.

src/unit.rs
philipc: Setting `chain_one_line_max = 100` would avoid this. I also like `fn_call_width = 100`.

rustfmt.toml
jonhoo: I also prefer block indent, but it does mean that we end up with block indents in a couple of places where rustfmt could choose not to wrap and indent at all. The issue you're referring to I believe was fixed by https://github.com/rust-lang-nursery/rustfmt/pull/1495, at least since then I haven't been seeing those format errors.

fitzgen:
I'm going to drop the `rustfmt` changes for now, since they seem mildly controversial, whereas teh clippy ones don't seem to be.

tests/parse_self.rs
fitzgen: I think incremental compilation + clippy results in disappearing warnings.

src/unit.rs
fitzgen: Some of the match clauses have the same arms, so clippy wants them to be a single match clause joined by `|`. That isn't appropriate in this case, IMO, although usually it is.

rustfmt.toml
fitzgen: The rustfmt.toml changes were there to prevent even more of this, although I'm not sure why this particular case still occurred.

fitzgen: Er wait. The diff @philipc posted is if you *don't* have this line, and is an example of why I did add it. Sorry misread the comments.

src/lib.rs
fitzgen: Will look into filing upstream bugs / seeing if these are known problems.

philipc:
> I'm going to drop the rustfmt changes for now, since they seem mildly controversial, whereas teh clippy ones don't seem to be.

I would like to use rustfmt regardless of whether its output has some quirks (we've lived with them in the past).  However, I tried rustfmt 0.8.4, and it introduces compilation errors now...

fitzgen:
> I would like to use rustfmt regardless of whether its output has some quirks (we've lived with them in the past).

I agree.

>  However, I tried rustfmt 0.8.4, and it introduces compilation errors now...

Yikes! I can repro, filed an issue: https://github.com/rust-lang-nursery/rustfmt/issues/1533

PR #198, opened 2017-05-09
Switch Cargo.toml's documentation link to docs.rs

fitzgen:


jonhoo:
Now that https://github.com/rust-lang/crates.io/pull/676 has landed, you can actually just leave out `documentation` entirely, and it will default to docs.rs.

fitzgen:
Going to merge this PR.

Don't care if we ultimately remove the documentation line and rely on crates.io's default behavior or not, but it seems we shouldn't point to github pages anymore either way, so this PR is an improvement over the status quo.

PR #203, opened 2017-06-06
Change AttributeValue::Data from reference to value 

philipc:
The first commit improves performance for `parse_attribute`. Without this, the second commit results in a decrease in performance, but with it we have a net gain.

The second commit changes `AttributeValue::Data` from reference to value. I did this as a step towards changing gimli to use a `Reader` trait. This should be more efficient than either allocating memory, or  storing an offset/length for something that is only up to 8 bytes, and then needing to dereference it via the Reader trait later.

For future work, there's still other reference variants in `AttributeValue` that can't be changed to inline values (`Block`, `ExprLoc` and `String`). Not sure if we have decided what direction to take for things like these: do we want to change these to offsets, or allocate memory for them immediately (possibly using `Cow`)? My preference is to use offsets, so that we're not paying an allocation for cases where we don't actually need these values. For cases where we do need them, the value should still be in the buffer of the `Reader`.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -13,8 +13,10 @@ use parser::{Error, Result, Format, DebugMacinfoOffset, parse_u8, parse_u16, par
              parse_address_as_offset, u64_to_offset};
 use ranges::DebugRangesOffset;
 use std::cell::Cell;
+use std::convert::AsMut;
 use std::ffi;
 use std::marker::PhantomData;
+use std::mem;
 use std::ops::{Range, RangeFrom, RangeTo};
 use std::{u8, u16};
 use str::{DebugStr, DebugStrOffset};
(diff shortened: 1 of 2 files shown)


src/unit.rs
```diff
@@ -908,13 +910,33 @@ pub enum AttributeValue<'input, Endian>
...
+    /// signed integer, an unsigned integer, a floating-point constant, or
+    /// anything else."
+    Data4([u8; 4]),
+
+    /// An eight byte constant data value. How to interpret the bytes depends on context.
+    ///
+    /// From section 7 of the standard: "Depending on context, it may be a
+    /// signed integer, an unsigned integer, a floating-point constant, or
+    /// anything else."
+    Data8([u8; 8]),
```
fitzgen: I don't see how we can `repr(u64)` *and* and have an eight byte sized variant *and* have room for the discriminator. The size of `AttributeValue` is surely larger than `8`, no? Am i missing something? Doing poor arithmetic?

That is, with this second commit, is `repr(u64)` still doing anything?

src/unit.rs
```diff
@@ -1542,6 +1564,20 @@ fn length_u32_value<Endian>(input: EndianBuf<Endian>)
     take(len as usize, rest)
 }
 
+fn parse_u8_array<A>(input: &[u8]) -> Result<(&[u8], A)>
+    where A: Sized + Default + AsMut<[u8]>
```
fitzgen: Ha! This works out really nicely at call sites with type inference :)

src/unit.rs
philipc: `repr(u64)` only applies to the discriminant. The total size of the enum is 24 bytes both before and after this PR.

fitzgen: Oh wow, I totally misunderstood `repr(...)` then... thanks for informing me!

PR #204, opened 2017-06-07
Expose the per-language default value of DW_AT_lower_bound.

khuey:
Some DWARF attributes are only present if they differ from the default value for that language.  `DW_AT_lower_bound` is one of these.  And rather than hardcoding these all in my debugger, they should be hardcoded somewhere in gimli.

All of these are mentioned in the DWARF spec, except Go, OCaml, Rust, Swift, Modula-3, and Julia.  I looked up (or knew) the values for those.  There are some other languages I didn't bother to find (e.g. DW_LANG_BORLAND_Delphi).  If it ever gets any consumers they can add support :)

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -794,6 +794,52 @@ dw!(DwLang(u16) {
     DW_LANG_BORLAND_Delphi = 0xb000,
 });
 
+impl DwLang {
+    /// Get the default DW_AT_lower_bound for this language.
+    pub fn default_lower_bound(&self) -> Option<usize> {
+        match *self {
+            DW_LANG_C89 |
+            DW_LANG_C |
+            DW_LANG_C_plus_plus |
+            DW_LANG_Java |
+            DW_LANG_C99 |
+            DW_LANG_ObjC |
+            DW_LANG_ObjC_plus_plus |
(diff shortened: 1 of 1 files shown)


philipc:
The DWARF 5 spec has default lower bounds for more languages.

src/constants.rs
```diff
@@ -794,6 +794,52 @@ dw!(DwLang(u16) {
...
+            DW_LANG_Pascal83 |
+            DW_LANG_Modula2 |
+            DW_LANG_Ada95 |
+            DW_LANG_Fortran95 |
+            DW_LANG_PLI |
+            DW_LANG_Modula3 |
+            DW_LANG_Julia |
+            DW_LANG_Fortran03 |
+            DW_LANG_Fortran08 => 1,
+            _ => unimplemented!(),
```
philipc: Not sure about this panic. We should either document that the function can panic, or return an `Option`.

fitzgen: +1 for returning `Option` and avoiding panics.

khuey:
Comments addressed

philipc:
Thanks!

khuey:
ty

PR #205, opened 2017-06-13
Run rustfmt 0.8.6

philipc:
Also enable the rustfmt option to convert `try!()` to `?`. Note that this does cause a small performance regression in FDE parsing (probably the same issue as https://github.com/rust-lang/rust/issues/37939).

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -27,7 +27,8 @@ pub fn read_section(section: &str) -> Vec<u8> {
 fn bench_parsing_debug_abbrev(b: &mut test::Bencher) {
     let debug_info = read_section("debug_info");
     let debug_info = DebugInfo::<LittleEndian>::new(&debug_info);
-    let unit = debug_info.units()
+    let unit = debug_info
+        .units()
         .next()
         .expect("Should have at least one compilation unit")
         .expect("And it should parse OK");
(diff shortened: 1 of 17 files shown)


benches/bench.rs
```diff
@@ -27,18 +27,19 @@ pub fn read_section(section: &str) -> Vec<u8> {
...
 
     b.iter(|| {
-        let debug_abbrev = DebugAbbrev::<LittleEndian>::new(&debug_abbrev);
-        test::black_box(unit.abbreviations(debug_abbrev)
-            .expect("Should parse abbreviations"));
-    });
+               let debug_abbrev = DebugAbbrev::<LittleEndian>::new(&debug_abbrev);
+               test::black_box(unit.abbreviations(debug_abbrev)
+                                   .expect("Should parse abbreviations"));
+           });
```
fitzgen: Bleh... I've tried to figure out how to fix this indentation behavior, haven't found anything that quite does the trick.

benches/bench.rs
```diff
@@ -152,7 +154,8 @@ fn bench_parsing_line_number_program_opcodes(b: &mut test::Bencher) {
     let debug_line = DebugLine::<LittleEndian>::new(&debug_line);
 
     b.iter(|| {
-        let program = debug_line.program(OFFSET, ADDRESS_SIZE, None, None)
+        let program = debug_line
+            .program(OFFSET, ADDRESS_SIZE, None, None)
```
fitzgen: And why do these ones not get teh same indentation? ...

philipc: This behaviour is controlled by `closure_block_indent_threshold`, which defaults to 7 lines. So closures of up to 7 lines use visual, and longer closures use block indent. I'd be happy with a setting that made it use the same indentation always.

PR #207, opened 2017-06-14
WIP Preparation for Reader trait

philipc:
I'm not sure that this PR is ready to merge yet, but I'd like feedback on the direction it is taking. My goal is to define a `Reader` trait, implement that trait for `EndianBuf`, and gradually replace all usage of `EndianBuf` with the `Reader` trait instead.

This PR takes two steps:
- use `EndianBuf` everywhere, instead of `&[u8]`. We were already using it in most places, but this extends it to things like `parse_u8`. There are likely still some places I have missed.
- reduce the amount of cloning of `EndianBuf`. This is achieved by changing parsing functions to take `&mut EndianBuf`, instead of returning a copy and leaving the input unmodified. The motivation is that while cloning `EndianBuf` is cheap, this may not be true for other implementations of `Reader`.

One reason I'm asking for feedback is that I suspect @fitzgen likes the current immutable input for parsing, so this may be a controversial change.

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -443,18 +443,9 @@ impl<'input, Endian> _UnwindSectionPrivate<'input, Endian> for EhFrame<'input, E
         debug_assert!(section.len() > 0);
         debug_assert!(input_before_offset.len() > 0);
 
-        // Additionally, the input_before_offset slice must be a subset of the
-        // section's slice.
-        debug_assert!(section.as_ptr() as usize <= input_before_offset.as_ptr() as usize);
-        debug_assert!(input_before_offset.as_ptr() as usize + input_before_offset.len() <=
-                      section.as_ptr() as usize + section.len());
-
-        let offset_ptr = (input_before_offset.as_ptr() as usize)
-            .saturating_sub(input_relative_offset);
-        if section.as_ptr() as usize <= offset_ptr &&
-           offset_ptr < section.as_ptr() as usize + section.len() {
-            let section_relative_offset = offset_ptr - section.as_ptr() as usize;
-            Some(section_relative_offset)
+        let input_offset = input_before_offset.offset_from(section);
+        if input_relative_offset <= input_offset {
+            Some(input_offset - input_relative_offset)
         } else {
             None
         }
@@ -1121,7 +1112,7 @@ impl<'input, Endian, Section> FrameDescriptionEntry<'input, Endian, Section>
 
         {
             let mut func = bases.func.borrow_mut();
-            let offset = rest.as_ptr() as usize - section.section().as_ptr() as usize;
+            let offset = rest.offset_from(section.section());
             *func = Some(offset as u64);
         }
 
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -259,8 +259,8 @@ pub trait UnwindSection<'input, Endian>
             return Err(Error::UnexpectedEof);
         }
 
-        let input = self.section().range_from(offset..);
-        if let Some((_, entry)) = CommonInformationEntry::parse(bases, *self, input)? {
+        let input = &mut self.section().range_from(offset..);
+        if let Some(entry) = CommonInformationEntry::parse(bases, *self, input)? {
             Ok(entry)
         } else {
             Err(Error::NoEntryAtGivenOffset)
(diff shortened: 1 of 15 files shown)


fitzgen:
Sorry I've been putting this off -- I'll try to take a look tomorrow.

src/endianity.rs
```diff
@@ -115,15 +115,27 @@ pub type NativeEndian = BigEndian;
...
 /// A `&[u8]` slice with compile-time endianity metadata.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct EndianBuf<'input, Endian>(pub &'input [u8], pub PhantomData<Endian>)
-    where Endian: Endianity;
+pub struct EndianBuf<'input, Endian>
+    where Endian: Endianity
+{
+    buf: &'input [u8],
+    endian: PhantomData<Endian>,
+}
```
fitzgen: These should have stopped being tuple structs a long time ago -- thanks!

src/endianity.rs
```diff
@@ -252,6 +255,17 @@ impl<'input, Endian> Into<&'input [u8]> for EndianBuf<'input, Endian>
     }
 }
 
+impl<'input, Endian> Read for EndianBuf<'input, Endian>
+    where Endian: Endianity
+{
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
```
fitzgen: >Takes a performance hit though, so may want to inline it later.

Do you mean _manually_ inlining this code? Did you try adding `#[inline]` here? Any difference?

src/cfi.rs
```diff
@@ -443,18 +443,9 @@ impl<'input, Endian> _UnwindSectionPrivate<'input, Endian> for EhFrame<'input, E
...
-        debug_assert!(section.as_ptr() as usize <= input_before_offset.as_ptr() as usize);
-        debug_assert!(input_before_offset.as_ptr() as usize + input_before_offset.len() <=
-                      section.as_ptr() as usize + section.len());
-
-        let offset_ptr = (input_before_offset.as_ptr() as usize)
-            .saturating_sub(input_relative_offset);
-        if section.as_ptr() as usize <= offset_ptr &&
-           offset_ptr < section.as_ptr() as usize + section.len() {
-            let section_relative_offset = offset_ptr - section.as_ptr() as usize;
-            Some(section_relative_offset)
```
fitzgen: 👍 💯

fitzgen:
> I'd like to get your opinion on the immutable input change at least, specifically changing from:
> 
> `let (rest, val) = parse_xxx(rest)?;`
> 
> to
> 
> `let val = parse_xxx(rest)?;`
> 
> I understand that the former style is great for a backtracking parser, but I'm not aware of any benefit it provides for DWARF parsing, and I think it makes the code more complicated. So I'm not sure if the code is written like this simply as a leftover from using nom, or if you prefer it this way.

In general, I strongly prefer immutability and referential transparency when all else is equal. However, we're moving towards a world where all else is *not* equal, we will have rampant mutation of the reader while parsing, and trying to paper over that would not be fun.

I also tried thinking of a way to split handles/cursors into the image being read out of the reader itself, and that the reader can be indexed with those handles/cursors or something like that. The idea being that this would allow us to take and return these handles/cursors in a referentially transparent way like we do with slices now. In the end, I think it would be a lot of complexity that wouldn't end up paying for itself, and the mutation is still happening under the covers so if our abstractions leak at all, then it will be blech.

In conclusion, I think the changes in this PR make sense, and this is the road we want to continue down.

> 
> Having thought about it more, I'm not sure that the clone cost is going to be important, so don't weight that too heavily in making your decision. Even for the buffered implementation of Reader, the struct it needs to clone should still be small, eg something like { image: &'input Image, offset: usize, len: usize }. I can imagine some consumers wanting to use Rc<Image> or Arc<Image>.

Yes, my intuition is that these clones shouldn't have any observable overhead compared to parsing each sections contents, but measurement is king.

fitzgen:
Thanks again @philipc !

src/endianity.rs
philipc: Adding `#[inline]` here made no difference. I meant moving the functions from the leb128 crate into gimli, 'inline' is probably the wrong term. We already have an impedance mismatch with the `Error` types, and now we're going to be adding a `Reader` trait that doesn't match the `io::Read` that leb128 wants. I'm not looking into performance much until this is all done, but my feeling is that if we want leb128 parsing to be fast then we'll need an implementation that more closely matches our needs.

Issue #84, opened 2016-08-30
Add support for dynamic endianity


fitzgen:
@philipc says:

> I think we still need something that is runtime selectable as well though. https://github.com/gimli-rs/gimli/blob/master/examples/dwarfdump.rs#L79-L83 is fine for simple utilities, but we need some way to stop the endianity bubbling up through all the APIs.

fitzgen:
@philipc: What if we made `BigEndian` and `LittleEndian` into zero sized structs with no members and then made `DynamicEndian` an enum between the two?

```rust
#[derive(all the things)]
pub struct CompileTimeEndian<B: byteorder::ByteOrder>(PhantomData<B>);

impl<B: byteorder::ByteOrder> byteorder::ByteOrder for CompileTimeEndian<B> {
    // ...
}

pub type BigEndian = CompileTimeEndian<byteorder::BigEndian>;
pub type LittleEndian = CompileTimeEndian<byteorder::LittleEndian>;

pub enum RunTimeEndian {
    Big(BigEndian),
    Little(LittleEndian),
}

impl byteorder::ByteOrder for RunTimeEndian {
    // ...
}
```

fitzgen:
I think this means we will need all the various `read_u64` methods to take `&self`, so we can't just use `ByteOrder` instead of `Endianity` as #155 proposes.

philipc:
> What if we made BigEndian and LittleEndian into zero sized structs with no members and then made DynamicEndian an enum between the two?

Yes, that is what I was thinking, and agree that it means we still need `Endianity`.

PR #209, opened 2017-06-26
Define Reader trait and start converting modules to use it

philipc:
So far I've only converted abbrev/unit/str.  The unit module is one of the more complicated ones, so I think this should be enough to decide whether this trait is suitable.

Some outstanding flaws:
- requires `Copy` and `Eq`, but I think we should work to avoid that later
- each `Reader` impl currently requires its own definition for `read_u64` etc, but we can avoid that later as part of adding support for dynamic endianity (#84)

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1,0 +1,136 @@
+use std::fmt::Debug;
+use std::io;
+use std::io::Read;
+
+use endianity::Endianity;
+use leb128;
+use parser::{Error, Result, Format, u64_to_offset};
+
+/// A trait for reading the data from a DWARF section.
+///
+/// All read operations advance the section offset of the reader
+/// unless specified otherwise.
+///
+// TODO: Don't require Copy, Eq
+pub trait Reader: Debug + Clone + Copy + PartialEq + Eq + Read {
+    /// The endianity of bytes that are read.
+    type Endian: Endianity;
+
+    /// Return the number of bytes remaining.
+    fn len(&self) -> usize;
+
+    /// Return true if the number of bytes remaining is zero.
+    fn is_empty(&self) -> bool;
+
+    /// Set the number of bytes remaining to zero.
+    fn empty(&mut self);
+
+    /// Set the number of bytes remaining to the specified length.
+    fn truncate(&mut self, len: usize) -> Result<()>;
+
+    /// Return the offset of this reader's data relative to the start of
+    /// the given base reader's data.
+    ///
+    /// May panic if this reader's data is not contained within the given
+    /// base reader's data.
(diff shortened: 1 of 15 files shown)


fitzgen:
> So far I've only converted abbrev/unit/str. The unit module is one of the more complicated ones, so I think this should be enough to decide whether this trait is suitable.

Yep, makes sense.

> requires Copy and Eq, but I think we should work to avoid that later

Interesting, I wouldn't have expected this (but I'm not terribly surprised). I suppose I'll find out more when I dig into the patch.

philipc:
> Interesting, I wouldn't have expected this (but I'm not terribly surprised). I suppose I'll find out more when I dig into the patch.

Ah yeah, I should have mentioned where they are needed.

`Copy` is required for `after_attrs: Cell<Option<R>>`. This one is probably going to be tricky to avoid.

`Eq` is needed for the assert at https://github.com/gimli-rs/gimli/blob/b5258a1fc9c58a33eef49f9e1560ed77dc404f08/src/unit.rs#L1754. We can probably just drop that assert, but I want to finish converting the rest to see if there are any other cases.

philipc:
> How much mileage do you think we would get out of requiring Seek as a super trait for Reader? It seems to me that it is essentially Endianity + Read + Seek, but we have our own versions of Seek functions just sort of.

The main problem with that is `Seek` doesn't provide `tell` functionality. We also don't need to be quite as generic with the seeking (eg we don't need to seek to the start, and that would be hard implement with `EndianBuf` without introducing another offset field in it).

src/reader.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// A trait for reading the data from a DWARF section.
+///
+/// All read operations advance the section offset of the reader
+/// unless specified otherwise.
+///
+// TODO: Don't require Copy, Eq
+pub trait Reader: Debug + Clone + Copy + PartialEq + Eq + Read {
+    /// The endianity of bytes that are read.
+    type Endian: Endianity;
```
philipc: My motivation was that in general, users of `Reader` don't care what the endianity is. It is an implementation detail that should be transparent to them. Using an associated type avoids the noise of declaring the endianity everywhere that `Reader` is used.

I don't think that restricting a type to a single implementation of `Endianity` is a problem. For example, `EndianBuf<LittleEndian>` can only work with `LittleEndian`, not any arbitrary endianity. For types that support dynamic endianity, the implementation of endianity that they use will be `DynamicEndianity` or something like that.

philipc:
Thanks for the review. I'm going to hold off merging this so that master isn't left sitting in a partially converted state.

PR #213, opened 2017-07-01
Fix pubnames/pubtypes offset handling

philipc:
- add die_offset() methods
- pubtypes is still for .debug_info, not .debug_types
- improve parse_self tests
- add dwarfdump support


philipc:
Also added a commit to avoid Rc usage. No significant difference to benchmarks, but I think it makes it simpler.

PR #214, opened 2017-07-03
Convert remaining modules to use Reader

philipc:
Also avoid `Copy`/`Eq` requirement by using offsets.

Next step after this is to convert dwarfdump to use `Reader`. Probably need to add `to_vec` and `to_string` methods to the `Reader` trait to allow this.

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -766,7 +766,7 @@ impl Default for Augmentation {
 }
 
 impl Augmentation {
-    fn parse<'aug, 'bases, 'input, Endian, Section>(augmentation_str: &'aug str,
+    fn parse<'aug, 'bases, 'input, Endian, Section>(augmentation_str: &'aug [u8],
                                                     bases: &'bases BaseAddresses,
                                                     address_size: u8,
                                                     section: Section,
@@ -778,12 +778,12 @@ impl Augmentation {
         debug_assert!(!augmentation_str.is_empty(),
                       "Augmentation::parse should only be called if we have an augmentation");
 
-        let mut chars = augmentation_str.chars();
+        let mut chars = augmentation_str.iter();
 
         let first = chars
             .next()
-            .expect("Is valid UTF-8 and length > 0, so must have at least one char.");
-        if first != 'z' {
+            .expect("length > 0, so must have at least one char.");
+        if *first != b'z' {
             return Err(Error::UnknownAugmentation);
         }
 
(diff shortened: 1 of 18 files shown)


src/loc.rs
```diff
@@ -28,14 +27,26 @@ impl<'input, Endian> DebugLoc<'input, Endian>
...
+        Self::from_reader(EndianBuf::new(debug_loc_section))
+    }
+}
+
+impl<R: Reader> DebugLoc<R> {
+    /// Construct a new `DebugLoc` instance from the data in the `.debug_loc`
+    /// section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_loc` section and
+    /// present it as a `&[u8]` slice. That means using some ELF loader on
```
fitzgen: ditto

src/ranges.rs
```diff
@@ -27,14 +26,26 @@ impl<'input, Endian> DebugRanges<'input, Endian>
...
+        Self::from_reader(EndianBuf::new(debug_ranges_section))
+    }
+}
+
+impl<R: Reader> DebugRanges<R> {
+    /// Construct a new `DebugRanges` instance from the data in the `.debug_ranges`
+    /// section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_ranges` section and
+    /// present it as a `&[u8]` slice. That means using some ELF loader on
```
fitzgen: ditto

fitzgen:
> I have a concern about the first commit

Ah, OK I see now that this was just a temporary thing until it became a `&mut R` in the next commit, which I guess I glazed over.

LGTM, r+ with the nitpicks on copy-paste

Thanks a third time! :)

src/cfi.rs
```diff
@@ -766,7 +766,7 @@ impl Default for Augmentation {
 }
 
 impl Augmentation {
-    fn parse<'aug, 'bases, 'input, Endian, Section>(augmentation_str: &'aug str,
+    fn parse<'aug, 'bases, 'input, Endian, Section>(augmentation_str: &'aug [u8],
```
philipc: So that we can trivially convert this to use the `Reader` trait. Since all of the characters that we understand currently are single-byte, it doesn't matter if we treat it as `&[u8]`. If we do need to handle multi-byte UTF-8 characters in future, we'll need to either potentially read the whole string into a `Vec` and then validate it as a UTF-8 string, or add a `read_utf8_char` method to the trait.

One thing we do lack from this approach is that we no longer validate that unknown augmentation strings are UTF-8. I don't see that as much of a problem though, since the user will handle this when they convert the `Reader` into a string.

PR #216, opened 2017-07-06
Add leb128 module and stop using io::Read

philipc:
This recovers some of the performance we lost during the `Reader` conversion (numbers in commit).

### DIFF
--- a/src/leb128.rs
+++ b/src/leb128.rs
@@ -7,7 +7,7 @@
 //! Read and write signed integers:
 //!
 //! ```
-//! use gimli::leb128;
+//! use gimli::{EndianBuf, NativeEndian, leb128};
 //!
 //! let mut buf = [0; 1024];
 //!
@@ -17,8 +17,8 @@
 //!     leb128::write::signed(&mut writable, -12345).expect("Should write number");
 //! }
 //!
-//! // Read from anything that implements `std::io::Read`.
-//! let mut readable = &buf[..];
+//! // Read from anything that implements `gimli::Reader`.
+//! let mut readable = EndianBuf::<NativeEndian>::new(&buf[..]);
 //! let val = leb128::read::signed(&mut readable).expect("Should read number");
 //! assert_eq!(val, -12345);
 //! ```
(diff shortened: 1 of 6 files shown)


src/leb128.rs
```diff
@@ -0,0 +1,512 @@
...
+
+#[doc(hidden)]
+#[inline]
+pub fn low_bits_of_byte(byte: u8) -> u8 {
+    byte & !CONTINUATION_BIT
+}
+
+#[doc(hidden)]
+#[inline]
+pub fn low_bits_of_u64(val: u64) -> u8 {
```
fitzgen: All of these `doc(hidden)` things can be `pub(crate)` now, instead. Obviously doesn't need to block anything from landing.

src/leb128.rs
```diff
@@ -35,27 +35,23 @@
...
 //! ```
 
 use std;
 
-#[doc(hidden)]
-pub const CONTINUATION_BIT: u8 = 1 << 7;
-#[doc(hidden)]
-pub const SIGN_BIT: u8 = 1 << 6;
+const CONTINUATION_BIT: u8 = 1 << 7;
+const SIGN_BIT: u8 = 1 << 6;
```
fitzgen: Ok, there we go :)

src/reader.rs
```diff
@@ -114,26 +112,12 @@ pub trait Reader: Debug + Clone + Read {
...
     fn read_sleb128(&mut self) -> Result<i64> {
-        match leb128::read::signed(self) {
-            Ok(val) => Ok(val),
-            Err(leb128::read::Error::IoError(ref e)) if e.kind() ==
-                                                        io::ErrorKind::UnexpectedEof => {
-                Err(Error::UnexpectedEof)
-            }
-            Err(_) => Err(Error::BadSignedLeb128),
-        }
+        leb128::read::signed(self)
```
fitzgen: Might make sense for these functions to either go away or be `#[inline]` too now (although I suspect rustc/llvm is doing the expected inlining anyways).

PR #217, opened 2017-07-09
Convert Section trait to use Reader

philipc:
I tried to keep the `From<&[u8]>`, but couldn't get it to work, eg something like:

```
// error[E0207]: the type parameter `Endian` is not constrained by the impl trait, self type, or predicates
impl<'input, Endian, T> From<&'input [u8]> for T
    where T: Section<EndianBuf<'input, Endian>>,
          Endian: Endianity
{
    fn from(v: &'input [u8]) -> Self {
        Self::from(EndianBuf::new(v))
    }
}
```

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -107,26 +107,25 @@ fn main() {
 fn dump_file<Endian>(file: &object::File, flags: &Flags)
     where Endian: gimli::Endianity
 {
-    let debug_abbrev = file.get_section(".debug_abbrev").unwrap_or(&[]);
-    let debug_abbrev = &gimli::DebugAbbrev::<gimli::EndianBuf<Endian>>::new(debug_abbrev);
-    let debug_aranges = file.get_section(".debug_aranges").unwrap_or(&[]);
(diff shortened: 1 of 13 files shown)


examples/dwarfdump.rs
```diff
@@ -107,26 +107,25 @@ fn main() {
...
+    fn load_section<'input, 'file, S, Endian>(file: &'file object::File<'input>) -> S
+        where S: gimli::Section<gimli::EndianBuf<'input, Endian>>,
+              Endian: gimli::Endianity,
+              'file: 'input
+    {
+        let data = file.get_section(S::section_name()).unwrap_or(&[]);
+        S::from(gimli::EndianBuf::new(data))
+    }
+
+    let debug_abbrev: &gimli::DebugAbbrev<_> = &load_section::<_, Endian>(file);
```
fitzgen: Why does this one need a turbo fish but the others don't?

philipc: `Endian` needs to be specified at least once, because it can't be inferred. But specifying it once is enough to let the rest be inferred.

PR #222, opened 2017-07-14
Add self tests and benchmarks for parsing and evaluating expressions

philipc:
Also added an `Expression` newtype, to make it easier for a user to find out what to do with the contents of `AttributeValue::Exprloc` and `LocationListEntry`. Maybe we want to have separate newtypes for expressions and locations? Not sure if it adds any value.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -3,8 +3,9 @@
 extern crate gimli;
 extern crate test;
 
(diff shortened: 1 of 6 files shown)


philipc:
The expression evaluations are dominated by allocations. Maybe https://github.com/servo/rust-smallvec is worth using.

fitzgen:
> The expression evaluations are dominated by allocations. Maybe https://github.com/servo/rust-smallvec is worth using.

👍

PR #224, opened 2017-07-20
EntriesTree improvements

philipc:
- Move DIE parse code from `EntriesCursor` to `DebuggingInformationEntry`, needed for following changes (plus makes more sense).
- Don't use `EntriesCursor` in the `EntriesTree` implementation. This makes the code easier to understand. Also gives a performance improvement.
- Add `EntriesTreeNode`, so that users can't wrongly try to access the entry after iterating children.

### DIFF
--- a/src/unit.rs
+++ b/src/unit.rs
@@ -2102,9 +2102,9 @@ impl<'abbrev, 'unit, R: Reader> EntriesCursor<'abbrev, 'unit, R> {
 /// The state information for a tree view of the Debugging Information Entries.
 ///
 /// The `EntriesTree` can be used to recursively iterate through the DIE
-/// tree, following the parent/child relationships. It maintains a single
-/// `EntriesCursor` that is used to parse the entries, allowing it to avoid
-/// any duplicate parsing of entries.
+/// tree, following the parent/child relationships. The `EntriesTree` contains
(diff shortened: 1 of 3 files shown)


src/unit.rs
```diff
@@ -824,6 +824,26 @@ impl<'abbrev, 'unit, R: Reader> DebuggingInformationEntry<'abbrev, 'unit, R> {
...
+
+    /// Parse an entry. Returns `Ok(None)` for null entries.
+    fn parse(input: &mut R,
+             unit: &'unit UnitHeader<R>,
+             abbreviations: &'abbrev Abbreviations)
+             -> Result<Option<Self>> {
+        let offset = unit.header_size() + input.offset_from(&unit.entries_buf);
+        let code = input.read_uleb128()?;
+        if code == 0 {
+            return Ok(None);
```
fitzgen: Slight aside: Does it make sense to delay calculating the offset until after reading the code? Maybe the compiler can reorder these operations anyways, and it isn't worth it.

philipc: The problem is we want the offset from before the `read_uleb128`, which is a variable length. Do you see a good way to do that?

fitzgen: Oh hrm yeah I totally overlooked that. Don't see a good way around it, no.

PR #225, opened 2017-07-28
Add Reader::Offset associated type

philipc:
This allows 64-bit readers to be used on 32-bit systems.

One annoyance is that in a few places I've had to use:
 `struct S<R: Reader<Offset=Offset>, Offset: ReaderOffset> { ... }`
instead of `struct S<R: Reader>`, due to rust's variance rules (specifically, structs containing associated types are invariant). This manifested as lifetime errors in the tests. I've only used this form in places where the tests required it. I'm not sure if this error is something that will occur in practice. If it is, then there may be more places that we need to convert to use it.

There is also a performance regression for `bench_parsing_debug_info_tree`. I think this is due to changes in what is inlined, but I haven't tracked it down fully. I'm not happy with how dependent the benchmarks are on inlining; it means that consumers of this crate may not be getting the same performance as our benchmarks.

### DIFF
--- a/src/reader.rs
+++ b/src/reader.rs
@@ -1,24 +1,199 @@
+use std;
 use std::borrow::Cow;
 use std::fmt::Debug;
+use std::ops::{Add, AddAssign, Sub};
 
 use endianity::Endianity;
 use leb128;
-use parser::{Error, Result, Format, u64_to_offset};
+use parser::{Error, Result, Format};
+
+/// A trait for offsets with a DWARF section.
+///
+/// This allows consumers to choose a size that is appropriate for their address space.
+pub trait ReaderOffset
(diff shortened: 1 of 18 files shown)


src/reader.rs
```diff
@@ -142,8 +317,8 @@ pub trait Reader: Debug + Clone {
         }
     }
 
-    /// Parse a word-sized integer according to the DWARF format, and return it as a `usize`.
-    fn read_offset(&mut self, format: Format) -> Result<usize> {
-        self.read_word(format).and_then(u64_to_offset)
+    /// Parse a word-sized integer according to the DWARF format, and return it an offset.
```
fitzgen: typo: return it **as** an offset

PR #230, opened 2017-08-02
Reader improvements

philipc:
This is the result of an experiment in implementing `Reader` for:
```rust
pub struct EndianRead<T, Endian>
    where T: Read + Seek + Clone,
          Endian: Endianity
{
    buf: T,
    position: u64,
    len: u64,
    endian: Endian,
}
```

I looked into benchmarking this for `std::io::File`, but I didn't think it was going to be worth it without implementing buffering, and the `Clone` requirement also makes it a bit more complicated (eg we can't just use `std::io::BufReader`).

I did check that dwarfdump produces identical output when using `EndianRead<io::Cursor<&'input [u8]>, Endian>` instead of `EndianBuf`.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -2,6 +2,7 @@
 
 use std::error;
 use std::fmt::{self, Debug};
+use std::io;
 use std::result;
 use cfi::BaseAddresses;
 use constants;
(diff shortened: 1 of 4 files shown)


src/parser.rs
```diff
@@ -10,6 +11,8 @@ use reader::{Reader, ReaderOffset};
 /// An error that occurred when parsing.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Error {
+    /// An I/O error occurred while reading.
+    Io,
```
fitzgen: Would it be super difficult to remove the `derive(Copy)` so we could embed the `io::Error` inside here? Just a thought, not a blocker for anything.

philipc: It's not just `Copy` that is the problem. We would have to remove, `Clone`, `PartialEq` and `Eq` too. We currently need `Clone` or `Copy` for returning errors that are stored in `EvaluationState`; not sure if that is fixable. We use `Eq` in our tests, but that could be avoided by using `match` instead. It's a bigger change than I want to do right now though.

PR #233, opened 2017-08-09
Add EndianBuf::to_string()

philipc:


src/endianity.rs
```diff
@@ -241,6 +241,17 @@ impl<'input, Endian> EndianBuf<'input, Endian>
         ptr - base_ptr
     }
 
+    /// Converts the buffer to a string using `str::from_utf8`.
+    ///
+    /// Returns an error if the buffer contains invalid characters.
+    #[inline]
+    pub fn to_string(&self) -> Result<&'input str> {
+        match str::from_utf8(self.buf) {
+            Ok(s) => Ok(s.into()),
```
fitzgen: Why is `into()` needed here?

philipc: Because copy/paste, thanks. I switched to `map_err`.

Issue #231, opened 2017-08-02
Improve error handling in dwarfdump


philipc:
Using `expect` and `unwrap` is a poor example.

Issue #140, opened 2016-10-06
Support DWARF version 5 operations and corresponding GNU extensions


philipc:
My rust binaries are linked with objects that use `DW_OP_GNU_entry_value`. Now that gimli's dwarfdump is displaying locs, it panics when trying to dump my rust binaries.

I think I've seen `DW_OP_GNU_implicit_pointer` in use somewhere too.

DWARF 5 operations:
- [x] DW_OP_implicit_pointer
- [x] DW_OP_addrx
- [x] DW_OP_constx
- [x] DW_OP_entry_value
- [x] DW_OP_const_type
- [x] DW_OP_regval_type
- [x] DW_OP_deref_type
- [x] DW_OP_xderef_type
- [x] DW_OP_convert
- [x] DW_OP_reinterpret

tromey:
`dwarf2.def` lists the GNU extensions and in most cases provides a link to documentation:

https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=include/dwarf2.def;h=64850714ca436e988a11dfeef4d22ea0dc437c59;hb=HEAD#l598

Not all of these are in active use.  `DW_OP_GNU_encoded_addr` isn't implemented by gdb, so probably not used.  IIRC `DW_OP_GNU_uninit` isn't generated by GCC by default.  The others can be generated though.  Note that the typed stack stuff will need relatively large changes to the evaluator.

tromey:
Also, I think it would be better not to panic when dumping something unknown.

philipc:
> Also, I think it would be better not to panic when dumping something unknown.

Yeah, gimli's dwarfdump currently uses `expect` everywhere. This is fine as a validation tool. Definitely should be fixed if dwarfdump is going to see general usage, but that's not a high priority goal for me now (and if that were the goal, then I would probably stop trying to clone libdwarf's dwarfdump output so closely).

tromey:
The DWARF 5 draft is out now and promotes these to standard.

philipc:
`DW_OP_implicit_pointer` and `DW_OP_entry_value` implemented in #150

tromey:
I started work on this here: https://github.com/tromey/gimli/tree/dwarf-5-eval.  But it needs a rebase and isn't complete.

philipc:
Typed operations were implemented in #293. `DW_OP_addrx ` was started in #242.

philipc:
[...]

PR #234, opened 2017-08-10
Improve error handling in dwarfdump

bheisler:
I've changed all of the expect() calls, but I left the unwraps since it doesn't look like there's any way they could actually fail in practice.

I wasn't entirely sure what approach to take here since issue #231 is pretty low on details, so I hope this is OK. If not, let me know what you're looking for and I'll try to fix it up.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -20,7 +20,6 @@ use std::fmt::{self, Debug};
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Error {
     GimliError(gimli::Error),
-    MissingFileIndex,
     IoError,
     MissingDIE,
 }
@@ -36,7 +35,6 @@ impl error::Error for Error {
     fn description(&self) -> &str {
         match *self {
             Error::GimliError(ref err) => err.description(),
-            Error::MissingFileIndex => "Expected a file index but none was found",
             Error::IoError => "An I/O error occurred while reading.",
             Error::MissingDIE => "Expected a DIE but none was found",
         }
@@ -386,7 +384,10 @@ fn dump_entries<R: Reader>(offset: R::Offset,
             if flags.raw {
                 println!("{:?}", attr.raw_value());
             } else {
-                dump_attr_value(&attr, &unit, debug_loc, debug_ranges, debug_str)?;
+                match dump_attr_value(&attr, &unit, debug_loc, debug_ranges, debug_str) {
+                    Ok(_) => (),
+                    Err(ref err) => println!("Failed to dump attribute value: {}", error::Error::description(err)),
+                };
             }
         }
     }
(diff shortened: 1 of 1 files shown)


examples/dwarfdump.rs
```diff
@@ -315,22 +373,23 @@ fn dump_entries<R: Reader>(offset: R::Offset,
...
 
         let mut attrs = entry.attrs();
-        while let Some(attr) = attrs.next().expect("Should parse attribute OK") {
+        while let Some(attr) = attrs.next()? {
             print!("{:indent$}{:27} ", "", attr.name(), indent = indent + 18);
             if flags.raw {
                 println!("{:?}", attr.raw_value());
             } else {
-                dump_attr_value(&attr, &unit, debug_loc, debug_ranges, debug_str);
+                dump_attr_value(&attr, &unit, debug_loc, debug_ranges, debug_str)?;
```
philipc: It would be nice if we printed the error and continued dumping in cases where the error doesn't prevent it, such as here.

examples/dwarfdump.rs
```diff
@@ -896,22 +954,22 @@ fn dump_pubnames<R: Reader>(debug_pubnames: &gimli::DebugPubNames<R>,
                  cu_offset.0,
                  pubname.name().to_string_lossy().unwrap())
```
philipc: The function is already returning a `Result`, so I would remove the unwrap here too, even though it can't fail for `EndianBuf`. It'll get optimized away, and it makes it easier when looking for other places that unwrap. There might be a few more unwraps like this that you could remove too.

examples/dwarfdump.rs
```diff
@@ -507,34 +568,35 @@ fn dump_type_signature<Endian>(signature: gimli::DebugTypeSignature, endian: End
...
-        return;
+        return Ok(());
     }
     let header = match unit.line_program {
         Some(ref program) => program.header(),
-        None => return,
+        None => return Ok(()),
     };
-    let file = header.file(file).expect("File index should be valid");
+    let file = header.file(file).ok_or(Error::MissingFileIndex)?;
```
philipc: We could print an error here instead of returning it.

examples/dwarfdump.rs
bheisler: How can I tell which errors should continue dumping and which ones shouldn't?

philipc: We can continue dumping if the iterator doesn't need the result of the operation to find the next item. That's going to need a bit of knowledge of the internals to figure out though. The main other spot that I think is useful are the loops for `debug_info.units()` and `debug_types.units()`. There's possibly a couple more similar to these (anything that uses `parse_initial_length` internally) but it's not worth spending a lot of time on.

examples/dwarfdump.rs
```diff
@@ -633,7 +633,7 @@ fn dump_exprloc<R: Reader>(data: &gimli::Expression<R>, unit: &Unit<R>) -> Resul
     Ok(())
 }
 
-fn dump_op<R: Reader>(dwop: gimli::DwOp, op: gimli::Operation<R, R::Offset>, newpc: &R) {
+fn dump_op<R: Reader>(dwop: gimli::DwOp, op: gimli::Operation<R, R::Offset>, newpc: &R) -> Result<()> {
```
philipc: This function needs `Ok(())` at the end now.

bheisler: Well, that's embarassing. Should be fixed now.

examples/dwarfdump.rs
bheisler: OK, thanks. I'll take a look at that tomorrow if I have time.

bheisler:
Thanks for responding quickly. It was a pleasure working with you.

fitzgen:
@bheisler thanks for the patch :) Interested in hacking on gimli some more? We could use some help with https://github.com/gimli-rs/gimli/issues/140 if you're interested

PR #238, opened 2017-08-18
Bump to 0.14.1

fitzgen:
r? @philipc


fitzgen:
Just realized I need a couple more things. Will close this, make PRs for those other things, and then make a new version bump PR.

PR #239, opened 2017-08-21
Display more operations in dwarfdump

philipc:


tromey:
FWIW, looks reasonable to me.

philipc:
Thanks @tromey

Issue #244, opened 2017-09-22
libobjc.A.dylib has `__eh_frame` CIE with version 3


fitzgen:
My understanding is that the only valid version for `eh_frame` is 1. IIRC, 3 is a valid version for `debug_frame` though. I wonder if this is just a buggy unwind info producer?

Here is what binutils' `dwarfdump` gives on `libobjc.A.dylib`:

```
0x00000270: CIE
        length: 0x00000014
        CIE_id: 0xffffffff
       version: 0x03
  augmentation: "zR"
    code_align: 1
    data_align: -8
   ra_register: 0x10
 aug_arg["zR"]: DW_GNU_EH_PE_pcrel + DW_GNU_EH_PE_absptr
                DW_CFA_def_cfa (rsp, 8)
                DW_CFA_offset (rip, -8)
                DW_CFA_nop
                DW_CFA_nop
  Instructions: Init State: CFA=rsp+8     rip=[rsp]
```

Will upload a copy of the section in a second.

fitzgen:
From https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-PDA/LSB-PDA/ehframechpt.html

> Version
>
>    Version assigned to the call frame information structure. This value shall be 1.

:-/

fitzgen:
> IIRC, 3 is a valid version for debug_frame though. I wonder if this is just a buggy unwind info producer?

`debug_frame` CIEs in DWARF 3 use version 3.

fitzgen:
The funky thing is, there are tons of CIEs in this section, but only one uses version 3.

philipc:
Are we decoding the version 1 CIEs correctly? Looking at the gdb source (gdb/dwarf2-frame.c), the version field is handled identically for eh_frame and debug_frame. So I suspect our current decoding of the return address register is wrong for eh-frame version 1.

fitzgen:
> Are we decoding the version 1 CIEs correctly?

Nope, no one is using our CFI stuff yet, AFAIK, so I don't have much faith ;)

Note that the `dwarfdump` in that command is the binutils'.

Will look into what GDB is doing when I circle back here.

PR #246, opened 2017-09-22
WIP eh_frame CIE with version 3

fitzgen:
This is the `__eh_frame` section of my `/usr/lib/libobjc.A.dylib`.

See #244 

WIP pull request until we figure out what to do with this.


philipc:
That file is full of null bytes.

fitzgen:
D'oh not sure what I did... Going to just close this for now.

PR #250, opened 2017-10-04
Initial .eh_frame_hdr implementation

main--:


src/cfi.rs
```diff
@@ -89,6 +90,149 @@ impl<R: Reader> From<R> for DebugFrame<R> {
...
+        assert_eq!(version, 1); // FIXME return error
+
+        let eh_frame_ptr_enc = DwEhPe(reader.read_u8()?);
+        let fde_count_enc = DwEhPe(reader.read_u8()?);
+        let table_enc = DwEhPe(reader.read_u8()?);
+        let eh_frame_ptr = parse_encoded_pointer(eh_frame_ptr_enc, bases, addr_size, &self.0, &mut reader)?;
+        let fde_count = parse_encoded_pointer(fde_count_enc, bases, addr_size, &self.0, &mut reader)?;
+        let fde_count = match fde_count {
+            Pointer::Direct(c) => c,
+            Pointer::Indirect(_) => unreachable!(),
```
fitzgen: Is this really unreachable? I would expect there to be some early return if `fde_count_enc.is_indirect()` in order to make this actually unreachable, and not just a bad `.eh_frame_hdr`.

src/cfi.rs
```diff
@@ -89,6 +90,149 @@ impl<R: Reader> From<R> for DebugFrame<R> {
...
+}
+
+impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
+    /// Returns a pointer to the FDE for the given address.
+    ///
+    /// This performs a binary search.
+    pub fn lookup(&self, address: u64, bases: &BaseAddresses) -> Result<Pointer> {
+        let size = match self.hdr.table_enc.format() {
+            constants::DW_EH_PE_uleb128 | constants::DW_EH_PE_sleb128
+                => return Err(Error::VariableLengthSearchTable),
```
fitzgen: I don't see anything in https://refspecs.linuxfoundation.org/LSB_1.3.0/gLSB/gLSB/ehframehdr.html that suggests that using LEBs is not valid here, although it does break binary search, as you point out. How about a comment to the effect that it doesn't make sense because of the binary search stuff, but that it isn't strictly prohibited.

src/cfi.rs
```diff
@@ -89,6 +90,149 @@ impl<R: Reader> From<R> for DebugFrame<R> {
...
+            let tail = reader.clone();
+
+            let pivot = parse_encoded_pointer(self.hdr.table_enc, bases, self.hdr.addr_size,
+                                              &self.hdr.section, &mut reader)?;
+            let pivot = match pivot {
+                Pointer::Direct(x) => x,
+                Pointer::Indirect(_) => return Err(Error::UnsupportedPointerEncoding),
+            };
+
+            match pivot.cmp(&address) {
```
fitzgen: I guess the reason that we can't read the table into a slice and use `slice::binary_search_by` is because we would need to repeat that code for all the different sizes?

I'd prefer to use the off-the-shelf binary search implementation if possible.

fitzgen:
Oh, and if you have any questions about `test-assembler` or anything like that, don't hesitate to ask me :)

src/cfi.rs
fitzgen: Perhaps some macro_rules to hide the code repetitition?

src/cfi.rs
```diff
@@ -89,6 +90,149 @@ impl<R: Reader> From<R> for DebugFrame<R> {
     }
 }
 
+/// `EhFrameHdr` contains the information about the `.eh_frame` section.
```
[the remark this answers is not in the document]
fitzgen: Oh, also please add this type to the list of sections / API entry points in the top level crate docs in src/lib.rs! Thanks!

src/cfi.rs
```diff
@@ -89,6 +90,149 @@ impl<R: Reader> From<R> for DebugFrame<R> {
...
+///
+/// A pointer to the start of the `.eh_frame` data, and optionally, a binary
+/// search table of pointers to the `.eh_frame` records are found in this section.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct EhFrameHdr<R: Reader>(R);
+
+/// `ParsedEhFrameHdr` contains the parsed information fron the `.eh_frame` section.
+#[derive(Clone, Debug)]
+pub struct ParsedEhFrameHdr<R: Reader> {
```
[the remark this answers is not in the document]
main--: same

src/cfi.rs
main--: I think this probably should just return an error.

I mean, this isn't even a pointer, it's just a count that uses the pointer-encoding mechanism. I can't imagine any sane (or insane) reason why a compiler would emit an indirect pointer here and supporting this case just makes everything so very awkward (since we definitely can't just deref) - so why bother when this won't ever happen?

src/cfi.rs
main--: Yeah, another tricky problem. Just like with `fde_count`, the compiler *could* in theory do something completely nonsensical here for no reason at all.

Again, I decided not to bother as that's what other implementations do ([llvm](https://github.com/llvm-mirror/libunwind/blob/6a5c438e9ef71c86d1d5717519c618dcf5e8100d/src/EHHeaderParser.hpp#L150), [libunwind (lol)](https://github.com/pathscale/libunwind/blob/vanilla_pathscale/src/dwarf/Gfind_unwind_table.c#L153)).

So I would say it's de facto prohibited.

src/cfi.rs
fitzgen: > I think this probably should just return an error.

Agreed, and we should have test coverage.

src/cfi.rs
main--: I actually originally intended to use std's binary search here. I ended up doing it this way instead when I didn't find a safe way (as in: not `transmute`) to convert from `&[u8]` to `&[u32]` and friends.

src/cfi.rs
fitzgen: Makes sense. We should leave a comment to this effect here and ... *queue broken record* ... have test coverage for this case.

> libunwind (lol)

Wow...

src/cfi.rs
fitzgen: Yeah, I guess that `transmute` involves making sure that alignment is correct and all that too or else we'll perform unaligned loads and go into UB.

Ok, let's keep this binary search implementation.

philipc:
Is this ready for another round of review?

main--:
Yes

fitzgen:
(For the record, it's always a safe-bet to drop a comment when a PR is ready to be looked at again -- half the time I don't get the "more commits pushed" emails...)

PR #252, opened 2017-10-14
Make 'dw' macro only one branch

nvzqz:
Trailing commas can be handled in a single branch.

The benefit is that any changes to the first branch don't need to be reflected in a second branch.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -55,7 +55,7 @@ use std::fmt;
 //         }
 //     }
 macro_rules! dw {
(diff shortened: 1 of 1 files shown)


philipc:
This makes multiple trailing commas valid, which I would prefer to disallow.

The only changes that need to be reflected in the second branch are to the pattern itself, which is unlikely to change at this stage, so I'm not sure this is worth it.

philipc:
Closing for the reason given above. I'm happy to reconsider this if there is a change that affects both patterns.

PR #254, opened 2017-10-14
Use 'write' macro instead of 'format' in 'dw'

nvzqz:
Using `format!` allocates a new `String` that then immediately gets dropped.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -69,9 +69,9 @@ macro_rules! dw {
                     $(
                         $name => f.pad(stringify!($name)),
                     )+
-                    otherwise => f.pad(&format!("Unknown {}: {}",
-                                                stringify!($struct_name),
-                                                otherwise.0)),
+                    otherwise => write!(f, "Unknown {}: {}",
+                                        stringify!($struct_name),
+                                        otherwise.0),
                 }
             }
         }


philipc:
Hmm, this loses the pad though, which we wanted for our dwarfdump clone.

I'm actually not sure that we should be implementing `Display` at all, since not all values can be properly represented as strings, because we don't know their name. The "Unknown {}: {}" formatting should be something determined by users of gimli. Maybe we should replace this `Display` impl with an `as_str` method that returns `Option<&'static str>`. Thoughts @fitzgen?

fitzgen:
We can pad inside `dwarfdump` with `{: >10}` or whatever the exact formatting syntax is (I always have to experiment / look it up). https://doc.rust-lang.org/nightly/std/fmt/index.html#syntax

philipc:
We are padding with `{:27}` (see [here](https://github.com/gimli-rs/gimli/blob/69cae44c90c887d7ce4dfc420f393cb41e6638e3/examples/dwarfdump.rs#L677)), but that only works if the `Display` implementation supports padding (see the last sentence [here](https://doc.rust-lang.org/std/fmt/#fillalignment)), and we support padding in this implementation using `Formatter::pad` which requires the complete string.

@nvzqz What's your use case for which this allocation is a problem? The current behaviour is very convenient, so I don't really want change it. Can we solve your problem in another way?

nvzqz:
I don't have a problem with the allocation. I just like to generally avoid them when they seem unnecessary. In this case, it seems that padding is more important.

philipc:
Ok thanks, I'm going to leave this as is.

PR #256, opened 2017-10-27
Make RequiresEntryValue contain an Expression instead of a Reader

khuey:
since consumers will generally want to evaluate the contents rather than inspecting them.

Simple API cleanup


fitzgen:
From Travis CI:

> error: process didn't exit successfully: `/Users/travis/build/gimli-rs/gimli/target/debug/deps/gimli-8d609308feda9cb4` (signal: 10, SIGBUS: access to undefined memory)

I don't think that is related to these changes, but that doesn't look good...

fitzgen:
I restarted it to see if it goes away...

philipc:
The SIGBUS is probably https://github.com/rust-lang/rust/issues/45731

PR #257, opened 2017-10-28
Basic support for DWARF 5 .debug_info

khuey:
The compilation unit header layout has changed, and the new `DW_FORM_implicit_const` requires handling while traversing the abbreviations.

### DIFF
--- a/src/abbrev.rs
+++ b/src/abbrev.rs
@@ -259,15 +259,21 @@ impl Abbreviation {
 pub struct AttributeSpecification {
     name: constants::DwAt,
     form: constants::DwForm,
+    implicit_const_value: i64,
 }
 
 impl AttributeSpecification {
-    /// Construct a new `AttributeSpecification` from the given name and form.
+    /// Construct a new `AttributeSpecification` from the given name and form
+    /// and implicit const value.
     #[inline]
-    pub fn new(name: constants::DwAt, form: constants::DwForm) -> AttributeSpecification {
+    pub fn new(name: constants::DwAt, form: constants::DwForm,
+               implicit_const_value: Option<i64>) -> AttributeSpecification {
+        debug_assert!((form == constants::DW_FORM_implicit_const && implicit_const_value.is_some()) ||
+                      (form != constants::DW_FORM_implicit_const && implicit_const_value.is_none()));
         AttributeSpecification {
             name: name,
             form: form,
+            implicit_const_value: implicit_const_value.unwrap_or(0),
         }
     }
 
@@ -283,6 +289,13 @@ impl AttributeSpecification {
         self.form
     }
 
+    /// Get the attribute's implicit const value.
+    #[inline]
+    pub fn implicit_const_value(&self) -> i64 {
+        assert!(self.form == constants::DW_FORM_implicit_const);
+        self.implicit_const_value
+    }
+
     /// Return the size of the attribute, in bytes.
     ///
     /// Note that because some attributes are variably sized, the size cannot
@@ -290,6 +303,8 @@ impl AttributeSpecification {
     pub fn size<R: Reader>(&self, header: &UnitHeader<R, R::Offset>) -> Option<usize> {
         match self.form {
             constants::DW_FORM_addr => Some(header.address_size() as usize),
+
+            constants::DW_FORM_implicit_const => Some(0),
 
             constants::DW_FORM_flag |
             constants::DW_FORM_flag_present |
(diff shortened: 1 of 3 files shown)


src/abbrev.rs
```diff
@@ -157,6 +157,7 @@ pub struct Abbreviation {
     tag: constants::DwTag,
     has_children: constants::DwChildren,
     attributes: Vec<AttributeSpecification>,
+    implicit_consts: Vec<i64>,
```
philipc: I feel that we should add the implicit constant to the `AttributeSpecification` instead. It'll simply things a lot, and I don't think the increased memory usage will be significant (it may even be less memory for abbreviations with only a couple of attributes).

src/unit.rs
```diff
@@ -660,8 +666,21 @@ fn parse_unit_header<R: Reader>(input: &mut R) -> Result<UnitHeader<R, R::Offset
...
+        address_size = rest.read_u8()?;
+    } else if version == 5 {
+        let unit_type = parse_compilation_unit_type(&mut rest)?;
+        if unit_type != constants::DW_UT_compile && unit_type != constants::DW_UT_type {
+            return Err(Error::UnsupportedUnitType.into());
+        }
+        address_size = rest.read_u8()?;
+        offset = parse_debug_abbrev_offset(&mut rest, format)?;
+    } else {
+        unreachable!()
```
philipc: Delete `parse_version` and return `UnknownVersion` here instead.

src/abbrev.rs
khuey: I had assumed that an AttributeSpecification was 4 bytes but it looks like it's actually 16 (why are DwAt and DwForm u64s and not u16s?) so we'd break even after 3 attributes.  I could go either way here.

src/unit.rs
khuey: I assume you mean `unreachable!`?

philipc: No, I meant to delete `parse_version` and replace it with `read_u16`. I don't think `parse_version` provides any value: we do a check of the version in it, then we need to immediately repeat the  check when it returns, so I think the code is clearer without `parse_version`.

src/abbrev.rs
philipc: DwAt and DwForm are u64s because that's what `read_uleb128` returns, and the standard doesn't explicitly define otherwise. Probably not a good reason, and I had noticed it in the past but never got around to doing anything about it. So this is something we may change in future.

This is a 2-3% performance hit to `bench_parsing_debug_info` (although I'm comparing to before this change, not the alternative implementation). So due to that, and due to the code complexity, I prefer changing this, but not strongly enough to require it. FWIW, gdb and llvm both store it in the attribute specification.

src/unit.rs
```diff
@@ -660,8 +666,21 @@ fn parse_unit_header<R: Reader>(input: &mut R) -> Result<UnitHeader<R, R::Offset
...
-    let offset = parse_debug_abbrev_offset(&mut rest, format)?;
-    let address_size = rest.read_u8()?;
+    let offset;
+    let address_size;
+    if 2 <= version && version <= 4 {
+        offset = parse_debug_abbrev_offset(&mut rest, format)?;
+        address_size = rest.read_u8()?;
+    } else if version == 5 {
+        let unit_type = parse_compilation_unit_type(&mut rest)?;
+        if unit_type != constants::DW_UT_compile && unit_type != constants::DW_UT_type {
```
philipc: Change this to only accept `DW_UT_compile`. Currently it will not work for `DW_UT_type`, because we only parse the extra type unit fields when it is in the `.debug_types` section, but DWARF 5 type units are in the `.debug_info` section.

src/abbrev.rs
khuey: On my system this actually seems to *speed up* `bench_parsing_debug_info`, not that that makes a ton of sense.

I don't feel very strongly about this so I'll change it.

khuey:
Comments addressed.

src/unit.rs
```diff
@@ -1592,11 +1600,13 @@ fn length_uleb128_value<R: Reader>(input: &mut R) -> Result<R> {
...
-fn parse_attribute<'unit, R: Reader>(
+fn parse_attribute<'unit, 'abbrev, R: Reader>(
     input: &mut R,
     unit: &'unit UnitHeader<R, R::Offset>,
-    spec: AttributeSpecification,
-) -> Result<Attribute<R>> {
+    mut specs: &'abbrev [AttributeSpecification],
+) -> Result<(Attribute<R>, &'abbrev [AttributeSpecification])> {
+    let spec = specs[0];
+    specs = &specs[1..];
```
philipc: This change isn't needed anymore, and it's part of what I thought would affect performance, but reverting it didn't help, so going to leave any cleanup here for future performance work.

philipc:
Thanks @khuey!

PR #259, opened 2017-11-04
Update dep arrayvec to its latest version (0.4.x)

bluss:
- Update arrayvec to 0.4.x
- More consistent API and it dropped a dependency (odds), so faster to compile

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -1844,8 +1844,8 @@ where
 
     fn reset(&mut self) {
         self.stack.clear();
-        let res = self.stack.push(UnwindTableRow::default());
-        debug_assert!(res.is_none());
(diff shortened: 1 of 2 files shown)


src/cfi.rs
```diff
@@ -1912,11 +1912,8 @@ where
 
     fn push_row(&mut self) -> Result<()> {
         let new_row = self.row().clone();
-        if self.stack.push(new_row).is_some() {
-            Err(Error::CfiStackFull)
-        } else {
-            Ok(())
-        }
+        self.stack.try_push(new_row)
+                  .map_err(|_| Error::CfiStackFull)
```
bluss: I haven't rustfmt'ed this. Maybe it won't like this line

fitzgen: No worries.

PR #261, opened 2017-11-12
Upgrade object

sfackler:


fitzgen:
👍 thanks! Will merge once CI comes back green.

PR #262, opened 2017-11-13
Add GnuDebuglink

### LINKED ISSUE #260: Add a parser for the .gnu_debuglink section
As described in https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Debugging_with_gdb/separate-debug-files.html.

It's pretty small - just a string and a CRC value.

sfackler:
Closes #260

### DIFF
--- a/src/gnu_debuglink.rs
+++ b/src/gnu_debuglink.rs
@@ -1,0 +1,80 @@
+use reader::{Reader, ReaderOffset};
+use endianity::{EndianBuf, Endianity};
+use parser::Result;
+use Section;
+
+/// The `GnuDebuglink` struct represents the external debuginfo information
+/// found in the `.gnu_debuglink` section.
+#[derive(Debug, Clone, Copy)]
+pub struct GnuDebuglink<R: Reader> {
+    gnu_debuglink_section: R,
+}
+
+impl<'input, Endian> GnuDebuglink<EndianBuf<'input, Endian>>
+where
+    Endian: Endianity
+{
+    /// Constructs a new `GnuDebuglink` instance from the data in the
+    /// `.gnu_debuglink` section.
+    ///
+    /// It is the caller's responsibility to read the `.gnu_debuglink` section
+    /// and present it as a `&[u8]` slice. This means using some ELF loader on
+    /// Linux, a Mach-O loader on OSX, etc.
+    ///
+    /// ```
+    /// use gimli::{GnuDebuglink, LittleEndian};
+    ///
+    /// # let buf = [0x00, 0x01, 0x02, 0x03];
+    /// # let read_gnu_debuglink_section_somehow = || &buf;
+    /// let gnu_debuglink = GnuDebuglink::new(read_gnu_debuglink_section_somehow(), LittleEndian);
+    /// ```
+    pub fn new(gnu_debuglink_section: &'input [u8], endian: Endian) -> Self {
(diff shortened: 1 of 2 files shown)


src/gnu_debuglink.rs
```diff
@@ -0,0 +1,80 @@
...
+impl<R: Reader> GnuDebuglink<R> {
+    /// Returns the filename of the debug information file.
+    pub fn filename(&self) -> Result<R> {
+        self.gnu_debuglink_section.clone().read_null_terminated_slice()
+    }
+
+    /// Returns the CRC32 checksum of the debug information file.
+    pub fn crc32(&self) -> Result<u32> {
+        let mut section = self.gnu_debuglink_section.clone();
+        let base = section.len() - ReaderOffset::from_u32(4);
```
philipc: This works fine for valid data, but doesn't handle invalid data very well. It'd be better to have a single method that reads both the filename and crc32. I guess it doesn't matter too much though because the crc won't match if it's wrong.

philipc:
Is there a compatible implementation of the crc32 that we can use in conjunction with this?

src/gnu_debuglink.rs
sfackler: Yeah, I wasn't sure what the right API is here - is there another section type that's similar? It seems like the others currently supported are all sequences.

sfackler:
I think https://docs.rs/crc/1.5.0/crc/crc32/fn.checksum_ieee.html is the right crc variant?

src/gnu_debuglink.rs
philipc: I didn't see any other sections that are similar. I think just a `fn read(&self) -> Result<(R, u32)>` would be fine.

tromey:
I wonder if this should be in moria, not gimli.  It doesn't really seem DWARF-related to me.

OTOH I suppose it wouldn't be too bad to just move it to moria when that has some content.

sfackler:
Yeah it might make sense for this to live in Moria. We'll also need to pull `.note.gnu.build-id`, which is even less related to DWARF stuff.

tromey:
> We'll also need to pull .note.gnu.build-id, which is even less related to DWARF stuff.

I checked and goblin can already extract this: https://github.com/m4b/goblin/blob/master/src/elf/note.rs

luser:
I'm going to submit a PR to make `moria` an actual crate soon: https://github.com/gimli-rs/moria/issues/1#issuecomment-352134551 so you could definitely put your work into it then!

luser:
My moria PR is up: https://github.com/gimli-rs/moria/pull/2

I wound up adding APIs to `object` to make this easier: https://github.com/gimli-rs/object/pull/39

You'll probably want to implement `Object::debug_file_info` for ELF: https://github.com/gimli-rs/object/pull/39/commits/ae5b66a5ee3e4abfa40f4a5f7e0e813f8d12c926#diff-11f7414d9fed445f4231a47e0491cadaR157 and add some other variants to the `DebugFileInfo` enum to support Build ID etc: https://github.com/gimli-rs/object/pull/39/commits/ae5b66a5ee3e4abfa40f4a5f7e0e813f8d12c926#diff-b4aea3e418ccdb71239b96952d9cddb6R62

philipc:
Implemented in the object crate instead (and used by moria).

PR #268, opened 2017-12-14
Add comment documenting type inference in dump_file in dwarfdump.rs

### LINKED ISSUE #267: Is the dump_file function in the dwarfdump example too magical?
The dump_file function makes heavy use of type inference for its `load_section` subfunction, inferring the types of `debug_abbrev`, `debug_ranges`, etc. from their use in the function calls below. While this is a neat demonstration of Rust's type inference abilities, it took me (admittedly a novice Rustacean) a great deal of time to understand how the right sections were getting loaded into the appropriate variables.

Given that `dwarfdump.rs` is intended as an example, is this level of "magic" appropriate?

iamthad:
Fixes #267

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -222,6 +222,8 @@ where
         S::from(gimli::EndianBuf::new(data, endian))
     }
 
+    // Variables representing sections of the file. The type of each is inferred from its use in the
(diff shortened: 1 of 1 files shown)


fitzgen:
OSX CI seems busted at the moment, so I'm just going to merge this in.

PR #271, opened 2017-12-25
Add #![no_std] support

### LINKED ISSUE #138: no_std support
It would be nice for gimli to have `no_std` support so it could be used inside kernels.

whitequark:
Fixes #138.

This was surprisingly easy. There are essentially no meaningful changes required.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,9 +160,43 @@
 // False positives when block expressions are used inside an assertion.
 #![allow(panic_params)]
 
+#![no_std]
+#![cfg_attr(feature = "nightly", feature(alloc))]
+
+#[cfg(feature = "std")]
+#[macro_use]
+extern crate std;
+
+#[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate core as std;
+#[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate alloc;
+
 extern crate arrayvec;
 extern crate byteorder;
 extern crate fallible_iterator;
+
+#[cfg(feature = "std")]
+mod imports {
+    pub use std::boxed;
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -160,10 +160,44 @@
 // False positives when block expressions are used inside an assertion.
 #![allow(panic_params)]
 
+#![no_std]
+#![cfg_attr(feature = "nightly", feature(alloc))]
```
philipc: Is there a reason for this to use a `nightly` feature, instead of `not(feature = "std")` like we do elsewhere? It'd be nice if `cargo build --no-default-features` could work.

whitequark: Yes. See Cargo.toml:
```
nightly = ["fallible-iterator/alloc"]
```

philipc: Ok, this is probably nitpicky, but I just want to make sure I understand if you don't mind:
- do we need to use `fallible-iterator/alloc` in gimli?
- why don't the uses of the alloc crate depend on the `nightly` feature?

whitequark: > do we need to use fallible-iterator/alloc in gimli?

Yes. Otherwise, it cannot be tested without the `std` feature, I believe.

fitzgen:
Thanks for adding `no_std` support @whitequark ! Was the btreemap vs hashmap related to `no_std` or was it for performance? And if the latter, what kind of difference did you observe?

whitequark:
@fitzgen Just `#![no_std]`. There is no hash-anything without std, because they implicitly rely on the global state (the default RNG).

fitzgen:
Ah, yeah that makes sense. Thanks again!

Issue #273, opened 2017-12-29
DW_CFA_restore_state overwrites location


xmartinez:
Evaluating `DW_CFA_restore_state` should pop the set of rules while
preserving the current location.  Currently, the previous location is
also (incorrectly) restored, resulting in the following situation:

```
Path: /usr/lib/firefox/libxul.so
Build ID: 1e6a2b8b84fe61d484e96d82aab4ad132c0d08cd
FDE (offset: 007f44a0  vaddr: 038cce10-038cce54)
Instructions:
  DW_CFA_advance_loc: 1 to 038cce11
  DW_CFA_def_cfa_offset: 16
  DW_CFA_offset: r3 at cfa-16
  DW_CFA_advance_loc: 41 to 038cce3a
  DW_CFA_remember_state
  DW_CFA_def_cfa_offset: 8
  DW_CFA_advance_loc: 6 to 038cce40
  DW_CFA_restore_state
  DW_CFA_advance_loc: 19 to 038cce53
  DW_CFA_def_cfa_offset: 8
  DW_CFA_nop

Evaluated unwind table:
  038cce10-038cce11  cfa: r7+8  r16: c-8
  038cce11-038cce3a  cfa: r7+16  r16: c-8  r3: c-16
  038cce3a-038cce40  cfa: r7+8  r16: c-8  r3: c-16
  038cce3a-038cce4d  cfa: r7+16  r16: c-8  r3: c-16    <-- (1)
  038cce4d-038cce54  cfa: r7+8  r16: c-8  r3: c-16
```

Row (1) has an incorrect address range (which also affects following
rows), as its start address is restored when popping the register
rules.

The correct unwind table should be:
```
  038cce10-038cce11  cfa: r7+8  r16: c-8
  038cce11-038cce3a  cfa: r7+16  r16: c-8  r3: c-16
  038cce3a-038cce40  cfa: r7+8  r16: c-8  r3: c-16
  038cce40-038cce53  cfa: r7+16  r16: c-8  r3: c-16
  038cce53-038cce54  cfa: r7+8  r16: c-8  r3: c-16
```

philipc:
Fixed by #274

PR #274, opened 2017-12-29
Preserve location when evaluating DW_CFA_restore_state

xmartinez:
DW_CFA_restore_state pops the set of rules off the implicit stack, but
should preserve the current location.


xmartinez:
Closes #273

PR #275, opened 2017-12-29
Some docs and CI things

fitzgen:
Also renames "nightly" feature to "alloc". cc @whitequark 

I messed with the travis ci config a bit, so I might need to do a couple force pushes before tests are passing.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -ex
+
+case "$GIMLI_JOB" in
+    "test")
+        cargo clean
+        cargo build $GIMLI_PROFILE
+        cargo test $GIMLI_PROFILE
+        ;;
+
+    "examples")
+        cargo build $GIMLI_PROFILE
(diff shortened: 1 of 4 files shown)


whitequark:
:+1:

ci/script.sh
```diff
@@ -0,0 +1,51 @@
...
+        cargo build --no-default-features --features alloc $GIMLI_PROFILE
+        ;;
+
+    "bench")
+        cargo bench
+        ;;
+
+    "coverage")
+        pip2 install -v 'travis-cargo<0.2' --user
+        TRAVIS_CARGO_NIGHTLY_FEATURE= travis-cargo coveralls --no-sudo --verify
```
philipc: Looks like this line is causing the build error. We should ignore any error on this line because we're only using it to install kcov. This wasn't a problem before because we were doing coverage in the after_success commands.

Or we could switch to tarpaulin.

fitzgen:
Decrease of coverage is due to switch to tarpaulin.

PR #276, opened 2017-12-30
Optimise line number iteration

philipc:
Useful for addr2line. Adds another `inline(always)` unfortunately.

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -195,12 +195,16 @@ where
         let maximum_operations_per_instruction =
             self.header().maximum_operations_per_instruction as u64;
 
-        let op_index_with_advance = self.row.registers.op_index + operation_advance;
-
-        self.row.registers.address += minimum_instruction_length *
-            (op_index_with_advance / maximum_operations_per_instruction);
-
-        self.row.registers.op_index = op_index_with_advance % maximum_operations_per_instruction;
+        if maximum_operations_per_instruction == 1 {
+            self.row.registers.address += minimum_instruction_length * operation_advance;
+            self.row.registers.op_index = 0;
+        } else {
+            let op_index_with_advance = self.row.registers.op_index + operation_advance;
+            self.row.registers.address += minimum_instruction_length
+                * (op_index_with_advance / maximum_operations_per_instruction);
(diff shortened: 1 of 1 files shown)


src/line.rs
```diff
@@ -195,12 +195,16 @@ where
         let maximum_operations_per_instruction =
             self.header().maximum_operations_per_instruction as u64;
 
-        let op_index_with_advance = self.row.registers.op_index + operation_advance;
-
-        self.row.registers.address += minimum_instruction_length *
-            (op_index_with_advance / maximum_operations_per_instruction);
-
-        self.row.registers.op_index = op_index_with_advance % maximum_operations_per_instruction;
+        if maximum_operations_per_instruction == 1 {
```
fitzgen: Wow, a little surprised this is such a large speed up.

src/line.rs
```diff
@@ -503,7 +503,7 @@ pub enum Opcode<R: Reader> {
     UnknownStandard1(constants::DwLns, u64),
 
     /// An unknown standard opcode with multiple operands.
-    UnknownStandardN(constants::DwLns, Vec<u64>),
+    UnknownStandardN(constants::DwLns, R),
```
fitzgen: To be clear, the issue is the size of `Vec` bloating the decoded instruction size, *not* that we are hitting a lot of unknown standard opcodes that also have more than one operand, right?

fitzgen: If so, we could save another word (assuming this is still the widest variant) by boxing `R`, since `R` is usually two words with `EndianBuf`.

philipc: No, it's not the size of the `Vec`. The largest variant is currently `DefineFile` at 48 bytes, and `UnknownStandardN` was already only 32 before this change.

We're not hitting unknown standard opcodes.

I haven't investigated the assembly in great depth (there's too much), but there's a few inlined `Vec` functions that are no longer needed, and I guess this helps the optimiser. It also means there's no code to run to drop the `Opcode`.

src/line.rs
philipc: Agreed. It definitely helped in addr2line too though, so it's not just a synthetic benchmark.

PR #280, opened 2018-01-27
Add LocationLists and RangeLists

philipc:
These allow users to parse lists without caring about DWARF version details.

The old debug_loc/debug_ranges interfaces are removed, since LocationsLists/RangeLists parse these now, and it was too easy to wrongly use them for DWARF version 5.

cc @rocallahan


src/loclists.rs
```diff
@@ -39,39 +39,83 @@ where
     }
 }
 
+/// An offset into either the `.debug_loc` section or the `.debug_loclists` section,
+/// depending on the version of the unit the offset was contained in.
```
fitzgen: Should we return an `enum` of either a `DebugLocOffset` or a `DebugLocListsOffset`? I'm concerned that we are making it easier to misuse our APIs and weakening the benefits that strongly typing offsets gives.

fitzgen: Nevermind -- I understand how this works better now.

fitzgen:
Thanks @philipc !

src/loclists.rs
philipc: Yeah I wasn't sure which way was better.  We could make it an enum, and then you wouldn't need to pass the unit version into the other functions, but you would still need to pass the unit address size (or we could include that in the enum too).

PR #291, opened 2018-04-15
Run rustfmt

philipc:
This is the rustfmt version bundled with rust 1.25.

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -138,8 +138,10 @@ impl<R: Reader> EhFrameHdr<R> {
             return Err(Error::UnexpectedNull);
         }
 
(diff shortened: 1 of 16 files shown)


src/cfi.rs
```diff
@@ -3893,9 +3904,11 @@ mod tests {
...
         };
 
-        let section = section
-            .fde(Endian::Big, &cie1_location, &mut fde1)
-            .fde(Endian::Big, &cie2_location, &mut fde2);
+        let section = section.fde(Endian::Big, &cie1_location, &mut fde1).fde(
+            Endian::Big,
+            &cie2_location,
+            &mut fde2,
+        );
```
philipc: I've filed an issue about this change.

PR #292, opened 2018-04-17
Operation simplifications

philipc:
Just a few code cleanups I did while making sure I understood the code. The parsing change increased performance, but otherwise there shouldn't be anything noticeable to users.

### DIFF
--- a/src/op.rs
+++ b/src/op.rs
@@ -222,41 +222,12 @@ where
     },
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 enum OperationEvaluationResult<R: Reader> {
     Piece,
     Incomplete,
-    Complete {
-        location: Location<R, R::Offset>,
-    },
-    AwaitingMemory {
-        address: u64,
-        size: u8,
-        space: Option<u64>,
-    },
-    AwaitingRegister {
-        register: u64,
-        offset: u64,
-    },
-    AwaitingFrameBase {
-        offset: u64,
-    },
-    AwaitingTls {
-        index: u64,
-    },
-    AwaitingCfa,
-    AwaitingAtLocation {
-        location: DieReference<R::Offset>,
-    },
-    AwaitingEntryValue {
-        expression: R,
-    },
-    AwaitingParameterRef {
-        parameter: UnitOffset<R::Offset>,
-    },
-    AwaitingTextBase {
-        offset: u64,
-    },
+    Complete { location: Location<R, R::Offset> },
+    Waiting(EvaluationWaiting, EvaluationResult<R>),
 }
 
 /// A single location of a piece of the result of a DWARF expression.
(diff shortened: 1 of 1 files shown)


src/op.rs
```diff
@@ -1,7 +1,7 @@
 //! Functions for parsing and evaluating DWARF expressions.
 
 use constants;
-use parser::{Error, Format};
+use parser::{Error, Format, Result};
```
fitzgen: Nice :+1:

src/op.rs
```diff
@@ -491,291 +491,109 @@ where
...
-            }
-            constants::DW_OP_breg30 => {
-                let value = bytes.read_sleb128()?;
-                Ok(Operation::RegisterOffset {
-                    register: 30,
-                    offset: value,
-                })
-            }
-            constants::DW_OP_breg31 => {
+            constants::DW_OP_lit0
```
fitzgen: Also nice :+1:

src/op.rs
```diff
@@ -1244,28 +1240,28 @@ impl<R: Reader> Evaluation<R> {
                 if base_type != UnitOffset(R::Offset::from_u64(0).unwrap()) {
                     return Err(Error::UnsupportedTypedStack);
                 }
-                terminated = true;
-                current_location = Location::Register { register: register };
+                let location = Location::Register { register: register };
```
fitzgen: Nitpick: this can use object short hand:

```rust
                let location = Location::Register { register }; 
```

src/op.rs
```diff
@@ -740,12 +711,25 @@ where
...
+enum EvaluationWaiting {
+    Memory,
+    Register { offset: i64 },
+    FrameBase { offset: i64 },
+    Tls,
+    Cfa,
+    AtLocation,
+    EntryValue,
+    ParameterRef,
+    TextBase { offset: u64 },
```
fitzgen: Since @khuey introduced the waiting stuff so that he could integrate well with tokio/futures APIs, it seems to me like it would be good to get his take on this commit's changes.

Looks good to me, FWIW.

khuey: As far as I can tell this doesn't change the public API at all so it's fine for me. I don't really see why this is a win *internally* though. If I understand what this does correctly it violates the "make impossible states unrepresentable" principle by allowing an `OperationEvaluationResult` to contain mismatched `EvaluationWaiting` and `EvaluationResult` enum values.

philipc: I don't understand how this is worse than before in terms of "make impossible states unrepresentable". There was nothing before that disallowed a mismatch between the state and the result either (ie the value stored in `EvaluationState::state` and the value returned from `Evaluation::evaluate`).

It's a win IMO because the logic for handling an operation is all in one spot, instead of split between `evaluate_one_operation` and `evaluate_internal`.

khuey: Mmm I guess that's fair.

Carry on.

PR #293, opened 2018-04-25
Implement operations on typed values

philipc:
A couple of things of note:
- parsing the DIEs and maintaining a hash of base types is the user's responsibility, but we may want to add more to help with this.
- the evaluation requests base types from the user via `EvaluationResult::RequiresBaseType`. Even though this probably isn't something that will need a Future, I chose this because it avoids adding a type parameter to `Evaluation` for a closure or trait.

I used @tromey's work as a starting point, but the implementation has significantly changed. I'm now using an enum for the value, instead of a struct with type and value members. The reason for this is I think it makes it easier to get the operations correct in some cases. For example, I think that division for unsigned base types should be an unsigned division (I think the requirement for signed division in the standard only applies to the generic type). One drawback of this approach is that the reinterpret operation is ugly, and I'm open to ideas on how to simplify that. @tromey Can you please take a quick look at this approach and let me know if you see any design problems with it?

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -1,0 +1,1615 @@
+//! Definitions for values used in DWARF expressions.
+
+use constants;
+use parser::{Error, Result};
+use reader::Reader;
+use std::mem;
+use unit::{AttributeValue, DebuggingInformationEntry};
+
+/// Convert a u64 to an i64, with sign extension if required.
+///
+/// This is primarily used when needing to treat `Value::Generic`
+/// as a signed value.
+#[inline]
+fn sign_extend(value: u64, mask: u64) -> i64 {
+    let value = (value & mask) as i64;
+    let sign = ((mask >> 1) + 1) as i64;
+    (value ^ sign).wrapping_sub(sign)
+}
+
+#[inline]
+fn mask_bit_size(addr_mask: u64) -> u32 {
+    64 - addr_mask.leading_zeros()
+}
+
+/// The type of an entry on the DWARF stack.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ValueType {
+    /// The generic type, which is address-sized and of unspecified sign,
+    /// as specified in the DWARF 5 standard, section 2.5.1.
+    /// This type is also used to represent address base types.
+    Generic,
+    /// Signed 8-bit integer type.
+    I8,
+    /// Unsigned 8-bit integer type.
+    U8,
+    /// Signed 16-bit integer type.
+    I16,
+    /// Unsigned 16-bit integer type.
+    U16,
+    /// Signed 32-bit integer type.
+    I32,
+    /// Unsigned 32-bit integer type.
+    U32,
+    /// Signed 64-bit integer type.
+    I64,
+    /// Unsigned 64-bit integer type.
+    U64,
+    /// 32-bit floating point type.
+    F32,
+    /// 64-bit floating point type.
+    F64,
+}
+
+/// The value of an entry on the DWARF stack.
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum Value {
+    /// A generic value, which is address-sized and of unspecified sign.
+    Generic(u64),
+    /// A signed 8-bit integer value.
+    I8(i8),
+    /// An unsigned 8-bit integer value.
+    U8(u8),
+    /// A signed 16-bit integer value.
+    I16(i16),
+    /// An unsigned 16-bit integer value.
+    U16(u16),
+    /// A signed 32-bit integer value.
+    I32(i32),
+    /// An unsigned 32-bit integer value.
+    U32(u32),
+    /// A signed 64-bit integer value.
+    I64(i64),
+    /// An unsigned 64-bit integer value.
+    U64(u64),
+    /// A 32-bit floating point value.
+    F32(f32),
+    /// A 64-bit floating point value.
+    F64(f64),
+}
+
+impl ValueType {
+    /// The size in bits of a value for this type.
+    pub fn bit_size(self, addr_mask: u64) -> u32 {
+        match self {
+            ValueType::Generic => mask_bit_size(addr_mask),
+            ValueType::I8 | ValueType::U8 => 8,
+            ValueType::I16 | ValueType::U16 => 16,
+            ValueType::I32 | ValueType::U32 | ValueType::F32 => 32,
+            ValueType::I64 | ValueType::U64 | ValueType::F64 => 64,
+        }
+    }
+
+    /// Construct a `ValueType` from the attributes of a base type DIE.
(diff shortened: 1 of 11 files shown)


tromey:
This seems reasonable to me.

philipc:
This is ready for review now.

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+fn mask_bit_size(addr_mask: u64) -> u32 {
+    64 - addr_mask.leading_zeros()
+}
+
+/// The type of an entry on the DWARF stack.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ValueType {
+    /// The generic type, which is address-sized and of unspecified sign.
+    /// This type is also used to represent address base types.
+    Generic,
```
fitzgen: Maybe point to 2.5.1. in dwarf5 here as well?

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+            _ => return Err(Error::UnsupportedTypeOperation),
+        };
+        Ok(value)
+    }
+
+    /// Convert a `Value` to a `u64`.
+    ///
+    /// The `ValueType` of `self` must be integral.
+    /// Values are sign extended if the source value is signed.
+    pub fn to_u64(self, addr_mask: u64) -> Result<u64> {
```
fitzgen: We should file a follow up issue to implement `TryFrom<Value> for u64` in a month and a half, when it hits stable.

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+        };
+        Ok(value)
+    }
+
+    /// Create a `Value` with the given `value_type` from a `u64` value.
+    ///
+    /// The `value_type` may be integral or floating point.
+    /// The result is truncated if the `u64` value does
+    /// not fit the bounds of the `value_type`.
+    pub fn from_u64(value_type: ValueType, value: u64) -> Result<Value> {
```
fitzgen: And also `TryFrom<u64> for Value`

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+        };
+        Ok(value)
+    }
+
+    /// Create a `Value` with the given `value_type` from a `f64` value.
+    ///
+    /// The `value_type` may be integral or floating point.
+    /// The result is not defined if the `f64` value does
+    /// not fit the bounds of the `value_type`.
+    fn from_f64(value_type: ValueType, value: f64) -> Result<Value> {
```
fitzgen: And I guess all of these variants.

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+                Value::Generic(v1.wrapping_add(v2) & addr_mask)
+            }
+            (Value::I8(v1), Value::I8(v2)) => Value::I8(v1.wrapping_add(v2)),
+            (Value::U8(v1), Value::U8(v2)) => Value::U8(v1.wrapping_add(v2)),
+            (Value::I16(v1), Value::I16(v2)) => Value::I16(v1.wrapping_add(v2)),
+            (Value::U16(v1), Value::U16(v2)) => Value::U16(v1.wrapping_add(v2)),
+            (Value::I32(v1), Value::I32(v2)) => Value::I32(v1.wrapping_add(v2)),
+            (Value::U32(v1), Value::U32(v2)) => Value::U32(v1.wrapping_add(v2)),
+            (Value::I64(v1), Value::I64(v2)) => Value::I64(v1.wrapping_add(v2)),
+            (Value::U64(v1), Value::U64(v2)) => Value::U64(v1.wrapping_add(v2)),
```
fitzgen: I feel like all these operation implementations could use a macro or something to reduce boilerplate.

src/value.rs
```diff
@@ -0,0 +1,1613 @@
...
+            (Value::I8(v1), Value::I8(v2)) => v1 >= v2,
+            (Value::U8(v1), Value::U8(v2)) => v1 >= v2,
+            (Value::I16(v1), Value::I16(v2)) => v1 >= v2,
+            (Value::U16(v1), Value::U16(v2)) => v1 >= v2,
+            (Value::I32(v1), Value::I32(v2)) => v1 >= v2,
+            (Value::U32(v1), Value::U32(v2)) => v1 >= v2,
+            (Value::I64(v1), Value::I64(v2)) => v1 >= v2,
+            (Value::U64(v1), Value::U64(v2)) => v1 >= v2,
+            (Value::F32(v1), Value::F32(v2)) => v1 >= v2,
+            (Value::F64(v1), Value::F64(v2)) => v1 >= v2,
```
fitzgen: Thinking about it a little more, something like this could be useful:

```rust
enum SameTypeValues {
    Generic(u64, u64),
    I8(i8, i8),
    U8(u8, u8),
    I16(i16, i16),
    U16(u16, u16),
    I32(i32, i32),
    U32(u32, u32),
    I64(i64, i64),
    U64(u64, u64),
    F32(f32, f32),
    F64(f64, f64),
}

impl SameTypeValues {
    fn try_from(a: Value, b: Value) -> Result<SameTypeValues> {
        match (a, b) {
            (Generic(x), Generic(y)) => Ok(SameTypeValues::Generic(x, y)),
            (I8(x), I8(y)) => Ok(SameTypeValues::I8(x, y)),
            (U8(x), U8(y)) => Ok(SameTypeValues::U8(x, y)),
            (I16(x), I16(y)) => Ok(SameTypeValues::I16(x, y)),
            (U16(x), U16(y)) => Ok(SameTypeValues::U16(x, y)),
            (I32(x), I32(y)) => Ok(SameTypeValues::I32(x, y)),
            (U32(x), U32(y)) => Ok(SameTypeValues::U32(x, y)),
            (I64(x), I64(y)) => Ok(SameTypeValues::I64(x, y)),
            (U64(x), U64(y)) => Ok(SameTypeValues::U64(x, y)),
            (F32(x), F32(y)) => Ok(SameTypeValues::F32(x, y)),
            (F64(x), F64(y)) => Ok(SameTypeValues::F64(x, y)),
            _ => Err(Error::UnsupportedTypeOperation),
        }
    }
}
```

src/value.rs
philipc: I don't think that allows the `addr_mask` parameter.

src/value.rs
philipc: And here we also need the `value_type` parameter. I guess you could implement `TryFrom<(ValueType, u64)>`, but I think the function call is clearer.

src/value.rs
philipc: The users of `SameTypeValues` would still have to match on just as many cases though?

src/value.rs
philipc: What do you think of this: https://github.com/philipc/gimli/commit/d0b80cf09c4bd2ddd533ca84e31f3d668d5ef69d?

fitzgen: Looks nice!

Issue #294, opened 2018-04-28
Add an Arc-/Rc-based reader


fitzgen:
https://www.reddit.com/r/rust/comments/8fcm5i/a_sampling_cpu_profiler_for_linux_written_in_rust/dy2mm2a/?context=3

> 3) The EndianBuf does take a lifetime, which is somewhat annoying when you want to just mmap a file and you don't want to permanently leak it. What I did here (for now) is I've put everything into one structure (FrameDescriptions in frame_descriptions.rs), put the mmaped data in an Arc, wrapped everything in mem::ManuallyDrop, set the lifetimes to 'static and used mem::transmute quite liberally.

I think we don't want to just make `EndianBuf` but with replacing the `&[u8]` with `Arc<[u8]>`, because that means that slicing or splitting the reader would require copying the data. We want something like `(start, end, Arc<[u8]>)` and have slicing/splitting just change the start and end indices and bump the ref count.

cc @koute

philipc:
I'm going to look into using https://crates.io/crates/bytes (or at least use its ideas).

fitzgen:
https://carllerche.github.io/bytes/src/bytes/bytes.rs.html#155-295 seems relevant

fitzgen:
It seems that it is similar to the `(start, end, Arc<[u8]>)` representation but with variants for static buffers (I don't think this is too relevant for our use case) and for inlining small buffers directly into the handle instead of putting them behind a pointer.

koute:
If I'll mmap a file I'll have some `Arc<T>` instead of an `Arc<[u8]>`, where `T` is going to be an arbitrary type which will control the lifetime of the mmap. So it'd be the best to allow the user to use a custom type and constraint `T` to be `T: Deref<[u8]>` or `T: bytes::Buf` or something along these lines.

fitzgen:
SGTM!

fitzgen:
Some initial ground work being laid in https://github.com/gimli-rs/gimli/pull/295

PR #295, opened 2018-04-28
Rename EndianBuf to EndianSlice

fitzgen:
Makes it more clear that it isn't an owning pointer, and clears the way for adding an owned arc/rc/whatever version in #294


src/endian_slice.rs
```diff
@@ -0,0 +1,362 @@
+//! Types for compile-time and run-time endianity.
```
philipc: This needs updating.

src/endian_slice.rs
```diff
@@ -0,0 +1,362 @@
...
+use parser::{Error, Result};
+use reader::Reader;
+
+/// A `&[u8]` slice with endianity metadata.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct EndianSlice<'input, Endian>
+where
+    Endian: Endianity,
+{
+    buf: &'input [u8],
```
philipc: Worth renaming this too?

fitzgen:
> I agree it's a better name, but not completely convinced its worth the churn for users of the library (they'll probably have their variables called buf still).

Did you see that there is a type alias `type EndianBuf = EndianSlice;` to ease upgrades? It is marked deprecated, with a note saying that the type has been renamed. Should hopefully mean that upgrading isn't actually very hard.

fitzgen:
Which `rustfmt` have you been using? I just ran the latest nightly's `rustfmt-preview` component on `rustup` and it is introducing a ton of changes outside of the files I touched here.

philipc:
The stable channel rustfmt-preview from rustc 1.25.

philipc:
Don't know what's going on then. I'll run it when I rebase my PR.
```
$ rustc +stable --version
rustc 1.25.0 (84203cac6 2018-03-25)
$ cargo +stable fmt --version
0.3.8-nightly (346238f 2018-02-04)
```

PR #300, opened 2018-05-19
Fix EndianReader example

philipc:
`Arc<MmapFile>` does not implement `Deref<Target=[u8]>`

### DIFF
--- a/src/endian_reader.rs
+++ b/src/endian_reader.rs
@@ -92,9 +92,21 @@ pub type EndianArcSlice<Endian> = EndianReader<Endian, Arc<[u8]>>;
 ///     }
 /// }
 ///
+/// /// A type that represents a shared `mmap`ed file.
+/// #[derive(Debug, Clone)]
+/// pub struct ArcMmapFile(Arc<MmapFile>);
+///
+/// // And `ArcMmapFile` can deref to a slice of the `mmap`ed region of memory.
+/// impl Deref for ArcMmapFile {
+///     type Target = [u8];
+///     fn deref(&self) -> &[u8] {
+///         &*self
+///     }
+/// }
+///
 /// /// A `gimli::Reader` that is backed by an `mmap`ed file!
-/// pub type MmapFileReader<Endian> = gimli::EndianReader<Endian, Arc<MmapFile>>;
-/// # fn main() {}
+/// pub type MmapFileReader<Endian> = gimli::EndianReader<Endian, ArcMmapFile>;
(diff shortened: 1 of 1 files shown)


src/endian_reader.rs
```diff
@@ -62,7 +62,7 @@ pub type EndianArcSlice<Endian> = EndianReader<Endian, Arc<[u8]>>;
 /// use std::sync::Arc;
 ///
 /// /// A type that represents an `mmap`ed file.
-/// #[derive(Debug)]
+/// #[derive(Debug, Clone)]
```
fitzgen: This should *not* be `Clone` since it represents the mapped region directly and calls `unmap` on drop.

src/endian_reader.rs
```diff
@@ -92,9 +92,21 @@ pub type EndianArcSlice<Endian> = EndianReader<Endian, Arc<[u8]>>;
 ///     }
 /// }
 ///
+/// /// A type that represents a shared `mmap`ed file.
+/// #[derive(Debug, Clone)]
+/// pub struct ArcMmapFile(Arc<MmapFile>);
```
fitzgen: Right, and this *is* the thing that should be clone, since it is just bumping a refcount.

src/endian_reader.rs
```diff
@@ -92,9 +92,21 @@ pub type EndianArcSlice<Endian> = EndianReader<Endian, Arc<[u8]>>;
...
+///     fn deref(&self) -> &[u8] {
+///         &*self
+///     }
+/// }
+///
 /// /// A `gimli::Reader` that is backed by an `mmap`ed file!
-/// pub type MmapFileReader<Endian> = gimli::EndianReader<Endian, Arc<MmapFile>>;
-/// # fn main() {}
+/// pub type MmapFileReader<Endian> = gimli::EndianReader<Endian, ArcMmapFile>;
+/// # fn test(_: &MmapFileReader<gimli::NativeEndian>) { }
```
fitzgen: Is this change just so that `MmapFileReader` is used?

src/endian_reader.rs
philipc: Good catch, no idea why I changed that.

src/endian_reader.rs
philipc: The type checker doesn't check type alias definitions, only their uses. So the lack of `Deref` wasn't apparent until it was used.

fitzgen: Ah, yes that makes a lot of sense. Thanks!

PR #302, opened 2018-05-22
Optimize EndianReader to use ptr/len

philipc:
This requires `stable_deref_trait::CloneStableDeref` for the bytes. If needed in future, we could relax that to `stable_deref_trait::StableDeref` by implementing a `Clone` for `SubRange` that recalculates the `ptr`.

This doesn't show any significant change in the gimli benchmarks, but I have modified `addr2line` to use `Rc<Cow<[u8]>>`, and it shows about 15% improvement for all operations in that (both context creation and lookup).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ coveralls = { repository = "gimli-rs/gimli" }
 arrayvec = { version = "0.4.6", default-features = false }
 byteorder = { version = "1.0", default-features = false }
 fallible-iterator = { version = "0.1.4", default-features = false }
+stable_deref_trait = "1.0.0"
 
 [dev-dependencies]
 crossbeam = "0.3.2"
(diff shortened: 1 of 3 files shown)


fitzgen:
> This doesn't show any significant change in the gimli benchmarks, but I have modified addr2line to use Rc<Cow<[u8]>>, and it shows about 15% improvement for all operations in that (both context creation and lookup).

Interesting that our benches didn't show any speed up but the larger program did. I wonder if our benches are small enough that the caches are hiding some of the indirection costs for us here.

philipc:
perf reports that a large chunk of the time in the benches is spent in a move and drop_in_place, and this didn't change due to this PR. This PR did make the code smaller and remove some slice bounds checks, but they weren't being reported as hot. That move and drop aren't present in the `EndianSlice` benchmark. I didn't try to see where the improvement comes from in addr2line, because its benchmarks aren't suitable for running under perf.

PR #306, opened 2018-06-07
Disable default features of stable_deref_trait

est31:
std is listed in the default features of stable_deref_trait

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ coveralls = { repository = "gimli-rs/gimli" }
 arrayvec = { version = "0.4.6", default-features = false }
 byteorder = { version = "1.0", default-features = false }
 fallible-iterator = { version = "0.1.4", default-features = false }
-stable_deref_trait = "1.0.0"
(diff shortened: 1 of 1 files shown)


est31:
Given that @philipc said that we'd need that PR I've thought this PR wouldn't run successfully on CI but it did, so I'd say this PR is ready to merge. We certainly have to turn off the std feature of `stable_deref_trait` if our std feature is turned off.

Given that (against my expectations) CI ran successfully for my PR, I'd say that only `#![no_std]` users of gimli like addr2line need the PR that @philipc linked to. If that PR is merged and a new version of `stable_deref_trait` is released e.g. `1.0.1` it seems to me that we don't really need to change anything on the gimli side beyond this current PR.

fitzgen:
Probably worth waiting for that PR to merge and publish first, so we can require that newly published point release to guarantee that `no_std` builds get the correct `stable_deref_trait` then.

est31:
@fitzgen sure, let's wait

PR #307, opened 2018-06-22
Edits to next_dfs() docstring example.

data-pup:
Edited the code example in the `next_dfs()` docstring, following the same loop style as shown in dwarfdump example found here: https://github.com/gimli-rs/gimli/blob/master/examples/dwarfdump.rs#L799


data-pup:
No problem! :D

fitzgen:
Failing CI is due to https://github.com/rust-lang/rust/issues/51699 -- going to just go ahead and merge this.

philipc:
This is an improvement over the previous example, which was a non-typical use. But I think this is still a non-typical example because if you're traversing starting at the root then it's good enough to stop when `next_dfs` returns `None`. Keeping track of the depth is more useful when you start at a specific entry.

PR #316, opened 2018-07-31
[WIP] Add EhFrameHdr::lookup_and_parse function

roblabla:
Currently, `EhFrameHdr::lookup` returns a Pointer, which is an absolute pointer which may (or may not) point to the correct FDE. This API is used by unwind-rs to speed up stack unwinding. Unfortunately, unwind-rs requires patches to gimli (see gimli-rs/unwind-rs#6) in order to make full use of this API. This is because there is no method to parse the returned Pointer into a FrameDescriptionEntry. Unwind makes parse_cfi_entry, and uses it as such:

```rust
let fdeptr = eh_frame_hdr.table().unwrap().lookup(address, bases).unwrap();
let fdeptr = match fdeptr {
    Pointer::Direct(x) => x,
    _ => unreachable!(),
};

let entry = gimli::parse_cfi_entry(bases, *sel, &mut EndianBuf::new(unsafe { std::slice::from_raw_parts(fdeptr as *const u8, 0x1000000) }, NativeEndian)).unwrap().unwrap();
let target_fde = match entry {
    CieOrFde::Fde(fde) => Some(fde.parse(|offset| sel.cie_from_offset(bases, offset))?),
    CieOrFde::Cie(_) => unimplemented!(), // return error here probably
};
if let Some(fde) = target_fde {
    if !fde.contains(address) {
        return Err(gimli::Error::NoUnwindInfoForAddress);
    }
}
```

As can be seen here, there is a big fat unsafe in the middle, which sucks. I believe there is a better solution. 

The complicated part is that the fdeptr is absolute, and has no length associated to it. It is, however, guaranteed to always point to an eh_frame section. Furthermore, the eh_frame_hdr has a pointer to the start of this section. As such, I propose this solution, which allows us to get rid of the unsafety.

I additionally took the liberty of adding the guarantee 

I am in the process of testing this, and the implementation and doc might need a bit of refinement.

CC @main--

### DIFF
--- a/src/cfi.rs
+++ b/src/cfi.rs
@@ -269,6 +269,37 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
             &mut reader,
         )
     }
+
+    /// Returns a parsed FDE for the given address, or NoUnwindInfoForAddress
+    /// if there are none.
+    pub fn lookup_and_parse(&self, address: u64, bases: &BaseAddresses, frame: EhFrame<R>) -> Result<FrameDescriptionEntry<EhFrame<R>, R, R::Offset>> {
+        let fdeptr = self.lookup(address, bases)?;
+        let fdeptr = match fdeptr {
+            Pointer::Direct(x) => x,
+            _ => unreachable!(),
+        };
+
+        let eh_frame_ptr = match self.hdr.eh_frame_ptr() {
+            Pointer::Direct(x) => x,
+            _ => unreachable!(),
+        };
+
+        // Calculate the offset in the EhFrame section
+        let offset = R::Offset::from_u64(fdeptr - eh_frame_ptr)?;
+        let mut input = &mut frame.section().clone();
+        input.skip(offset)?;
+
+        let entry = parse_cfi_entry(bases, frame.clone(), &mut input)?;
+        let target_fde = match entry {
+            Some(CieOrFde::Fde(fde)) => Some(fde.parse(|offset| frame.cie_from_offset(bases, offset))?),
+            Some(CieOrFde::Cie(_)) => unimplemented!(),
+            None => None
+        };
+        match target_fde {
+            Some(ref fde) if fde.contains(address) => Ok((*fde).clone()),
+            _ => Err(Error::NoUnwindInfoForAddress)
+        }
+    }
 }
 
 /// `EhFrame` contains the frame unwinding information needed during exception


src/cfi.rs
```diff
@@ -269,6 +269,37 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+        };
+
+        // Calculate the offset in the EhFrame section
+        let offset = R::Offset::from_u64(fdeptr - eh_frame_ptr)?;
+        let mut input = &mut frame.section().clone();
+        input.skip(offset)?;
+
+        let entry = parse_cfi_entry(bases, frame.clone(), &mut input)?;
+        let target_fde = match entry {
+            Some(CieOrFde::Fde(fde)) => Some(fde.parse(|offset| frame.cie_from_offset(bases, offset))?),
```
main--: This might be inefficient, one may want to cache CIEs.

src/cfi.rs
```diff
@@ -269,6 +269,37 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+            Some(CieOrFde::Cie(_)) => unimplemented!(),
```
main--: I never was entirely sure about this (hence the `unimplemented!()` but I believe this should actually be unreachable.

Either way, none of these cases should panic - returning an error is the correct reaction for the gimli API.

src/cfi.rs
```diff
@@ -269,6 +269,37 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+            None => None
+        };
+        match target_fde {
+            Some(ref fde) if fde.contains(address) => Ok((*fde).clone()),
```
main--: The `ref` here is required because of the pattern guard, right? You can get around the awkwardness of cloning for no reason by moving this into the match above (since the other arms of that match are dead anyways).

src/cfi.rs
```diff
@@ -269,6 +269,37 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+    pub fn lookup_and_parse(&self, address: u64, bases: &BaseAddresses, frame: EhFrame<R>) -> Result<FrameDescriptionEntry<EhFrame<R>, R, R::Offset>> {
+        let fdeptr = self.lookup(address, bases)?;
+        let fdeptr = match fdeptr {
+            Pointer::Direct(x) => x,
+            _ => unreachable!(),
+        };
+
+        let eh_frame_ptr = match self.hdr.eh_frame_ptr() {
+            Pointer::Direct(x) => x,
+            _ => unreachable!(),
```
main--: Not sure if it makes sense to deal with indirect pointers here. I certainly have never encountered any. Returning an error is probably fine.

src/cfi.rs
main--: (which, by the way, is the same reason why `fde.parse()` takes a closure)

roblabla: I see. Should we return PartialFrameDescriptionEntry ?

src/cfi.rs
roblabla: It should be unreachable!() yeah, eh_frame_hdr points to FDEs only, not CIEs (that would make no sense)

src/cfi.rs
main--: Either this or also take a closure parameter (which has the advantage that we can keep the `contains` check here which is a potential footgun otherwise).

roblabla:
I'll add an example in the doc, and I think this is ready to merge.

roblabla:
I added a test and an example (and hopefully fixed compilation, woops). If it compiles and the tests pass, this should be ready to go.

src/cfi.rs
```diff
@@ -269,6 +269,67 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+            Pointer::Direct(x) => x,
+            _ => return Err(Error::UnsupportedPointerEncoding),
+        };
+
+        // Calculate the offset in the EhFrame section
+        let offset = R::Offset::from_u64(fdeptr - eh_frame_ptr)?;
+        let mut input = &mut frame.section().clone();
+        input.skip(offset)?;
+
+        let entry = parse_cfi_entry(bases, frame.clone(), &mut input)?;
```
philipc: Unneeded clone.

src/cfi.rs
```diff
@@ -269,6 +269,67 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+        let entry = match entry {
+            Some(CieOrFde::Fde(fde)) => Ok(fde.parse(cb)?),
+            Some(CieOrFde::Cie(_)) => Err(Error::NotFdePointer),
+            None => Err(Error::NoUnwindInfoForAddress)
+        }?;
```
philipc: Style nit: I think I prefer returns in the match arms instead.

src/cfi.rs
```diff
@@ -269,6 +269,67 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+    /// # let eh_frame_hdr: ParsedEhFrameHdr<EndianRcSlice<NativeEndian>> = unimplemented!();
+    /// # let addr = 0;
+    /// # let address_of_cfi_section_in_memory = unimplemented!();
+    /// # let address_of_text_section_in_memory = unimplemented!();
+    /// # let address_of_data_section_in_memory = unimplemented!();
+    /// # let address_of_the_start_of_current_func = unimplemented!();
+    /// # let bases = BaseAddresses::default()
+    /// #     .set_cfi(address_of_cfi_section_in_memory)
+    /// #     .set_text(address_of_text_section_in_memory)
+    /// #     .set_data(address_of_data_section_in_memory);
```
philipc: Simpler to  `let bases = unimplemented!();`?

roblabla:
All ready.

src/cfi.rs
```diff
@@ -269,6 +269,64 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
...
+    /// let fde = table.lookup_and_parse(addr, &bases, eh_frame.clone(),
+    ///     |offset| eh_frame.cie_from_offset(&bases, offset))?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn lookup_and_parse<F>(&self, address: u64, bases: &BaseAddresses, frame: EhFrame<R>, cb: F) -> Result<FrameDescriptionEntry<EhFrame<R>, R, R::Offset>>
+    where
+        F: FnMut(EhFrameOffset<R::Offset>) -> Result<CommonInformationEntry<EhFrame<R>, R, R::Offset>>
+    {
+        let fdeptr = self.lookup(address, bases)?;
```
philipc: Is it possible for this to be encoded as `DW_EH_PE_pcrel`? If so, we'll get this wrong, since `bases.cfi` will be for the `eh_frame` section.

Issue #130, opened 2016-09-25
Support GNU extensions to CFI


fitzgen:
http://refspecs.linux-foundation.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/dwarfext.html

<table>
<tbody></td></tr><tr><td>DW_CFA_GNU_args_size</td><td>0x2e</td><td>The DW_CFA_GNU_args_size instruction takes an unsigned LEB128 operand
representing an argument size. This instruction specifies the total of the
size of the arguments which have been pushed onto the stack.</td></tr><tr><td>DW_CFA_GNU_negative_offset_extended</td><td>0x2f</td><td>The DW_CFA_def_cfa_sf instruction takes two operands: an unsigned LEB128 value representing a register
number and an unsigned LEB128 which represents the magnitude of the offset. This instruction is identical
to DW_CFA_offset_extended_sf except that the operand is subtracted to produce the offset. This
instructions is obsoleted by DW_CFA_offset_extended_sf.</td></tr></tbody></table>

----------------------------

I think the first one is an assertion?

The second one is obsolete, but we should probably support it for backwards compatibility.

fitzgen:
There's also `DW_EH_PE_indirect` which (AFAICT) means that the resulting pointer is actually an indirect pointer to the resulting pointer. Seems safe. Might not want to support this one.

roblabla:
So I looked up `DW_CFA_GNU_args_size` since unwind-rs ran into issues with executables using it. LLVM's libunwind (which I base most of my findings on because I find the code much cleaner than GCC's).

- When it encounters it, it stores the result in [spExtraArgSize](https://github.com/llvm-mirror/libunwind/blob/b154837e4b1891ac1347e22941d81413b3361f00/src/DwarfParser.hpp#L652), and then in [info.gp](https://github.com/llvm-mirror/libunwind/blob/master/src/UnwindCursor.hpp#L982).

- When it needs to set the IP register, it [additionally increments](https://github.com/llvm-mirror/libunwind/blob/940b92a8d45f2a24c7f94524ce136f0c1c1c5043/src/libunwind.cpp#L196) the SP register by that value.

I suppose we could add an `sp_extra_arg_size` field to UnwindTableRow, and unwinders would be expected to reimplement the second step?

main--:
>When it needs to set the IP register, it additionally increments the SP register by that value.

[...]

PR #319, opened 2018-08-06
Properly handle DW_CFA_GNU_args_size

roblabla:
Pretty simple, really. It stores the value of DW_CFA_GNU_args_size inside the UnwindTableRow. Defaults to 0.

As we talked in #130 unwinders should add this value to SP when setting IP.

CC @main-- 

So indeed, i686-pc-windows-gnu generates those. I've been unable to make any other target generate them, even with custom target changing the data-layout. So testing this is a bit painful since there's no 32-bit x86 backend to unwind-rs (I might work on this soon though!)


philipc:
`cfi::tests::size_of_unwind_ctx` needs fixing.

roblabla:
Fixed.

PR #324, opened 2018-08-31
Make line::StateMachine covariant wrt R

philipc:
This is a breaking change for readers with `Offset != usize`. Not keen on doing this, but I don't know a better way to get variance...

### DIFF
--- a/src/line.rs
+++ b/src/line.rs
@@ -2385,4 +2385,14 @@ mod tests {
         let expected_registers = initial_registers.clone();
         assert_exec_opcode(header, initial_registers, opcode, expected_registers, false);
     }
+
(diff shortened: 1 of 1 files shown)


fitzgen:
The motivation is to allow `StateMachine<MyReader<'a>>` be passed to something that takes `StateMachine<MyReader<'b>>` where `'a: 'b`, correct?

Can we add a simple test case that will fail to compile if that is not the case?

Otherwise LGTM, unless I am misunderstanding something.

philipc:
Correct. I've added a test case.

fitzgen:
> Correct. I've added a test case.

Thanks!

PR #330, opened 2018-09-20
dwarfdump: handle empty unit in dump_line

### LINKED ISSUE #329: Zero length compile units aren't handled well
running dwarfdump -l on the attached dwarf file gives
```
Failed to dump '/Users/jrmuizel/tmp/inf/target/release/inf.dSYM/Contents/Resources/DWARF/inf-92af5f0a940e8864': Expected a DIE but none was found
```

It's trying to find a DIE in this compile unit:
```
0x000168f3: Compile Unit: length = 0x00000007  version = 0x0002  abbr_offset = 0x00000000  addr_size = 0x08  (next CU at 0x000168fe)
```

philipc:
If an error occurs when dumping line info for a unit, then continue to the next unit instead of exiting.

Fixes #329

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -49,7 +49,7 @@ impl error::Error for Error {
     fn description(&self) -> &str {
         match *self {
             Error::GimliError(ref err) => err.description(),
-            Error::IoError => "An I/O error occurred while reading.",
+            Error::IoError => "An I/O error occurred while writing.",
             Error::MissingDIE => "Expected a DIE but none was found",
         }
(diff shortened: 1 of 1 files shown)


philipc:
Hide whitespace changes when viewing the diff.

philipc:
We probably should do something like this for every `while` loop in dwarfdump, and the `io::Error` handling is ugly...

philipc:
Maybe we should just be exiting immediately on IO errors. I guess we could hack that in the `impl From<io::Error> for Error`.

PR #333, opened 2018-10-03
fix a few clippy warnings

llogiq:


src/value.rs
```diff
@@ -319,8 +319,8 @@ impl Value {
             ValueType::U32 => Value::U32(bits as u32),
             ValueType::I64 => Value::I64(bits as i64),
             ValueType::U64 => Value::U64(bits),
-            ValueType::F32 => Value::F32(unsafe { mem::transmute(bits as u32) }),
-            ValueType::F64 => Value::F64(unsafe { mem::transmute(bits) }),
+            ValueType::F32 => Value::F32(f32::from_bits(bits as u32)),
+            ValueType::F64 => Value::F64(f64::from_bits(bits)),
```
philipc: Is this change due to a clippy warning? If so, should there be a similar warning to use `f32::to_bits` and `f64::to_bits` in the previous block?

philipc: https://github.com/rust-lang-nursery/rust-clippy/issues/1675 already lists these.

PR #336, opened 2018-10-06
Fix documentation for Piece

philipc:


src/op.rs
```diff
@@ -302,11 +302,20 @@ where
...
-    /// piece takes its size from the enclosed location.
+    /// If given, the size of the piece in bits.  If `None`, there
+    /// must be only one piece whose size is all of the object.
     pub size_in_bits: Option<u64>,
-    /// If given, the bit offset of the piece.  If `None`, then the
-    /// piece starts at the next byte boundary.
+    /// If given, the bit offset of the piece within the location.
+    /// If the location is a `Location::Register` or `Location::Value`,
+    /// then this offset is from the least significant bit end of
+    /// the register or value.
```
tromey: Is this true for big-endian machines?  I can't recall offhand.

philipc: It's what the DWARF 5 standard says. I haven't looked at what gdb does.

tromey: I think I was remembering http://lists.dwarfstd.org/pipermail/dwarf-discuss-dwarfstd.org/2016-December/004286.html and http://dwarfstd.org/ShowIssue.php?issue=161206.2.

Generally I suppose the docs would be improved by references to the relevant sections of the standard.

Anyway as long as it conforms to what's written, it is fine; clarifications can be done later.

philipc: Thanks for the links. So this is all complicated... I think what I've written conforms to the standard, but it looks like gdb doesn't exactly follow the standard here: it treats `DW_OP_piece(size)` the same as `DW_OP_bit_piece(size*8, offset: 0)`. There's a [long discussion](https://sourceware.org/ml/gdb/2016-01/msg00013.html) about this linked to from the email you gave. I'm going to go with what I've written, but there will likely be need for clarifications later.

tromey:
>  it looks like gdb doesn't exactly follow the standard here

A bunch of patches in this area landed in June 2017, from Andreas Arnez.  So I'd imagine (I didn't check) that gdb is conforming now.

PR #337, opened 2018-10-22
[WIP]  dwarfdump: handle relocations

philipc:
WIP: needs a release of the object crate

Fixes #180

PR #339, opened 2018-11-02
Create a read module

philipc:
I've moved all the parser code into the read module. There will be some small parts that can be moved out again (eg Format, Register, Range), but that can be done as needed.

I've exported the constants module. This prevents the constants from cluttering the main page of the docs. The constants are still re-exported in the root, but rust is smart enough not to repeat them in the docs.

Also ran rustfmt, so you probably want to look at individual commits.

This will conflict with #337. I'll rebase whichever is needed.


src/lib.rs
```diff
@@ -225,7 +225,8 @@ pub use stable_deref_trait::{CloneStableDeref, StableDeref};
 mod arch;
 pub use arch::*;
 
-mod constants;
+pub mod constants;
+// For backwards compat.
```
fitzgen: We could also do `#[doc(hidden)]` to ensure that it never shows up in rustdoc.

philipc: It's only one line in the re-export section of the doc, which I think is fine since we are in fact re-exporting it. I'm not trying to discourage users from using the re-export if it's more convenient for them, I just don't want all the constants drowning out the rest of the main page of the docs. I'm not sure what's idiomatic for re-exports in rust though.

PR #340, opened 2018-11-09
[WIP] Implement writing of .debug_info, .debug_abbrev and .debug_str

philipc:
This can read the self test data for these sections, write it to new sections, and read it back again, and get the same number of entries. Correctness beyond that hasn't been checked, and isn't trivial to do, but I will work on this some more.

To do:
- expand API for modification of units/entries/attributes
- tests
- docs
- ...

I'm creating this PR to try a travis build, and in case anyone wants to comment on the design before I start polishing it.


philipc:
This now has good test coverage and minimal docs.

Some notes about the implementation:
- This is more of a higher level builder API, rather than low-level writes. This is required though in order to correctly handle references.
- References are stored as `Id` values (eg `UnitEntryId`), rather than offsets. When writing, these are initially written as zeroes, then we go back and write the offsets once we know what they are. Same for unit lengths.
- A lot of `AttributeValue` variants still use offsets, but these will change to ids once we support writing more sections.
- Relocations can be supported if the user implements `Writer::write_address/write_offset` (similar to what was done for reading relocations).

src/write/abbrev.rs
```diff
@@ -0,0 +1,222 @@
...
+            has_children,
+            attributes,
+        }
+    }
+
+    /// Write the abbreviation to the `.debug_abbrev` section.
+    pub fn write<W: Writer>(&self, w: &mut DebugAbbrev<W>) -> Result<()> {
+        w.write_uleb128(self.tag.0)?;
+        w.write_u8(if self.has_children {
+            constants::DW_CHILDREN_yes.0
```
fitzgen: It seems our `dw!` macro for constants could also generate a thin little `write` method. Although, I suppose most of the time we want to write LEB128s, so maybe it wouldn't be super useful and this location would still be the exception.

src/write/endian_vec.rs
```diff
@@ -0,0 +1,107 @@
+use vec::Vec;
+
+use endianity::Endianity;
+use write::{Error, Result, Writer};
+
+/// A `Vec<u8>` with endianity metadata.
```
fitzgen: Worth pointing out in the doc comment that the raison d'etre for `EndianVec` is that it is a `Writer`.

src/write/str.rs
```diff
@@ -0,0 +1,182 @@
...
+// Requirements:
+// - values are `[u8]`, null bytes are not allowed
+// - insertion returns a fixed id
+// - inserting a duplicate returns the id of the existing value
+// - able to convert an id to a section offset
+// Optional?
+// - able to get an existing value given an id
+//
+// Limitations of current implementation (using IndexSet):
+// - inserting a duplicate requires an allocation or a double lookup
```
fitzgen: This is one of my longest standing annoyances with hash tables in Rust in general (mostly std, but apparently indexmap as well). I suspect a duplicated lookup is probably cheaper than an allocation (I haven't read further to see what you ended up implementing).

src/write/str.rs
```diff
@@ -0,0 +1,182 @@
...
+//   hard to implement `get`.
+#[derive(Debug, Default)]
+pub struct StringTable {
+    strings: IndexSet<Vec<u8>>,
+}
+
+impl StringTable {
+    /// Add a string to the string table and return its id.
+    ///
+    /// `bytes` must not contain a null byte.
```
fitzgen: Worth clarifying that this function will panic if `bytes` does contain a null byte.

src/write/unit.rs
```diff
@@ -0,0 +1,1914 @@
...
+/// A Debugging Information Entry (DIE).
+///
+/// DIEs have a set of attributes and optionally have children DIEs as well.
+#[derive(Debug)]
+pub struct DebuggingInformationEntry {
+    id: UnitEntryId,
+    parent: Option<UnitEntryId>,
+    tag: constants::DwTag,
+    /// Whether to emit `DW_AT_sibling`.
+    // TODO: should this be automatic?
```
fitzgen: What heuristics would we use to decide whether to emit this? Whether the children forest is of a certain size? It seems like maybe certain DIEs always end up having their children walked, while others are skipped past, so maybe there would be a whitelist of tags that should get the sibling links.

Anyways, agree that for now this should be manual.

src/write/unit.rs
```diff
@@ -0,0 +1,1914 @@
...
+        /* TODO: implement sibling support
+        if self.sibling {
+        }
+        */
+        for attr in &self.attrs {
+            attr.write(w, unit, strings, unit_refs, debug_info_refs)?;
+        }
+        if !self.children.is_empty() {
+            for child in &self.children {
+                unit.entries[child.0].write(
```
fitzgen: It *might* be worth constructing the tree relationship eagerly once in `CompilationUnitHeader::write` or maintain it constantly while adding new DIEs. Because the DIEs can be mutated after creation and aren't append only, someone could (presumably accidentally) create DAGs and even cycles in the DIE "tree".

This creates two problems for us:

1. If there is a cycle, then we will blow the stack here because of the recursive implementation.

2. For a DAG, we can maybe generate bad debug info (not actually totally sure if this implementation would create copies of shared tails of the graph or emit bad debug info -- either way, it seems like something we probably want to catch and handle early and notify the user that they are doing something buggy)

src/lib.rs
```diff
@@ -241,5 +243,8 @@ pub mod read;
 // For backwards compat.
 pub use read::*;
 
+#[cfg(feature = "std")]
+pub mod write;
```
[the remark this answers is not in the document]
philipc: I'll do this in a follow up PR.

src/write/abbrev.rs
philipc: I'm not sure about the usefulness:
- they would be simple one line functions
- many of them could be written in multiple forms that the caller would need to decide, since attribute values could potentially use either LEB128 or a fixed size, and this needs to match the form, so it's better to have this code closer to the form decision.

src/write/str.rs
philipc: It's actually allocation for duplicate, and double lookup for non-duplicate, so will depend on how often there are duplicates. I did the simplest (allocation), and we can change based on benchmarks.

src/write/unit.rs
philipc: I was thinking maybe we could do it based on a tag whitelist. The whitelist members would be chosen based on typical access patterns of debuggers/unwinders.

src/write/unit.rs
philipc: Currently we maintain the tree relationship constantly while adding new DIEs. Parents are specified when adding, and mutation doesn't allow changing parents. I'll add doc to make this clear.

bjorn3:
Hooray

PR #344, opened 2018-11-20
write: implement DW_AT_sibling

philipc:


src/write/unit.rs
```diff
@@ -439,13 +444,19 @@ impl DebuggingInformationEntry {
         offsets.entries[self.id.0] = w.offset();
         let code = abbrevs.add(self.abbreviation(unit.format)?);
         w.write_uleb128(code)?;
-        /* TODO: implement sibling support
-        if self.sibling {
-        }
-        */
+
+        let sibling_offset = if self.sibling {
```
fitzgen: Should we only write DW_AT_sibling if the DIE has children? I.e. make this condition be `if self.sibling && !self.children.is_empty()`

philipc: It makes it less flexible, but if there's never a valid reason to want to emit the attribute in this case, then that's fine. I've added a commit to do this.

PR #346, opened 2018-12-08
write: implement .debug_line

philipc:
The basics are there, but I expect this to evolve a bit as it gets used, both in API and internal data structures.

I've tried to take DWARF version 5 into account, but the support for it is still incomplete.

### DIFF
--- a/src/write/line.rs
+++ b/src/write/line.rs
@@ -1,0 +1,1517 @@
+use indexmap::{IndexMap, IndexSet};
+use std::ops::{Deref, DerefMut};
+use vec::Vec;
+
+use common::{DebugLineOffset, Format};
+use constants;
+use leb128;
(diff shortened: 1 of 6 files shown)


src/write/line.rs
```diff
@@ -0,0 +1,1517 @@
...
+            is_statement: DEFAULT_IS_STATEMENT,
+            basic_block: false,
+            prologue_end: false,
+            epilogue_begin: false,
+
+            isa: 0,
+        }
+    }
+
+    fn default(version: u16) -> Self {
```
fitzgen: Nitpick: since this can't be a `Default` impl, it should probably be called `new`, just so that it isn't slightly jarring for readers who would otherwise expect a `default` method to be part of a `Default` impl.

PR #351, opened 2019-01-03
Use separate base address structs for .eh_frame and .eh_frame_hdr

### LINKED ISSUE #350: `EhHdrTable::lookup` requires a base address to `.eh_frame_hdr` in `BaseAdresses::data` instead of a base address to `.data` as the docs suggest
I've been adding support for `.eh_frame_hdr` to [our profiler](https://github.com/nokia/nperf); after a lot of debugging and tweaking I've managed to get it to work, but it was **very** confusing.

First and foremost, currently `EhHdrTable::lookup`'s prototype looks like this:

```rust
pub fn parse(&self, bases: &BaseAddresses, addr_size: u8) -> Result<ParsedEhFrameHdr<R>>
```

I see this and think - great! Let's initialize the `bases` like I usually do - `set_text` with the base address to `.text`, `set_data` with the base address to `.data` and `set_cfi` with the base address to `.eh_frame`. However, this didn't work - `EhHdrTable::lookup` (and consequently also `EhHdrTable::lookup_and_parse` since it uses `lookup`) returned incorrect results! So after a lot of trial-and-error I got it to actually work correctly with `BaseAddresses` initialized like this:

```
let mut bases = BaseAddresses::default();
bases = bases.set_cfi( eh_frame_base );
bases = bases.set_data( eh_frame_hdr_base );
```

The docs for `gimli::BaseAddresses::data` say that:

> The address of the `.data` section in memory.

...and yet here I am setting it to the base address of the `.eh_frame_hdr` instead of `.data`. Is this correct? Or is this a bug in `gimli`?

Another minor issue is that the docs for `gimli::BaseAddresses::cfi` say that:

> The address of the current CFI unwind section (`.eh_frame` or `.debug_frame`) in memory.

However `EhFrameHdr::parse` requires the `cfi` to be the address to `.eh_frame_hdr`, which is not obvious from the docs.

philipc:
PC relative pointers need to use different bases for these sections.
Also, data relative pointers have different meanings, and the existing
documentation for `set_data` was wrong for both sections.

Fixes #350 
cc @koute

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -519,7 +519,10 @@ mod cfi {
         let eh_frame = read_section("eh_frame");
         let eh_frame = EhFrame::new(&eh_frame, LittleEndian);
 
-        let bases = BaseAddresses::default().set_cfi(0).set_data(0).set_text(0);
+        let bases = BaseAddresses::default()
+            .set_eh_frame(0)
+            .set_got(0)
+            .set_text(0);
 
         b.iter(|| {
             let mut entries = eh_frame.entries(&bases);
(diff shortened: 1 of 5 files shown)


koute:
Looks good to me!

PR #352, opened 2019-01-09
Add read::Dwarf

philipc:
The motivation is to reduce the long parameter lists containing section readers, which is only going to get worse as we implement the new DWARF 5 sections.

Currently I'm only including the sections used when parsing `.debug_info`, which is the section that contains the most references to other sections. There will only be one of these structs per file, so it's probably not a problem to add other sections too if needed.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -874,7 +874,7 @@ where
     let out = io::stdout();
     writeln!(&mut BufWriter::new(out.lock()), "\n.debug_info")?;
 
-    let units = dwarf.debug_info.units().collect::<Vec<_>>().unwrap();
+    let units = dwarf.units().collect::<Vec<_>>().unwrap();
     let process_unit =
         |unit: CompilationUnitHeader<R, R::Offset>, buf: &mut Vec<u8>| -> Result<()> {
             let abbrevs = match dwarf.abbreviations(&unit) {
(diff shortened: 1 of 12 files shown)


bjorn3:
Could there be a new release soon? The latest on crates.io doesn't have the writer support yet.

philipc:
My plan was to get DWARF 5 done for the .debug_line read/write, but I've gotten sidetracked doing other DWARF 5 stuff that will make it easier to do that, and I've also been a bit lacking in time recently. I also want to add something that makes it easier to write DWARF for a single compilation unit. There will be more breaking changes for this, so I wasn't going to do a release just yet, but yes hopefully soon.

bjorn3:
I have rebased your rustc_codegen_cranelift debuginfo branch and I am working on extending it a bit.

```
(gdb) info frame 
Stack level 0, frame at 0x7fffffffdd80:
 rip = 0x555555554a00 in mini_core_hello_world::start::h1e715bf6fab28115; saved rip = 0x555555555dca
 called by frame at 0x7fffffffdd88
 Arglist at 0x7fffffffdd70, args: main=..., _argc=<optimized out>, _argv=...
 Locals at 0x7fffffffdd70, Previous frame's sp is 0x7fffffffdd80
 Saved registers:
  rip at 0x7fffffffdd78
```

Notice the args. It is really annoying when gdb silently ignores stuff (eg `DW_AT_name` for functions and `DW_TAG_variable` definitions)

PR #353, opened 2019-01-11
Make field of Expression public

bjorn3:


philipc:
Not sure about this. Maybe this is okay temporarily, but I prefer to have proper writing of expressions instead.

bjorn3:
The only way of creating one currently is using `mem::transmute`, which I like to avoid.

PR #370, opened 2019-01-31
read: don't use rnglists and loclists header

philipc:
These headers don't give information that is required for parsing
range lists or location lists, because the encoding of these lists
must match the encoding of the compilation unit.

Additionally, while the standard seems to indicate that each section
has only one of these headers, in practice this is not true.
Furthermore, when there are multiple headers, they do not have to
each have the same encoding, so our parsing was wrong in this case
because we only used the encoding from the first header.

### DIFF
--- a/src/read/loclists.rs
+++ b/src/read/loclists.rs
@@ -115,6 +115,7 @@ impl Default for LocListsHeader {
 
 impl LocListsHeader {
     /// Return the serialized size of the table header.
+    #[allow(dead_code)]
     #[inline]
     fn size(self) -> u8 {
         // initial_length + version + address_size + segment_selector_size + offset_entry_count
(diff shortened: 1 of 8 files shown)


philipc:
Also use one `RangeListTable` per unit when writing, since units can have different encodings, and so the range list tables must too.

src/read/loclists.rs
```diff
@@ -154,33 +157,17 @@ fn parse_header<R: Reader>(input: &mut R) -> Result<LocListsHeader> {
 pub struct LocationLists<R: Reader> {
     debug_loc: DebugLoc<R>,
     debug_loclists: DebugLocLists<R>,
-    header: LocListsHeader,
 }
 
 impl<R: Reader> LocationLists<R> {
     /// Construct a new `LocationLists` instance from the data in the `.debug_loc` and
     /// `.debug_loclists` sections.
     #[allow(clippy::new_ret_no_self)]
```
bjorn3: I believe this can be removed.

philipc: Yep, thanks.

Issue #377, opened 2019-02-01
index out of bounds: the len is 0 but the index is 0'


bjorn3:
https://github.com/bjorn3/rustc_codegen_cranelift/tree/update_gimli

```
thread 'rustc' panicked at 'index out of bounds: the len is 0 but the index is 0', /rustc/147311c5fc62537da8eb9c6f69536bec6719d534/src/libcore/slice/mod.rs:2535:10
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:70
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:58
             at src/libstd/panicking.rs:200
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:215
   4: rustc::util::common::panic_hook
   5: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:482
   6: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:385
   7: rust_begin_unwind
             at src/libstd/panicking.rs:312
   8: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   9: core::panicking::panic_bounds_check
             at src/libcore/panicking.rs:61
  10: <usize as core::slice::SliceIndex<[T]>>::index
             at /rustc/147311c5fc62537da8eb9c6f69536bec6719d534/src/libcore/slice/mod.rs:2535
  11: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/147311c5fc62537da8eb9c6f69536bec6719d534/src/libcore/slice/mod.rs:2392
  12: <alloc::vec::Vec<T> as core::ops::index::Index<I>>::index
             at /rustc/147311c5fc62537da8eb9c6f69536bec6719d534/src/liballoc/vec.rs:1678
  13: gimli::write::range::RangeListOffsets::get
             at /home/bjorn/.cargo/git/checkouts/gimli-a0d193bd15a5ed96/c679a14/src/write/mod.rs:76
  14: gimli::write::unit::AttributeValue::write
[...]

PR #379, opened 2019-02-02
write: ensure id's are used with the correct table

philipc:
Add a BaseId field to all tables and id types, and check these
match when using the id to access to a table entry.

Makes it easier to diagnose #377

### DIFF
--- a/src/write/line.rs
+++ b/src/write/line.rs
@@ -6,20 +6,21 @@ use common::{DebugLineOffset, Encoding, Format};
 use constants;
 use leb128;
 use write::{
-    Address, DebugLineStrOffsets, DebugStrOffsets, Error, LineStringId, LineStringTable, Result,
-    Section, SectionId, StringId, Writer,
+    Address, BaseId, DebugLineStrOffsets, DebugStrOffsets, Error, LineStringId, LineStringTable,
+    Result, Section, SectionId, StringId, Writer,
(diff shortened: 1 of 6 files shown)


src/write/line.rs
```diff
@@ -37,7 +38,8 @@ impl LineProgramTable {
     /// Panics if `id` is invalid.
     #[inline]
     pub fn get(&self, id: LineProgramId) -> &LineProgram {
-        &self.programs[id.0]
+        assert_eq!(self.base_id, id.base_id);
```
bjorn3: Maybe make these `debug_assert!`? That way release mode doesn't pay this cost.

src/write/mod.rs
```diff
@@ -56,24 +57,53 @@ macro_rules! define_section {
     };
 }
 
+macro_rules! define_id {
+    ($name:ident, $docs:expr) => {
+        #[doc=$docs]
+        #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+        pub struct $name {
+            base_id: BaseId,
```
bjorn3: Maybe add `#[cfg(debug_assertions)]`. That would save a few bytes.

philipc: I'll try changing `BaseId`'s type to `()` in release mode.

PR #382, opened 2019-02-09
write: improve high level API

philipc:
Notably, there is now `Dwarf` (for multiple units) and `DwarfUnit` (for single units), and we write to `Sections`. Also `LineProgramTable` is gone.

### DIFF
--- a/src/write/line.rs
+++ b/src/write/line.rs
@@ -496,6 +496,10 @@ impl LineProgram {
     }
 
     /// Write the line number program to the given section.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `self.is_none()`.
     pub fn write<W: Writer>(
         &self,
         w: &mut DebugLine<W>,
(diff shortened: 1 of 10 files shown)


src/write/line.rs
```diff
@@ -456,6 +503,10 @@ impl LineProgram {
         debug_line_str_offsets: &DebugLineStrOffsets,
         debug_str_offsets: &DebugStrOffsets,
     ) -> Result<DebugLineOffset> {
+        if self.none {
+            return Err(Error::CannotWriteEmptyLineProgram);
```
fitzgen: This seems like something we should just assert -- either you create a non-empty line program and can write to it, or you create an empty one and you should never write it. But it is completely up to you the API user, not any external input, and so panics will help you catch your own mistakes more quickly.

philipc: Yeah I'm not sure about which is best. I probably tend to use errors instead of panics if we're already returning a result type, especially if it is something that is hard to ensure it is right. In this case, I think it's reasonable for the caller to check `!self.is_none()`  so I'll change it to that.

PR #386, opened 2019-02-14
Update high level doc

philipc:


philipc:
Added an example to the write docs (and ping @fitzgen).

src/read/mod.rs
```diff
@@ -1,4 +1,175 @@
...
+//!   * [`ddbug`](https://github.com/philipc/ddbug), a utility giving insight into
+//!     code generation by making debugging information readable
+//!
+//!   * [`dwprod`](https://github.com/fitzgen/dwprod), a tiny utility to list the
+//!     compilers used to create each compilation unit within a shared library or
+//!     executable (via `DW_AT_producer`)
+//!
+//!   * [`dwarf-validate`](http://github.com/gimli-rs/gimli/blob/master/examples/dwarf-validate.rs),
+//!     a program to validate the integrity of some DWARF and its references
+//!     between sections and ocmpilation units.
```
fitzgen:
```suggestion
//!     between sections and compilation units.
```

src/read/mod.rs
```diff
@@ -1,4 +1,175 @@
...
+//! having iterated through many items.
+//!
+//! To use `gimli` iterators with `FallibleIterator`, import the crate and trait
+//! into your code:
+//!
+//! ```
+//! // Add the `fallible-iterator` crate. Don't forget to add it to your
+//! // `Cargo.toml`, too!
+//! extern crate fallible_iterator;
+//! extern crate gimli;
```
fitzgen: New breaking version could be an opportunity to move to 2018 edition, too FWIW.

src/write/mod.rs
```diff
@@ -1,4 +1,64 @@
...
+//!         AttributeValue::RangeListRef(range_list_id),
+//!     );
+//!     // Create a `Vec` for each DWARF section.
+//!     let mut sections = Sections::new(EndianVec::new(gimli::LittleEndian));
+//!     // Finally, write the DWARF data to the sections.
+//!     dwarf.write(&mut sections)?;
+//!     sections.for_each(|id, data| {
+//!         // Here you can add the data to the output object file.
+//!         Ok(())
+//!     })
```
fitzgen:
```suggestion
//!     });
```

philipc: No, because it's returning the result from this function. Could instead do:
```rust
    })?;
    Ok(())
```

src/read/mod.rs
philipc: It's been a while since we did a release, so I was going to leave that until after this release.

src/write/mod.rs
fitzgen: Woops, I didn't look closely enough at the type signatures!

PR #387, opened 2019-02-15
Write fixes

philipc:
See individual commits.


philipc:
I shouId have done a better commit message for that. I think it is an unusual thing to do, and my concern is that it works worse for naive consumers (they'll make a copy of the decoded abbreviation table for every CU). Wasn't something I wanted to spend time on now, so went with the conservative approach.

PR #393, opened 2019-02-24
Fix variance of read::Unit

philipc:


src/read/dwarf.rs
```diff
@@ -533,4 +537,14 @@ mod tests {
             x
         }
     }
+
+    /// Ensure that `Unit<R>` is covariant wrt R.
+    #[test]
+    fn test_dwarf_unit_variance() {
+        /// This only needs to compile.
+        #[allow(dead_code)]
+        fn f<'a: 'b, 'b, E: Endianity>(x: Unit<EndianSlice<'a, E>>) -> Unit<EndianSlice<'b, E>> {
```
fitzgen: A little trick: can name this function `_f` and then you don't need to `#[allow(dead_code)]` ;)

Issue #296, opened 2018-04-28
Improve the ergonomy of dealing with `.eh_frame` and `.debug_frame`


koute:
Often we might not care whenever we're dealing with `.eh_frame` or `.debug_frame` and we just want to treat those two the same.

For dealing with `.eh_frame` we have to use the `EhFrame` type while for `.debug_frame` we use `DebugFrame`. Those two are then used as type parameters in `UninitializedUnwindContext`, `InitializedUnwindContext`, `CommonInformationEntry`, etc.

This is a problem when we just don't care which one it is, and we just want to dynamically dispatch to the relevant one at runtime based on whenever an `.eh_frame` or a `.debug_frame` was loaded.

Another problem is that manually caching the unwind contexts can be a little bit awkward, especially since the types are different.

Such an API might perhaps be a little too high level and specific to my use case, but something like this could be used as an inspiration to help think about what could be done here: (Copied from [here](https://github.com/nokia/nperf/blob/master/src/frame_descriptions.rs).)

```rust
pub struct FrameDescriptions< E: Endianity > {
    ...
}

impl< E: Endianity > FrameDescriptions< E > {
    pub fn load( binary: &Arc< BinaryData > ) -> Option< Self > {
        ...
    }

    pub fn find_unwind_info< 'a >( &'a self, ctx_cache: &'a mut ContextCache< E >,
        mappings: &[AddressMapping], address: u64 ) -> Option< UnwindInfo< 'a, E > >
    {
        ...
    }
}

pub struct UnwindInfo< 'a, E: Endianity + 'a > {
    ...
}

impl< 'a, E: Endianity > Drop for UnwindInfo< 'a, E > {
    #[inline]
    fn drop( &mut self ) {
        ... // Cache the context.
    }
}

impl< 'a, E: Endianity > Deref for UnwindInfo< 'a, E > {
    type Target = UnwindTableRow< EndianBuf< 'a, E > >;

    #[inline]
    fn deref( &self ) -> &Self::Target {
        &self.row
    }
}
```

Basically this treat both types of sections in an uniform way, and completely abstracts away the caching of the contexts, which makes it pretty nice to use. You just `load()` the sections, and then call `find_unwind_info`.

philipc:
One possible solution: get rid of the traits and use the same type for both sections, with a boolean parameter to distinguish between the behavior.

philipc:
I've done two changes to address this:
[...]

PR #400, opened 2019-03-06
cfi: various cleanups

philipc:
Try to simplify and makes things more consistent. I was looking into this as part of seeing if we could do more for #296, but I didn't come up with anything significant.

The commit I'm most unsure about is eeae767 (Change callbacks to match UnwindSection::cie_from_offset).

There are minor performance improvements:

```
 name                                                       before ns/iter  after ns/iter  diff ns/iter  diff %  speedup 
 cfi::eval_longest_fde_instructions_new_ctx_everytime       532             546                      14   2.63%   x 0.97 
 cfi::eval_longest_fde_instructions_same_ctx                353             329                     -24  -6.80%   x 1.07 
 cfi::iterate_entries_and_do_not_parse_any_fde              63,038          57,991               -5,047  -8.01%   x 1.09 
 cfi::iterate_entries_and_parse_every_fde                   490,942         450,314             -40,628  -8.28%   x 1.09 
 cfi::iterate_entries_and_parse_every_fde_and_instructions  1,065,654       1,013,824           -51,830  -4.86%   x 1.05 
 cfi::iterate_entries_evaluate_every_fde                    1,376,872       1,327,763           -49,109  -3.57%   x 1.04 
 cfi::parse_longest_fde_instructions                        224             222                      -2  -0.89%   x 1.01 
```

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -561,7 +561,7 @@ mod cfi {
                     }
                     CieOrFde::Fde(partial) => {
                         let fde = partial
-                            .parse(|offset| eh_frame.cie_from_offset(&bases, offset))
+                            .parse(EhFrame::cie_from_offset)
                             .expect("Should be able to get CIE for FED");
                         test::black_box(fde);
                     }
@@ -594,7 +594,7 @@ mod cfi {
                     }
                     CieOrFde::Fde(partial) => {
                         let fde = partial
-                            .parse(|offset| eh_frame.cie_from_offset(&bases, offset))
+                            .parse(EhFrame::cie_from_offset)
                             .expect("Should be able to get CIE for FED");
                         let mut instrs = fde.instructions();
                         while let Some(i) =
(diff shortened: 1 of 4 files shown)


philipc:
Another thing to consider is the use of `Box` for `UninitializedUnwindContext`. Since we now only ever accept it by reference, and don't hold long lived references to it (only in `UnwindTable` which is normally short lived), maybe we should remove the `Box` now? Then it would be up to the caller to decide where they want to allocate it.

fitzgen:
> Another thing to consider is the use of `Box` for `UninitializedUnwindContext`. Since we now only ever accept it by reference, and don't hold long lived references to it (only in `UnwindTable` which is normally short lived), maybe we should remove the `Box` now? Then it would be up to the caller to decide where they want to allocate it.

The original motivation for the box was to avoid moving large structs around on the stack, if I recall correctly. Now that we have switched to a reference-based API instead of a move-based API, this seems like a good change to make.

Issue #405, opened 2019-03-12
Unexpected EOF when constructing unit from compilation unit header


AjayMT:
Hello all,

I would first like to thank you for your incredible work -- gimli is by far the most approachable and convenient DWARF parsing library I have found. :smile:

I've been trying to parse some Mach-O files on macOS with [object](http://github.com/gimli-rs/object) and gimli, and I've encountered an issue that I haven't found much documentation on. It is likely an error on my part, but I'm not sure exactly what's causing it. This is the relevant code:

```rust
let s_debug_info = parsed_file.section_by_name(".debug_info").unwrap().data();
let s_debug_abbrev = parsed_file.section_by_name(".debug_abbrev").unwrap().data();

let debug_info = gimli::DebugInfo::new(&s_debug_info, gimli::LittleEndian);
let debug_abbrev = gimli::DebugAbbrev::new(&s_debug_abbrev, gimli::LittleEndian);
let dwarf = gimli::Dwarf {
    debug_info,
    debug_abbrev,
    ..Default::default()
};

let units = dwarf.units().collect::<Vec<_>>().unwrap();
for header in units {
    let unit = match dwarf.unit(header) {
        Ok(r) => r,
        Err(err) => {
            println!("error constructing unit for header {:?}: {}", header, err);
            continue;
        }
    };

    println!("unit name {:?}", unit.name.unwrap());
}
```

`dwarf.unit(header)` seems to fail every time with an `UnexpectedEof` error. Perhaps this is due to how I'm loading the data with object?

I am trying to read and parse DWARF files within dSYM bundles, i.e `example.dSYM/Contents/Resources/DWARF/example`. Please let me know if this is not the right place to ask this question. Any advice is appreciated.

philipc:
The is probably because `dwarf.unit()` tries to parse the line number program too, and you haven't provided the `.debug_line` section. If there's a reason you don't want to load `.debug_line`, then I can try to rethink the error handling in `dwarf.unit()` to see if we can handle this case better.

philipc:
I see our example does the same thing, so I'll have to fix that too.

AjayMT:
Thanks for the quick response. Unfortunately, loading `debug_line` does not seem to fix the error. The dwarfdump example works perfectly, so it's definitely an issue with my code. I will try loading every DWARF section.

philipc:
[...]

PR #406, opened 2019-03-12
Add Dwarf::load and Dwarf::borrow

philipc:
Use `Dwarf::load` in the `read` module example. See #405 for motivation. If users want to avoid loading some sections then they can handle that in the loader callback.

`Dwarf::borrow` can be used when the `Reader` needs to borrow from something else. In the dwarfdump example we use an arena, but instead of that we could first load everything into a `Dwarf<Cow<[u8]>>` and then borrow that to get `Dwarf<EndianSlice<_>>`. Note that this can't use the standard rust `Borrow` trait, and I couldn't come up with a way to define this as a `Section` trait method either (do we need type constructors or something like that?). I didn't update dwarfdump to use this because the arena is still convenient for the sections that aren't in `Dwarf`.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -188,3 +188,73 @@ pub enum UnitSectionOffset<T = usize> {
     /// An offset into the `.debug_types` section.
     DebugTypesOffset(DebugTypesOffset<T>),
 }
+
+/// An identifier for a DWARF section.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SectionId {
+    /// The `.debug_abbrev` section.
+    DebugAbbrev,
+    /// The `.debug_addr` section.
+    DebugAddr,
+    /// The `.debug_aranges` section.
+    DebugAranges,
+    /// The `.debug_frame` section.
+    DebugFrame,
+    /// The `.eh_frame` section.
+    EhFrame,
+    /// The `.eh_frame_hdr` section.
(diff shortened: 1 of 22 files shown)


src/read/abbrev.rs
```diff
@@ -54,13 +54,25 @@ impl<R: Reader> DebugAbbrev<R> {
     }
 }
 
-impl<R: Reader> Section<R> for DebugAbbrev<R> {
+impl<T> DebugAbbrev<T> {
+    /// Create a `DebugAbbrev` section that references the data in `self`.
+    ///
+    /// This is useful when `R` implements `Reader` but `T` does not.
```
fitzgen: An example here would be useful because it isn't otherwise immediately clear to me when one would would need to use this.

src/common.rs
```diff
@@ -188,3 +188,73 @@ pub enum UnitSectionOffset<T = usize> {
     /// An offset into the `.debug_types` section.
     DebugTypesOffset(DebugTypesOffset<T>),
 }
+
+/// An identifier for a DWARF section.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SectionId {
+    /// The `.debug_abbrev` section.
+    DebugAbbrev,
```
fitzgen: Nice, I was going to suggest adding this, but looks like you've anticipated it already :)

PR #412, opened 2019-04-04
write: implement .debug_frame

### LINKED ISSUE #389: Encoding for CFA expressions
`Expression::evaluate` requires an `Encoding`, but unwind sections don't readily provide this. We should add something that returns the correct encoding to use.

philipc:
Also fix #389.

Hasn't had any real world testing yet.

cc @yurydelendik

### DIFF
--- a/src/write/cfi.rs
+++ b/src/write/cfi.rs
@@ -1,0 +1,909 @@
+use crate::collections::hash_map;
+use crate::vec::Vec;
+use indexmap::IndexSet;
+use std::ops::{Deref, DerefMut};
+
+use crate::collections::HashMap;
+use crate::common::{DebugFrameOffset, Encoding, Format, Register, SectionId};
+use crate::constants;
+use crate::write::{Address, BaseId, Error, Expression, Result, Section, Writer};
+
+define_section!(
+    DebugFrame,
+    DebugFrameOffset,
+    "A writable `.debug_frame` section."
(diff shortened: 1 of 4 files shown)


src/write/cfi.rs
```diff
@@ -0,0 +1,909 @@
...
+            address,
+            length,
+            lsda: None,
+            instructions: Vec::new(),
+        }
+    }
+
+    /// Add an instruction.
+    ///
+    /// Instructions must be added in increasing order of offset, or writing will fail.
```
fitzgen: If we internally stored instructions as `BTreeMap<u32, Vec<CallFrameInstruction>>` the user-facing API could be a bit nicer here. We wouldn't have to have the order of adding instructions requirement, and `write_advance_loc` would also get a little simpler.

philipc: Lots of small Vecs seems bad for performance though?

philipc: Also, I don't think the ordering requirement is problematic, because the machine code and line number program need to be emitted in order too, and the CFA definitions depend on order (although maybe we could handle that internally too).

fitzgen: Fair

yurydelendik:
I used this API at https://github.com/CraneStation/wasmtime/pull/53 -- works good. Thank you.

PR #415, opened 2019-04-15
Add a convenience getter for the address range of a DIE

bjorn3:
This is used in `addr2line` and seems useful to other projects.


philipc:
I agree that gimli should provide something like this, but I need to give more thought to the exact API we should expose. I'm not keen on the `unit` module depending on the `dwarf` module. The `dwarf` module is meant to be an optional higher level API.

bjorn3:
Maybe move it to a utils mod?

bjorn3:
Closing in favor of #417.

Issue #390, opened 2019-02-20
Write support


philipc:
Sections:
- [ ] expressions/locations in .debug_info and .debug_loc/.debug_loclists
- [ ] .debug_str_offsets
- [ ] .debug_frame/.eh_frame/.eh_frame_hdr
- [ ] .debug_aranges
- [ ] .debug_pubnames/.debug_pubtypes/.debug_names
- [ ] .debug_macinfo/.debug_macro
- [ ] .apple_names/.apple_types/.apple_namespaces

Also need general improvements:
- [ ] more mutability (#373, #374)
- [ ] indexed addresses/range lists/location lists

bjorn3:
>  .debug_frame/.eh_frame/.eh_frame_hdr

Done right?

philipc:
`.eh_frame_hdr` isn't, but I think that's the linker's job so it's a low priority.

PR #425, opened 2019-05-01
Add .debug_loc/.debug_loclists sections writing

yurydelendik:
Mostly based on src/write/range.rs.

In support of #390

### DIFF
--- a/src/write/loc.rs
+++ b/src/write/loc.rs
@@ -368,7 +368,7 @@ mod convert {
                         Location::DefaultLocation { data }
                     }
                 };
-                // In some cases, existing data may contain begin==end, filtering
+                // In some cases, existing data may contain begin == end, filtering
                 // these out.
                 match loc {
                     Location::StartLength { length, .. } if length == 0 => continue,
@@ -430,6 +430,11 @@ mod tests {
                             data: Expression(vec![3, 0, 0, 0]),
                         },
                     ]);
+                    if version >= 5 {
+                        loc_list.0.push(Location::DefaultLocation {
+                            data: Expression(vec![4, 0, 0, 0]),
+                        });
+                    }
 
                     let mut locations = LocationListTable::default();
                     let loc_list_id = locations.add(loc_list.clone());
--- a/src/write/loc.rs
+++ b/src/write/loc.rs
@@ -1,0 +1,496 @@
+use crate::vec::Vec;
+use indexmap::IndexSet;
+use std::ops::{Deref, DerefMut};
+
+use crate::common::{Encoding, LocationListsOffset, SectionId};
+use crate::write::{Address, BaseId, Error, Expression, Result, Section, Sections, Writer};
+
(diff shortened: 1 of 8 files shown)


src/write/loc.rs
```diff
@@ -0,0 +1,496 @@
...
+                        }
+                    }
+                    read::RawLocListEntry::DefaultLocation { data } => {
+                        let data = Expression(data.0.to_slice()?.into());
+                        Location::DefaultLocation { data }
+                    }
+                };
+                // In some cases, existing data may contain begin==end, filtering
+                // these out.
+                match loc {
```
yurydelendik: The range.rs has no such logic. The fixtures/self/debug_loc contains entries with such invalid data.

philipc: Looks good, maybe we should add this to range.rs too.

src/write/loc.rs
```diff
@@ -0,0 +1,496 @@
...
+        for &version in &[2, 3, 4, 5] {
+            for &address_size in &[4, 8] {
+                for &format in &[Format::Dwarf32, Format::Dwarf64] {
+                    let encoding = Encoding {
+                        format,
+                        version,
+                        address_size,
+                    };
+
+                    let mut loc_list = LocationList(vec![
```
philipc: Please add a DefaultLocation test when version is 5.

src/write/section.rs
```diff
@@ -77,6 +78,10 @@ pub struct Sections<W: Writer> {
     pub debug_ranges: DebugRanges<W>,
     /// The `.debug_rnglists` section.
     pub debug_rnglists: DebugRngLists<W>,
+    /// The `.debug_loc` section.
+    pub debug_loc: DebugLoc<W>,
+    /// The `.debug_loclists` section.
+    pub debug_loclists: DebugLocLists<W>,
```
philipc: Please also use these in `Sections::for_each/for_each_mut`.

src/write/unit.rs
```diff
@@ -735,7 +747,7 @@ pub enum AttributeValue {
     /// It is the user's responsibility to ensure the offset is valid.
     /// This variant will be removed from the API once support for writing
     /// `.debug_loc`/`.debug_loclists` sections is implemented.
-    LocationListsRef(LocationListsOffset),
+    LocationListsRef(LocationListId),
```
philipc: Please rename to `LocationListRef` and update the comment.

philipc:
Related to this, we also need a better way of handling expressions. Do you have any ideas on what the API for creating expressions should look like?

yurydelendik:
> Do you have any ideas on what the API for creating expressions should look like?

@philipc DWARF expressions will likely require different abstractions for reading/writing. It is a stack machine language, so there not much can be done in terms of IR for writing (e.g. non-AST based, perhaps Vec of ops?). As it is now, the read API is look acceptable to some degree from my point of view.

PR #427, opened 2019-05-10
Support the -G flag from the binutils dwarfdump.

khuey:


examples/dwarfdump.rs
```diff
@@ -1093,8 +1129,9 @@ fn dump_attr_value<R: Reader, W: Write>(
             let address = dwarf.address(unit, index)?;
             writeln!(w, "0x{:08x}", address)?;
         }
-        gimli::AttributeValue::UnitRef(gimli::UnitOffset(offset)) => {
-            writeln!(w, "<0x{:08x}>", offset)?;
+        gimli::AttributeValue::UnitRef(o) => {
+            write_offset(w, unit, o, flags)?;
```
philipc: I prefer that within attributes we use a format consistent with the [`AttributeValue::DebugInfoRef`](https://github.com/gimli-rs/gimli/blob/423431f6747167af8e64cc8c8378f9a34a18ff8f/examples/dwarfdump.rs#L1100) variant, rather than the `GOFF` label.

examples/dwarfdump.rs
```diff
@@ -1093,8 +1129,9 @@ fn dump_attr_value<R: Reader, W: Write>(
...
+            write!(w, "\n")?;
```
philipc:
```suggestion
            writeln!(w)?;
```

khuey:
So I'm not thrilled about *only* showing the global offset, because 99% of my use of dwarfdump is checking the raw DWARF against what I'm getting out of my gimli embedding, and in gimli the `UnitOffset` is often what's handy. I wouldn't object to reversing the optionality though (showing global offsets all the time with a flag for the unit offsets).

philipc:
That's fine, let's leave that how you have it then.

examples/dwarfdump.rs
philipc: Also, I think it would be useful to always display both here.

philipc:
Thanks!

PR #434, opened 2019-08-15
Add new GNU constants.

khuey:
gcc 9.1 is emitting these values now.


src/constants.rs
```diff
@@ -721,6 +723,7 @@ DwLle(u8) {
     DW_LLE_base_address = 0x06,
     DW_LLE_start_end = 0x07,
     DW_LLE_start_length = 0x08,
+    DW_LLE_GNU_view_pair = 0x09,
```
philipc: Will this require parsing support too?

khuey: It will.

PR #438, opened 2019-09-19
Support CFI augmentation without data

jan-auer:
In a specific ARM build of `libc.so`, we've encountered CFI augmentations that neither use the `eh` nor the `z` format, and just define a signal trampoline (`"S"`). This patch changes parsing of the `z` operand to make it optional according to:

http://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -1107,24 +1107,29 @@ impl Augmentation {
             "Augmentation::parse should only be called if we have an augmentation"
         );
 
-        let first = augmentation_str.read_u8()?;
-        if first != b'z' {
-            return Err(Error::UnknownAugmentation);
-        }
-
         let mut augmentation = Augmentation::default();
 
-        let augmentation_length = input.read_uleb128().and_then(R::Offset::from_u64)?;
-        let rest = &mut input.split(augmentation_length)?;
+        let mut parsed_first = false;
+        let mut data = None;
 
         while !augmentation_str.is_empty() {
             let ch = augmentation_str.read_u8()?;
             match ch {
+                b'z' => {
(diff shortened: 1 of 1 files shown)


jan-auer:
The test failure on Rust `1.32.0` also happens on master and is unrelated.

fitzgen:
@philipc Regarding the CI failure: looks like `flate2` doesn't support rust 1.32 anymore, which is depended upon by `object`. If we want to guarantee 1.32 support then we should pin `object`'s dependency to an older version, I guess.

```
   Compiling flate2 v1.0.11
error[E0658]: use of unstable library feature 'try_from' (see issue #33417)
  --> /Users/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/flate2-1.0.11/src/ffi.rs:47:9
   |
47 |     use std::convert::TryFrom;
   |         ^^^^^^^^^^^^^^^^^^^^^
error[E0658]: use of unstable library feature 'try_from' (see issue #33417)
   --> /Users/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/flate2-1.0.11/src/ffi.rs:109:27
    |
109 |             .and_then(|i| usize::try_from(i).ok())
    |                           ^^^^^^^^^^^^^^^
```

fitzgen:
Going to go ahead and merge this, since CI failure is unrelated to this PR.

philipc:
> If we want to guarantee 1.32 support then we should pin object's dependency to an older version, I guess.

No, we can bump that version whenever it is reasonable to do so. That test is more for simply knowing what the minimum version is.

PR #442, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

LukasKalbertodt:
`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future 
change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -5477,7 +5477,7 @@ mod tests {
             (Register(0), RegisterRule::Offset(8)),
             (Register(3), RegisterRule::Offset(4)),
         ]
-        .into_iter()
+        .iter()
         .collect();
         assert_eq!(table.ctx.initial_rules, expected_initial_rules);
 
@@ -5495,7 +5495,7 @@ mod tests {
                     (Register(0), RegisterRule::Offset(8)),
                     (Register(3), RegisterRule::Offset(4)),
                 ]
-                .into_iter()
+                .iter()
                 .collect(),
             };
             assert_eq!(Some(&expected), row);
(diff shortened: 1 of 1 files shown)


LukasKalbertodt:
@philipc Do you think it would be possible to release this fix as 0.19.1? Many crates depend on this one and getting the fix out (particularly, as minor version bump) would be really nie.

philipc:
Is this to make it easier to see which other crates require fixes? My impression was that this was too large a breaking change for there to be any urgency.

LukasKalbertodt:
You are right in that the Rust PR won't be merged anytime soon. But there is still a decent chance that this change is introduced as is, so fixing exiting crates is still an important part of that. My reasoning for releasing soon goes like this:

- If it's possible to release as a minor version bump, then one should do that soon, as later changes could require a major bump (which is way less desirable). That would complicate the release process later on.
- If it's not easily possible to release as minor version bump, then releasing a new major version as soon as possible is good, because downstream crates then have more time to upgrade to the new version. 

If releasing this would require notably more work than a single `cargo publish` on your side, of course feel free to just release at your own convenience. I just wanted to ask and don't want to make other people put in a lot of work for this :)

LukasKalbertodt:
@philipc Ignore everything I said before! 

This PR only fixes tests. Tests are not compiled for crates that depend on this crate. So indeed, `gimli` did not cause a single other regression. I was mistaken and I'm sorry to have asked for a release when it wasn't worth anything. So yeah, nothing needs to be done here. Again, sorry!

philipc:
Ok no worries, thanks for looking into it.

PR #449, opened 2019-11-16
Fix dwarfdump machO relocations

bjorn3:


philipc:
Not sure about this. The relocations provided by `object` are meant to abstract these differences, and if that's not happening then I need to understand why. The bug may actually be that we need to update to a new version of `object` which has the [handling for the extern flag in mach-o relocations](https://github.com/gimli-rs/object/commit/e9faee995f9821c7712a6578a221ec17d35a3f44#diff-af6f3439b3f4a5caf6f342e6e1a9d041R548).

bjorn3:
I ave updated this to use object 0.15 instead.

bjorn3:
Forgot about your review.

philipc:
No worries, and thanks for the PR!

PR #454, opened 2019-11-21
read: handle more attributes names in Attribute::value()

philipc:
I think the only change of significance is the handling for `DW_AT_macros` from DWARF 5.

### DIFF
--- a/src/write/unit.rs
+++ b/src/write/unit.rs
@@ -3,8 +3,8 @@ use std::ops::{Deref, DerefMut};
 use std::{slice, usize};
 
 use crate::common::{
-    DebugAbbrevOffset, DebugInfoOffset, DebugLineOffset, DebugMacinfoOffset, DebugStrOffset,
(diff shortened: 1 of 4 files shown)


src/write/unit.rs
```diff
@@ -752,6 +752,13 @@ pub enum AttributeValue {
     /// `.debug_macinfo` sections is implemented.
     DebugMacinfoRef(DebugMacinfoOffset),
 
+    /// An offset into the `.debug_macro` section.
+    ///
+    /// It is the user's responsibility to ensure the offset is valid.
```
fitzgen: Maybe worth clarifying that *correctness* depends on this, but not *safety*

fitzgen: Maybe something like this?

```suggestion
    /// For correctness (but not safety) it is the user's responsibility to ensure the offset is valid.
```

philipc: I wasn't worried so much about correctness as conveying that the user was on their own here. I've updated all of these comments (this was a copy+paste from existing comments).

PR #455, opened 2019-11-25
 read: add `EntriesRaw`

philipc:
Provides a raw reader of Debugging Information Entries. This is useful for users who require performance and don't need the niceties of `EntriesCursor`. Comparative benchmarks:

```
test bench_parsing_debug_info         ... bench:   1,733,047 ns/iter (+/- 63,346)
test bench_parsing_debug_info_raw     ... bench:   1,182,351 ns/iter (+/- 36,013)
test bench_parsing_debug_info_tree    ... bench:   1,836,409 ns/iter (+/- 57,726)
```

(Note: if you run the benchmarks yourself you'll see that `bench_parsing_debug_info_with_endian_rc_slice` is faster; this is due to differences in inlining.)

These benchmarks don't do anything with the attributes, so in practical applications the gains aren't this much, but they are still significant (I have tested in addr2line).

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -134,6 +134,40 @@ fn parse_debug_info_tree<R: Reader>(node: EntriesTreeNode<R>) {
 }
 
 #[bench]
+fn bench_parsing_debug_info_raw(b: &mut test::Bencher) {
+    let debug_abbrev = read_section("debug_abbrev");
+    let debug_abbrev = DebugAbbrev::new(&debug_abbrev, LittleEndian);
+
+    let debug_info = read_section("debug_info");
+
+    b.iter(|| {
+        let debug_info = DebugInfo::new(&debug_info, LittleEndian);
+
+        let mut iter = debug_info.units();
+        while let Some(unit) = iter.next().expect("Should parse compilation unit") {
+            let abbrevs = unit
+                .abbreviations(&debug_abbrev)
+                .expect("Should parse abbreviations");
+
+            let mut raw = unit
+                .entries_raw(&abbrevs, None)
+                .expect("Should have entries");
+            while !raw.is_empty() {
+                if let Some(abbrev) = raw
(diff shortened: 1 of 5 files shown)


philipc:
I probably should add a test too... tomorrow.

benches/bench.rs
```diff
@@ -62,8 +62,16 @@ fn impl_bench_parsing_debug_info<R: Reader>(
         let mut cursor = unit.entries(&abbrevs);
         while let Some((_, entry)) = cursor.next_dfs().expect("Should parse next dfs") {
             let mut attrs = entry.attrs();
-            while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
-                test::black_box(&attr);
+            loop {
+                match attrs.next() {
```
fitzgen: What's the point of going from `while let Some` to `loop` + `match`?

src/read/unit.rs
```diff
@@ -2279,6 +2306,115 @@ impl<'abbrev, 'entry, 'unit, R: Reader> FallibleIterator for AttrsIter<'abbrev,
...
 }
 
+/// A raw reader of the data that defines the Debugging Information Entries.
+///
+/// This reader provides primitives to read the components of Debugging Information
+/// Entries (DIEs). A DIE consists of an abbreviation code (read with `read_abbreviation`)
+/// followed by a number of attributes (read with `read_attribute`).
+/// The user must provide the control flow to read these correctly.
+/// In particular, all attributes must always be read before reading another
+/// abbreviation code.
```
fitzgen: It would be nice to add a section about "why would I use the raw interface instead of the cursor/tree?" to these docs.

benches/bench.rs
philipc: It avoids a move of the attribute which the compiler fails to optimise away. But maybe this requires more explanation and deserves its own PR.

PR #456, opened 2019-11-27
bench: avoid moving attributes

philipc:
```
 name                                           before ns/iter  after ns/iter  diff ns/iter   diff %  speedup
 bench_parsing_debug_info                       1,854,403       1,730,007          -124,396   -6.71%   x 1.07
 bench_parsing_debug_info_raw                   1,374,276       1,187,147          -187,129  -13.62%   x 1.16
 bench_parsing_debug_info_tree                  1,926,297       1,821,076          -105,221   -5.46%   x 1.06
 bench_parsing_debug_info_with_endian_rc_slice  1,919,689       1,637,859          -281,830  -14.68%   x 1.17
```

For `bench_parsing_debug_info_raw`, the main difference in code is:
```
- 		37b:  call   0x2b9a0
- 		             0x2b9a0 = gimli::read::unit::parse_attribute<gimli::read::endian_slice::EndianSlice<gimli::endianity::LittleEndian>>
- 		380:  cmp    qword ptr [rsp + 0x98], 1
- 		389:  je     +3fe
- 		38b:  movups xmm0, xmmword ptr [r12]
- 		390:  movups xmm1, xmmword ptr [r12 + 0x10]
- 		396:  movaps xmmword ptr [rsp + 0xe0], xmm1
- 		39e:  movaps xmmword ptr [rsp + 0xd0], xmm0
- 		3a6:  mov    qword ptr [rsp + 0x98], rbx
- 		3ae:  add    rbp, 0x10
- 		3b2:  cmp    rbp, r13
- 		3b5:  jne    +360

+ 		36b:  call   0x2b960
+ 		             0x2b960 = gimli::read::unit::parse_attribute<gimli::read::endian_slice::EndianSlice<gimli::endianity::LittleEndian>>
+ 		370:  cmp    qword ptr [rsp + 0xf8], 1
+ 		379:  je     +3d3
+ 		37b:  mov    qword ptr [rsp + 0x50], rbx
+ 		380:  add    rbp, 0x10
+ 		384:  cmp    rbp, r13
+ 		387:  jne    +350
```

For the other benchmarks I didn't find any clear cause for the difference, but there is more inlining going on there which makes it harder to see. For these, changing `while let Some(attr)` to `while let Some(ref attr)` gave some improvement, but not as much as the change in this commit.

Doing a similar change in addr2line also helps the performance there, and I believe this sort of overhead is partly why inlining `parse_attribute` helps there (see https://github.com/gimli-rs/addr2line/pull/148).

I'm not happy about doing this, but I don't know any way to fix it. I had thought it might be related to unwinding, but adding `#[unwind(aborts)]` on `parse_attribute` made no difference.

I'm also curious whether this performance difference is a quirk of my older cpu (i5-4670), because I wasn't expecting a couple of moves to make this much difference, so if someone else could replicate these benchmark results then that would be great.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -62,8 +62,16 @@ fn impl_bench_parsing_debug_info<R: Reader>(
         let mut cursor = unit.entries(&abbrevs);
         while let Some((_, entry)) = cursor.next_dfs().expect("Should parse next dfs") {
             let mut attrs = entry.attrs();
-            while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
-                test::black_box(&attr);
+            loop {
+                match attrs.next() {
+                    Ok(Some(ref attr)) => {
+                        test::black_box(attr);
+                    }
+                    Ok(None) => break,
+                    e @ Err(_) => {
+                        e.expect("Should parse entry's attribute");
+                    }
+                }
             }
         }
     }
@@ -123,8 +131,16 @@ fn bench_parsing_debug_info_tree(b: &mut test::Bencher) {
 fn parse_debug_info_tree<R: Reader>(node: EntriesTreeNode<R>) {
     {
         let mut attrs = node.entry().attrs();
-        while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
-            test::black_box(&attr);
+        loop {
+            match attrs.next() {
+                Ok(Some(ref attr)) => {
+                    test::black_box(attr);
+                }
+                Ok(None) => break,
+                e @ Err(_) => {
+                    e.expect("Should parse entry's attribute");
+                }
+            }
         }
     }
     let mut children = node.children();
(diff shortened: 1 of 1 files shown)


benches/bench.rs
```diff
@@ -62,8 +62,16 @@ fn impl_bench_parsing_debug_info<R: Reader>(
         let mut cursor = unit.entries(&abbrevs);
         while let Some((_, entry)) = cursor.next_dfs().expect("Should parse next dfs") {
             let mut attrs = entry.attrs();
-            while let Some(attr) = attrs.next().expect("Should parse entry's attribute") {
```
bjorn3: Would using `while let Sone(ref attr)` or `let mut attrs = entry.attrs().by_ref()` work?

philipc: I mentioned `while let Some(ref attr) ` in the PR text. I'll try `by_ref` later, but it seems to be for creating a reference to the iterator, not the items.

bjorn3: Should have read the PR description better.

philipc: `by_ref` made no difference.

PR #460, opened 2019-12-09
Add simple_line example

philipc:
cc @eqv


eqv:
I know this is pretty late, but thanks a lot for this example. It really helped me a lot!

PR #462, opened 2019-12-22
Remove deprecated Error::description

AnderEnder:
`Error::description` has been documented as soft-deprecated since 1.27.0 (17 months ago). It is going to be hard-deprecated soon.

This PR:
- Removes the implementation of `description` in error type

Related PR: https://github.com/rust-lang/rust/pull/66919

### DIFF
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -384,7 +384,7 @@ pub enum Error {
 impl fmt::Display for Error {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> ::core::result::Result<(), fmt::Error> {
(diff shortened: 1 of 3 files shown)


src/read/mod.rs
```diff
@@ -384,7 +384,7 @@ pub enum Error {
 impl fmt::Display for Error {
     #[inline]
     fn fmt(&self, f: &mut fmt::Formatter) -> ::core::result::Result<(), fmt::Error> {
-        Debug::fmt(self, f)
+        write!(f, "{}", self.description())
```
philipc: This is a user-visible change. I think this is fine for now, but I will file a follow-up issue.

PR #466, opened 2020-01-09
Update dependencies and make more optional

philipc:


Cargo.toml
```diff
@@ -17,28 +17,27 @@ travis-ci = { repository = "gimli-rs/gimli" }
...
-typed-arena = "1"
-target-lexicon = "0.8"
+typed-arena = "2"
 
 [features]
-read = ["stable_deref_trait"]
-write = ["std", "indexmap"]
+read = ["arrayvec", "fallible-iterator", "smallvec", "stable_deref_trait"]
+write = ["indexmap"]
 std = ["fallible-iterator/std", "stable_deref_trait/std"]
```
bjorn3: I think this will cause `fallible-iterators` to be enabled when `std` is enabled but `read` is disabled.

src/lib.rs
```diff
@@ -40,7 +40,7 @@
 #[macro_use]
 extern crate alloc;
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "write"))]
```
bjorn3: Why not make `write` depend on `std` instead?

philipc: Because I couldn't figure out how to only enable some dependencies if both `read` and `std` are enabled.

Cargo.toml
philipc: Yes, as in the other comment, I don't know how to enable `fallible-iterators/std` only if both `read` and `std` are enabled. So my workaround is to not enable `std` for `write`.

PR #476, opened 2020-03-27
write/line: avoid division if minimum instruction length is 1

### LINKED ISSUE #468: [LineProgram] Don't make LineEncoding configurable
This would likely make it possible for LLVM to optimize a division away which accounts for ~30% of the time in `generate_row`. I don't see much use for it anyway.

https://github.com/gimli-rs/gimli/blob/1e49ffc9af4ec64a1b7316924d73c933dd7157c5/src/write/line.rs#L471

and

https://github.com/gimli-rs/gimli/blob/1e49ffc9af4ec64a1b7316924d73c933dd7157c5/src/write/line.rs#L475

philipc:
Fixes #468 

I haven't done any performance measurements, but a similar change was significant when reading.
cc @bjorn3

### DIFF
--- a/src/write/line.rs
+++ b/src/write/line.rs
@@ -467,12 +467,14 @@ impl LineProgram {
 
     fn op_advance(&self) -> u64 {
         debug_assert!(self.row.address_offset >= self.prev_row.address_offset);
-        debug_assert_eq!(
-            self.row.address_offset % u64::from(self.line_encoding.minimum_instruction_length),
-            0
-        );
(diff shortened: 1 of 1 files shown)


bjorn3:
This reduces the time of `LineProgram::generate_row` into which `op_advance` is inlined from 0.55% (0.44% self) to 0.24% (0.13% self) when generating the debuginfo for the `image` crate. For comparison the total amount of time spent on debuginfo generation for that crate with this patch is 1.46% of the total compilation time.

Issue #475, opened 2020-03-26
[DWARF-5] Ambiguity when reading DW_FORM_sec_offset


ggreif:
DWARF-5 spec mentions (in 7.5.5 Classes and Forms, p. 212, l. 6) that "`DW_FORM_sec_offset` is a member of more than one class". It can format the attributes `DW_AT_addr_base`, `DW_AT_stmt_list` and `DW_AT_ranges`, for example.

However when reading with `gimli`, it always seems to get resolved as `AttributeValue::DebugAddrBase`, regardless of the `DW_AT_*`.

philipc:
It isn't always `AttributeValue::DebugAddrBase`. We have [tests for some of these cases](https://github.com/gimli-rs/gimli/blob/15bea8e39208dcb045ed20600fb8f9ee8d8b2b72/src/read/unit.rs#L4036-L4071). From reading the code `DW_AT_stmt_list` and `DW_AT_ranges` should be handled correctly and since these are such common attributes I would be surprised if they are wrong. Can you be more specific about which `DW_AT` value it is getting wrong or provide a test case?

ggreif:
@philipc I am seeing `AttributeValue::DebugAddrBase` arriving in `wasmtime`. It originates from
`DW_AT_stmt_list` with `DW_FORM_sec_offset`. So it might be an issue with `wasmtime` or me (generating bad DWARF?). Thanks for your comment, I'll look into it more closely and come back.

philipc:
> This is a piece of code I came across: https://github.com/gimli-rs/gimli/blob/master/src/read/unit.rs#L1231

That code is in a macro. You need to look where the macro used, and it is only used for [`DW_AT_addr_base`](https://github.com/gimli-rs/gimli/blob/15bea8e39208dcb045ed20600fb8f9ee8d8b2b72/src/read/unit.rs#L1699-L1701).

philipc:
I had a look a the wasmtime PR, and I think your check for zero address base is wrong.

From 7.27 of the standard:

> The DW_AT_addr_base attribute points to the first entry following the header. The entries are indexed sequentially from this base entry, starting from 0.

Since this is pointing past the header, it won't be zero. Note that originally some compilers emitted .debug_addr without the header, so it may be zero in that case.

ggreif:
> I had a look a the wasmtime PR, and I think your check for zero address base is wrong.

[...]

PR #477, opened 2020-03-27
test reading of DW_AT_stmt_list as DW_FORM_sec_offset

ggreif:
I thought this would be a good test to have while tracking down #475 on my side.


ggreif:
I thought this would be a good test to have while tracking down #475 on my side.

Issue #478, opened 2020-03-30
write: need better handling of DW_FORM_ref_addr


philipc:
The current handling of `DW_FORM_ref_addr` assumes that you are writing all of the compilation units at once, so that it can calculate the section offset itself, but that is atypical behaviour. Instead, the producer will normally need to generate a symbol and relocation so that the reference can be resolved during static linking.

Fixing this will need a means to assign a symbol to the reference, and a means to assign a symbol to an entry that may be referred to.

This will also apply to `DW_FORM_call_ref` in expressions.

PR #479, opened 2020-03-30
write: implement references within expressions

### LINKED ISSUE #474: gimli::write: support relocation in expression
I am generating DWARF on fly with the object file. I want to specify `DW_AT_location` for a global variable, which would be a constant address in the .data section and would require relocation (since I am also generating the .data section before linking and don't know the final address yet). Unlike other places that needs relocation and that supports `gimli::write::Address::Symbol`, `DW_AT_location` expects an expression and I can't find a sane way to add the relocation. Here is what I have tried:
 - Use `DW_OP_addr` followed by the address. The address needs relocation, but I would need to know where the "from" point to when writing the relocation entry, which gimli doesn't give.
 - Use `DW_OP_addrx` followed by an index into `.debug_addr` section, and do relocation there. However, 
   - `gimli::write` doesn't support `.debug_addr`. Not a big deal as I can manually craft one, but
   - I need to set `DW_AT_addr_base` to the unit, which expects a section offset into `.debug_addr` (`addrptr` class) but `gimli::write::AttributeValue` doesn't support it.
   - This thing seems to be new in DWARF 5 and I wanted a lower version.

Any idea how I can do the relocation and/or can you implement the feature to support it?

philipc:
Changes `write::Expression` from `Vec<u8>` to `Vec<Operation>`, and correctly handles references in the operations. The main changes of this PR are in the last commit. The other commits are mostly refactoring to make it easier to implement.

Fixes #474 
Partial support for #478 (still need a way to assign symbols to DIEs)

@wwylele ~This is untested and still lacking a nicer API for adding operations. I'm unsure when I'll finish it off, but try it out if you want.~

### DIFF
--- a/src/write/loc.rs
+++ b/src/write/loc.rs
@@ -396,6 +396,7 @@ mod tests {
         ConvertUnitContext, EndianVec, LineStringTable, RangeListTable, StringTable,
     };
     use crate::LittleEndian;
+    use std::collections::HashMap;
 
     #[test]
     fn test_loc_list() {
@@ -484,6 +485,7 @@ mod tests {
                         base_address: Address::Constant(0),
                         line_program_offset: None,
                         line_program_files: Vec::new(),
+                        entry_ids: &HashMap::new(),
                     };
                     let convert_loc_list = LocationList::from(read_loc_list, &context).unwrap();
 
--- a/src/write/loc.rs
+++ b/src/write/loc.rs
@@ -3,7 +3,10 @@ use indexmap::IndexSet;
 use std::ops::{Deref, DerefMut};
 
 use crate::common::{Encoding, LocationListsOffset, SectionId};
-use crate::write::{Address, BaseId, Error, Expression, Result, Section, Sections, Writer};
+use crate::write::{
+    Address, BaseId, DebugInfoReference, Error, Expression, Result, Section, Sections, UnitOffsets,
+    Writer,
+};
 
 define_section!(
     DebugLoc,
@@ -45,14 +48,25 @@ impl LocationListTable {
         &self,
         sections: &mut Sections<W>,
         encoding: Encoding,
+        unit_offsets: Option<&UnitOffsets>,
     ) -> Result<LocationListOffsets> {
         if self.locations.is_empty() {
             return Ok(LocationListOffsets::none());
         }
 
         match encoding.version {
-            2..=4 => self.write_loc(&mut sections.debug_loc, encoding.address_size),
-            5 => self.write_loclists(&mut sections.debug_loclists, encoding),
+            2..=4 => self.write_loc(
+                &mut sections.debug_loc,
+                &mut sections.debug_loc_refs,
+                encoding,
+                unit_offsets,
+            ),
+            5 => self.write_loclists(
+                &mut sections.debug_loclists,
+                &mut sections.debug_loclists_refs,
+                encoding,
+                unit_offsets,
+            ),
             _ => Err(Error::UnsupportedVersion(encoding.version)),
         }
     }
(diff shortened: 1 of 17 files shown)


wwylele:
I tried it and it worked pretty well :+1:

src/write/section.rs
```diff
@@ -84,6 +85,12 @@ pub struct Sections<W: Writer> {
     pub debug_loclists: DebugLocLists<W>,
     /// The `.debug_str` section.
     pub debug_str: DebugStr<W>,
+    /// Unresolved references in the `.debug_info` section.
+    pub(crate) debug_info_refs: Vec<DebugInfoReference>,
+    /// Unresolved references in the `.debug_loc` section.
+    pub(crate) debug_loc_refs: Vec<DebugInfoReference>,
+    /// Unresolved references in the `.debug_loclists` section.
+    pub(crate) debug_loclists_refs: Vec<DebugInfoReference>,
```
philipc, line 93: This is a bit ad-hoc. Maybe we should move to something a bit more generic (closer to a traditional assembler).

src/write/loc.rs
```diff
@@ -265,6 +277,23 @@ pub enum Location {
...
+fn write_expression<W: Writer>(
+    w: &mut W,
+    refs: &mut Vec<DebugInfoReference>,
+    encoding: Encoding,
+    unit_offsets: Option<&UnitOffsets>,
+    val: &Expression,
+) -> Result<()> {
+    let size = val.size(encoding, unit_offsets);
+    if encoding.version <= 4 {
+        w.write_u16(size as u16)?;
```
bjorn3, line 289:
```suggestion
        w.write_u16(u16::try_from(size)?)?;
```

To prevent silent truncation when it is bigger than `65_535` bytes.

src/write/unit.rs
```diff
@@ -341,26 +371,40 @@ impl Unit {
...
+        for entry in &root.children {
+            if self.entries[entry.index].tag == constants::DW_TAG_base_type {
+                root_children.push(*entry);
+            }
+        }
+        for entry in &root.children {
+            if self.entries[entry.index].tag != constants::DW_TAG_base_type {
+                root_children.push(*entry);
+            }
+        }
```
bjorn3, line 405: You can use [`Iterator::partition`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.partition).

src/write/op.rs
```diff
@@ -0,0 +1,1558 @@
...
+    /// Add a `DW_OP_bit_piece` operation to the expression.
+    pub fn op_bit_piece(&mut self, size_in_bits: u64, bit_offset: u64) {
+        self.operations.push(Operation::BitPiece {
+            size_in_bits,
+            bit_offset,
+        });
+    }
+
+    /// Add a `DW_OP_GNU_parameter_ref` operation to the expression.
+    pub fn op_parameter_ref(&mut self, entry: UnitEntryId) {
```
bjorn3, line 239:
```suggestion
    pub fn op_gnu_parameter_ref(&mut self, entry: UnitEntryId) {
```

src/write/op.rs
```diff
@@ -0,0 +1,1558 @@
...
+
+    /// Add a `DW_OP_bit_piece` operation to the expression.
+    pub fn op_bit_piece(&mut self, size_in_bits: u64, bit_offset: u64) {
+        self.operations.push(Operation::BitPiece {
+            size_in_bits,
+            bit_offset,
+        });
+    }
+
+    /// Add a `DW_OP_GNU_parameter_ref` operation to the expression.
```
bjorn3, line 238: It would be nice to have some documentation about what those operations actually do as part of these doc comments.

src/write/unit.rs
philipc: That's not quite the same because I want to reorder them, not create two collections. I could then combine the collections, but I prefer to leave it how it is.

src/write/op.rs
philipc: I think that should be documented on the operations instead (and this could link to that). I started doing that, but it ends up just copying the standard. While that might be useful, it is a low priority for me, since users should be referring to the standard anyway.

src/write/op.rs
philipc: Ok, but that means the method will be changed when this becomes a standard operation. Note that other methods also add GNU operations for older DWARF versions.

PR #481, opened 2020-03-31
dwarddump update w.r.t object 0.18.0

vaibspider:
Updating the dwarfdump example to be consistent with the latest version of object - 0.18.0


examples/dwarfdump.rs
```diff
@@ -161,16 +161,18 @@ fn add_relocations(
...
-                                offset
-                            );
+                        let symbol_result = file.symbol_by_index(symbol_idx);
+                        if let Err(_some_error) = symbol_result {
+                          println!(
+                              "Relocation with invalid symbol for section {} at offset 0x{:08x}",
+                              section.name().unwrap(),
+                              offset
+                          );
+                          process::exit(1);
```
bjorn3: It previously continued on errors.

vaibspider: Removed process::exit()

examples/dwarfdump.rs
```diff
@@ -161,16 +161,17 @@ fn add_relocations(
...
-                            );
+                        let symbol_result = file.symbol_by_index(symbol_idx);
+                        if let Err(_some_error) = symbol_result {
+                          println!(
+                              "Relocation with invalid symbol for section {} at offset 0x{:08x}",
+                              section.name().unwrap(),
+                              offset
+                          );
                         }
+                        let symbol = symbol_result.unwrap();
```
bjorn3: This unwrap would panic when there was an error.

examples/dwarfdump.rs
```diff
@@ -161,16 +161,17 @@ fn add_relocations(
...
+                        let addend = symbol.address().wrapping_add(relocation.addend() as u64);
+                        relocation.set_addend(addend as i64);
```
bjorn3:
```suggestion
                        match file.symbol_by_index(symbol_idx) {
                            Ok(symbol) => {
                                let addend = symbol.address().wrapping_add(relocation.addend() as u64);
                                relocation.set_addend(addend as i64);
                            }
                            Err(_) => {
                                println!(
                                    "Relocation with invalid symbol for section {} at offset 0x{:08x}",
                                    section.name().unwrap(),
                                    offset
                                );
                            }
                        }
```

vaibspider: Sorry, I had committed and pushed (for previous review) before reading this review. So couldn't use the commit-suggestion directly. Have created a new commit, with the suggested code-refactoring.
Thanks.

examples/dwarfdump.rs
```diff
@@ -161,15 +161,18 @@ fn add_relocations(
...
-                            let addend = symbol.address().wrapping_add(relocation.addend() as u64);
-                            relocation.set_addend(addend as i64);
-                        } else {
-                            println!(
-                                "Relocation with invalid symbol for section {} at offset 0x{:08x}",
-                                section.name().unwrap(),
-                                offset
-                            );
+                        match file.symbol_by_index(symbol_idx) {
+                          Ok(symbol) => {
```
bjorn3: This uses two spaces as indent instead of four like the rest of this file.

examples/dwarfdump.rs
vaibspider: updated

examples/dwarfdump.rs
vaibspider: Ah, right. Fixed it.

examples/dwarfdump.rs
```diff
@@ -491,7 +494,7 @@ where
         let data = match file.section_by_name(id.name()) {
             Some(ref section) => {
                 add_relocations(&mut relocations, file, section);
-                section.uncompressed_data()
+                section.uncompressed_data().unwrap()
```
philipc: This should return an error instead of unwrapping.

vaibspider: Now returning the error, please check

vaibspider:
Observed that the build failed due to these two reasons:-
1) error[E0599]: no method named `section_data_by_name` found for reference `&object::read::any::File<'_>` in the current scope in many examples/ (simple.rs, simple_line.rs, dwarf-validate.rs)
2) error[E0658]: use of unstable library feature 'ptr_cast' in object 0.18.0

philipc:
It's fine to bump the rust version in `.travis.yml` and `README.md` to 1.38.0.

vaibspider:
Updated

Issue #485, opened 2020-04-08
Some CFI are skipped because of zeros in the entry


calixteman:
The `libxul.so` (firefox library) contains a `.debug_frame` with starting with zeros and the entries are just skipped because of an error.
I extracted the `.debug_frame` using `objcopy`: [libxul.sub.so.zip](https://github.com/gimli-rs/gimli/files/4449655/libxul.sub.so.zip).

PR #486, opened 2020-04-08
[CFI] Skip zeros when entry length is zero

calixteman:
Use the same trick as in readelf:
https://github.com/bminor/binutils-gdb/blob/master/binutils/dwarf.c#L7820

It aims to fix issue #485.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -1016,12 +1016,23 @@ where
     R: Reader,
     Section: UnwindSection<R>,
 {
-    let offset = input.offset_from(section.section());
-    let (length, format) = input.read_initial_length()?;
+    let (offset, length, format) = loop {
+        let offset = input.offset_from(section.section());
+        let (length, format) = input.read_initial_length()?;
 
-    if Section::length_value_is_end_of_entries(length) {
-        return Ok(None);
-    }
+        if Section::length_value_is_end_of_entries(length) {
+            return Ok(None);
+        }
+
+        // Hack: skip zero padding inserted by buggy compilers/linkers.
+        // We require that the padding is a multiple of 32-bits, otherwise
+        // there is no reliable way to determine when the padding ends. This
+        // should be okay since CFI entries must be aligned to the address size.
+
+        if length.into_u64() != 0 || format != Format::Dwarf32 {
+            break (offset, length, format);
+        }
+    };
 
     let mut rest = input.split(length)?;
     let cie_offset_base = rest.offset_from(section.section());


src/read/cfi.rs
```diff
@@ -1016,12 +1016,22 @@ where
...
+            return Ok(None);
+        }
+
+        if length.into_u64() != 0 {
+            break (offset, length, format);
+        }
+
+        // Same behaviour as in readelf
+        // https://github.com/bminor/binutils-gdb/blob/master/binutils/dwarf.c#L7820
+        input.skip_null()?;
```
philipc, line 1033: It seems this can only work correctly if the real length doesn't start with a zero?

calixteman: If the length start with a zero, then `read_initial_length()` will return a format Dwarf32.
So adding ` || format != Format::Dwarf32` in the condition to break shoud be ok.

philipc: I mean if the real length is [0x00, 0x10, 0x00, 0x00], then this will skip the first 0x00 and try to read the length starting at the 0x10. Or maybe I am misunderstanding what this is doing. I'll have a look at the sample file tomorrow.

philipc: I'm pretty sure my understanding is correct. I think the simplest fix is to delete this line so we loop around to try reading a length again, since CFI entries must be aligned to the address size. This also avoids the need to add the `skip_null` method.

src/read/cfi.rs
```diff
@@ -1016,12 +1016,22 @@ where
...
+        let offset = input.offset_from(section.section());
+        let (length, format) = input.read_initial_length()?;
 
-    if Section::length_value_is_end_of_entries(length) {
-        return Ok(None);
-    }
+        if Section::length_value_is_end_of_entries(length) {
+            return Ok(None);
+        }
+
```
philipc, line 1026: Instead of the comment on skip_null, add this:
```suggestion
        // Hack: skip zero padding inserted by buggy compilers/linkers.
        // We require that the padding is a multiple of the 32-bits, otherwise
        // there is no reliable way to determine when the padding ends.  This
        // should be okay since CFI entries must be aligned to the address size.

```

philipc:
It'd be nice to figure out which compiler/linker is causing this and fix them too. Note that llvm-dwarfdump doesn't handle this kind of padding either.

Also please squash the commits.

calixteman:
> It'd be nice to figure out which compiler/linker is causing this and fix them too. Note that llvm-dwarfdump doesn't handle this kind of padding either.
> 
I filed a bug for that:
https://bugzilla.mozilla.org/show_bug.cgi?id=1628673

src/read/cfi.rs
```diff
@@ -1016,12 +1016,23 @@ where
...
+        // Hack: skip zero padding inserted by buggy compilers/linkers.
```
glandium, line 1027: Technically speaking, skipping the zeros 4 by 4 works, but I don't think it's the right thing to do. What you actually have is not padding. What you have is multiple empty CIEs one after the other.

philipc: Technically they are zero terminators, not zero padding or empty CIEs. And zero terminators are a `.eh_frame` thing, not a `.debug_frame` thing so the bug is they shouldn't occur here. But why do you think that skipping them isn't the right thing to do? What should we do? Skipping them in `.eh_frame` would definitely be wrong, but I think it's okay to do it for `.debug_frame`?

philipc: From the mozilla bug:
> That's unfortunate, but not invalid. They "just" indicate empty records.

Where in the DWARF spec does it say these are valid?

philipc:
I filed a nasm bug: https://bugzilla.nasm.us/show_bug.cgi?id=3392657

src/read/cfi.rs
glandium: I'd ask the opposite question: where in the DWARF spec does it they they are invalid? There is no obvious mention of CIE length not being valid if it's 0. The spec doesn't say all fields must be there, and in fact, they can't since different versions of DWARF have different number of fields.

philipc: Fair enough, but different versions must at least have a version field, and a length of 0 doesn't allow for that.

Also there is no point to have a zero length:
- it provides no benefit over just omitting it (not even for padding, because padding should be including in the CFI entry length and use `DW_CFA_nop` instead)
- `.eh_frame` handling requires you to stop processing after a zero length, so it's not even compatible with that

And given that multiple DWARF consumers don't handle it correctly, it's not at all obvious that it is valid. (You can add ubuntu's dwarfdump to the list of consumers that don't handle it.)

> I don't think it's the right thing to do

What do you think should be done instead?

PR #487, opened 2020-04-16
Increase size of RegisterRuleMap to avoid TooManyRegisterRules

jan-auer:
Increases the size of `RegisterRuleMap` to fit more rules in rare cases.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -2305,9 +2305,21 @@ impl<'a, R: Reader> UnwindTable<'a, R> {
 // `RegisterRule::Undefined`, we never store a register's rule in this vec if it
 // is undefined and save a little bit more space and do a little fewer
 // comparisons that way.
+//
+// The maximum number of rules preallocated by libunwind is 97 for AArch64, 128
+// for ARM, and even 188 for MIPS. It is extremely unlikely to encounter this
+// many register rules in practice.
+//
+// See:
+// - https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-x86_64/dwarf-config.h#L36
+// - https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-aarch64/dwarf-config.h#L32
+// - https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-arm/dwarf-config.h#L31
+// - https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-mips/dwarf-config.h#L31
+//
+// TODO: Consider using const generics for the array size.
 #[derive(Clone, Debug)]
 struct RegisterRuleMap<R: Reader> {
-    rules: ArrayVec<[(Register, RegisterRule<R>); 32]>,
+    rules: ArrayVec<[(Register, RegisterRule<R>); 192]>,
 }
 
 impl<R: Reader> Default for RegisterRuleMap<R> {
@@ -6659,7 +6671,7 @@ mod tests {
     fn size_of_unwind_ctx() {
         use core::mem;
         let size = mem::size_of::<UnwindContext<EndianSlice<NativeEndian>>>();
-        let max_size = 5416;
+        let max_size = 30968;
         if size > max_size {
             assert_eq!(size, max_size);
         }
@@ -6670,7 +6682,7 @@ mod tests {
     fn size_of_register_rule_map() {
         use core::mem;
         let size = mem::size_of::<RegisterRuleMap<EndianSlice<NativeEndian>>>();
-        let max_size = 1040;
+        let max_size = 6152;
         if size > max_size {
             assert_eq!(size, max_size);
         }


philipc:
I think `ArrayVec` was intentional because this code is meant to be async signal safe and not allocate, as mentioned in the [comment](https://github.com/gimli-rs/gimli/blob/335fb1781243060e8620a54c7c962bb2c21ee547/src/read/cfi.rs#L2324-L2325).

I wonder how other unwinders handle this.

jan-auer:
Completely missed that comment. 

What if we keep the error and swap out ArrayVec vs SmallVec based on a feature flag? There may be contexts where signal safety is not required in which case it would be desirable to collect all rules.

philipc:
I'm not keen on using a feature flag to switch between ArrayVec and SmallVec. We need to be able to handle all rules, even when signal safety is required. libunwind uses an architecture specific value for its array (33 for x86-64, 97 for aarch64). How would you feel about increasing the fixed size for now, and later making it configurable with const generics?

jan-auer:
Sounds good to me. I should have done the reading in libunwind right away. Would you mind to link me to it, for some reason I'm not able to find it.

Increased the limit to 100 now (since `arrayvec::Array` is not implemented for size 97). The first part of this comment is now a bit redundant, since we *are* actually allocating that many rules:

https://github.com/gimli-rs/gimli/blob/53c802a69066b211984ef06fd96ba20946d47ca3/src/read/cfi.rs#L2297-L2299

Would you like me to change that?

philipc:
> Would you mind to link me to it, for some reason I'm not able to find it.

Sure: [the register rule table](https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/dwarf.h#L246-L251), and the [x86-64](https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-x86_64/dwarf-config.h#L36) and [aarch64](https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-aarch64/dwarf-config.h#L32) numbers. [ARM (128)](https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-arm/dwarf-config.h#L31) and [MIPS (188)](https://github.com/libunwind/libunwind/blob/11fd461095ea98f4b3e3a361f5a8a558519363fa/include/tdep-mips/dwarf-config.h#L31) are actually even larger.

> Would you like me to change that?

Yeah, can you change the comment to mention that it is based on the numbers from libunwind?

jan-auer:
Thanks for the links, @philipc! Was going back and forth with using a smaller array, but ended up increasing size to `192`. This bumps sizes of the register map to 6kB and the unwind context to a whopping 7+ pages (30kB).

philipc:
Just out of interest, the benchmark difference on my machine is:
```
 name                                                       before ns/iter  after ns/iter  diff ns/iter   diff %  speedup 
 cfi::eval_longest_fde_instructions_new_ctx_everytime       498             2,893                 2,395  480.92%   x 0.17 
 cfi::eval_longest_fde_instructions_same_ctx                314             651                     337  107.32%   x 0.48 
 cfi::iterate_entries_and_do_not_parse_any_fde              50,855          49,392               -1,463   -2.88%   x 1.03 
 cfi::iterate_entries_and_parse_every_fde                   387,405         391,387               3,982    1.03%   x 0.99 
 cfi::iterate_entries_and_parse_every_fde_and_instructions  889,495         896,817               7,322    0.82%   x 0.99 
 cfi::iterate_entries_evaluate_every_fde                    1,204,579       2,542,256         1,337,677  111.05%   x 0.47 
 cfi::parse_longest_fde_instructions                        203             209                       6    2.96%   x 0.97 
```

I don't have any ideas for what we can do better, but maybe @fitzgen does? I don't think anyone has spent any serious effort on optimising this yet though, other than #148 (which may no longer be an optimisation).

fitzgen:
@philipc 

> I wonder how other unwinders handle this.

Some versions of libunwind (there are so many forks floating around) have a specialized mmap-only, lock-free allocator for use inside signal handlers. Given the snippets you found, it looks like that isn't used for register rules, but it is something we could look into.

Other unwinders, that are focused just on fast, common case unwinding for profiling, only ever recover rsp and rbp.

Exposing this limit through generics somehow seems promising and not overly ocean boil-y.

fitzgen:
> I don't have any ideas for what we can do better, but maybe @fitzgen does? I don't think anyone has spent any serious effort on optimising this yet though, other than #148 (which may no longer be an optimisation).

The fastest implementations I've seen either compile CFI to machine code ahead of time and replace the CFI, or do it at runtime with a super limited JIT compiler and an associative cache mapping PCs to JIT stubs.

That is, they largely avoid the cost of repeatedly parsing and interpreting CFI rules.

philipc:
> Some versions of libunwind (there are so many forks floating around) have a specialized mmap-only, lock-free allocator for use inside signal handlers. Given the snippets you found, it looks like that isn't used for register rules, but it is something we could look into.

Yeah libunwind (at least the primary version of it) uses mmap for `DW_CFA_remember_state`. The libgcc unwinder does the same thing as libunwind for the register array, and uses alloca for `DW_CFA_remember_state`.

> The fastest implementations I've seen either compile CFI to machine code ahead of time and replace the CFI, or do it at runtime with a super limited JIT compiler and an associative cache mapping PCs to JIT stubs.

I'm aware of https://fzn.fr/projects/frdwarf/, but not sure if anything in production uses that technique.

Anyway, it sounds like this PR is the way to go for now at least.

fitzgen:
> but not sure if anything in production uses that technique.

Firefox's profiler's unwinder uses the JIT approach.

PR #493, opened 2020-05-06
Support all CIE versions in eh_frame

jan-auer:
While `.eh_frame` always has version 1, it can contain CIEs with version 3. It also seems that CIE v1 always has U8 encoding for the return address, as opposed to the currently used LEB128. This would make it equivalent to `.debug_frame`. I am not sure why it was implemented like this in the first place, but considering that the RA is rarely >= 128, this might just have been coincidental.

My guess is that version 4 could potentially occur, although I've not seen it in the wild and would conservatively not implement it at the moment. LLVM seems to treat CIEs coming from debug_frame and eh_frame the same.

Fixes https://github.com/gimli-rs/gimli/issues/244
See https://www.airs.com/blog/archives/460

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -468,15 +468,6 @@ pub enum CieOffsetEncoding {
     U64,
 }
 
-// Ditto about being `pub`.
-#[doc(hidden)]
-#[allow(missing_docs)]
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum ReturnAddressRegisterEncoding {
-    U8,
-    Uleb,
-}
-
 /// An offset into an `UnwindSection`.
 //
 // Needed to avoid conflicting implementations of `Into<T>`.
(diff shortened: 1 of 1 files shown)


philipc:
I think this bug is simply a result of bad documentation. The [LSB spec](https://refspecs.linuxfoundation.org/LSB_4.1.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html) says "Version: A 1 byte value that identifies the version number of the frame information structure. This value shall be 1.", and it doesn't give the return address encoding. We did notice problems in #244, so I guess this PR fixes that.

src/read/cfi.rs
```diff
@@ -808,8 +814,12 @@ impl<R: Reader> _UnwindSectionPrivate<R> for EhFrame<R> {
         0
     }
 
-    fn return_address_register_encoding(_version: u8) -> ReturnAddressRegisterEncoding {
```
philipc, line 811: Since they are both the same now, we should delete this method from the trait.

jan-auer: Good point, updated. Would you go even further and also support CIE v4; effectively removing the `compatible_version` method, too?

philipc: Yeah, given that we already know the doc is wrong, we should do what other consumers do. I haven't recently looked at gdb, but I think that it handles eh_frame and debug_frame identically for version checks.

jan-auer: Thanks, updated.

philipc: Can you delete the `pub enum ReturnAddressRegisterEncoding` definition too, and it's probably worth squashing these commits. Otherwise this looks good.

jan-auer: Ah yes, must have missed the warning. Amended.

jan-auer:
@philipc would it be possible to release a patch version?

philipc:
I would prefer to release a minor version with all of the fixes since the last release (e.g. #487 too), if that works for you.

PR #497, opened 2020-05-07
Remove the `arrayvec` dependency

alexcrichton:
This commit removes the `arrayvec` dependency from `gimli` to continue
the work started in #494. The goal here is to absolutely minimize the
number of dependencies (ideally zero) when using only the bare bones of
the crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ travis-ci = { repository = "gimli-rs/gimli" }
 coveralls = { repository = "gimli-rs/gimli" }
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


alexcrichton:
Rebased!

alexcrichton:
Er I also just realized that the `R` parameter could be stored by-value in the vector, which means that destructors and such need to be handled. I mistakenly thought `RegisterRule<R>` was effectively always `Copy`. Should be good now!

philipc:
I'm getting a little concerned about the amount of unsafe code that's needed for this now, but I don't have any better ideas.

PR #498, opened 2020-05-10
Remove dependency of `read` on `stable_deref_trait`

alexcrichton:
This commit adds a new feature to this crate, enabled by default, called
`endian-reader`. The goal here is to remove the need for the `backtrace`
crate to depend on `stable_deref_trait`, and it looks like currently
`addr2line` and friends don't need the only usage of
`stable_deref_trait`, which is the `EndianReader` type.

Another possible solution might be to break up the `read` feature, too,
and have `addr2line` only enable a small subset of the `read` feature
that it actually needs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,10 +33,11 @@ test-assembler = "0.1.3"
 typed-arena = "2"
 
 [features]
-read = ["stable_deref_trait"]
+read = []
+endian-reader = ["stable_deref_trait"]
 write = ["indexmap"]
 std = ["fallible-iterator/std", "stable_deref_trait/std"]
-default = ["read", "write", "std", "fallible-iterator"]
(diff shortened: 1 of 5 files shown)


philipc:
> Another possible solution might be to break up the read feature, too,
and have addr2line only enable a small subset of the read feature
that it actually needs.

Use of `EndianReader` is orthogonal to the rest of the read feature, so I think this PR is the correct solution.

Tests need to depend on this feature though to fix CI.

alexcrichton:
Whoops sorry about that, should be fixed now!

philipc:
Do you think that's the last change you'll need for now?

alexcrichton:
I believe so! If you're up for publishing the gimli family of crates I'll look next to switching over the `backtrace` crate on cates.io to gimli by default

PR #503, opened 2020-05-13
Prepare for inclusion into libstd

alexcrichton:
This commit adds the "standard preamble" in `Cargo.toml` which makes
this crate suitable for being used as a dependency of the standard
library. The intention here is that this is a "fire and forget" snippet
added to `Cargo.toml` where it doesn't really impact anything else in
this crate. It's unlikely to be used too too soon for libstd, but I
figure it wouldn't be bad to get ready!

(also see https://github.com/rust-lang/backtrace-rs/issues/328 for some context)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,12 @@ fallible-iterator = { version = "0.2.0", default-features = false, optional = tr
 indexmap = { version = "1.0.2", optional = true }
 stable_deref_trait = { version = "1.1.0", default-features = false, optional = true }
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
@philipc would you be ok publishing this to get the addr2line changes in as well to start removing some `[patch]` needed for building libstd?

philipc:
There's one outstanding addr2line PR that I want to get merged, and then I'll do gimli and addr2line releases.

PR #505, opened 2020-05-14
read: Fix subtraction overflow when `Abbreviations::get` is given `0`

fitzgen:
Alternatively, we could make this take `NonZeroU64` instead of `u64`, but those can be kind of a pain to work with.

### DIFF
--- a/src/read/abbrev.rs
+++ b/src/read/abbrev.rs
@@ -2,6 +2,7 @@
 
 use alloc::collections::btree_map;
 use alloc::vec::Vec;
+use core::convert::TryFrom;
 use core::fmt::{self, Debug};
 use core::iter::FromIterator;
 use core::ops::Deref;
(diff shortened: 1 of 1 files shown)


src/read/abbrev.rs
```diff
@@ -154,9 +154,9 @@ impl Abbreviations {
     /// Get the abbreviation associated with the given code.
     #[inline]
     pub fn get(&self, code: u64) -> Option<&Abbreviation> {
-        let code_usize = code as usize;
-        if code_usize as u64 == code && code_usize - 1 < self.vec.len() {
-            Some(&self.vec[code_usize - 1])
+        let index = (code as usize).checked_sub(1)?;
```
philipc, line 157: Can you change this to `try_into` as well?

I also feel like we should simply use `wrapping_sub` here, but it probably doesn't matter.  That would mean we might end up looking up a 0 in the map, but it would still be correct and this function should never be called with 0 anyway. Is this something you encountered, or are you just auditing arithmetic?

fitzgen: > Is this something you encountered, or are you just auditing arithmetic?

Encountered during fuzzing. Not that any gimli-internal debug_info look up code was doing `get(0)` but the fuzz target itself did.

If we don't want this to ever be called with zero, then we should probably switch it to use `NonZeroU64`.

PR #507, opened 2020-05-15
read: Make `Abbreviations::get` take a `NonZeroU64`

fitzgen:
This fixes a subtraction overflow when `Abbreviations::get` is called with zero now.

Alternative fix to #505; @philipc do you have a preference between these two patches?


philipc:
Let's stick with the `checked_sub` in #505.

fitzgen:
> Let's stick with the `checked_sub` in #505.

:+1:

PR #509, opened 2020-05-15
Let arithmetic overflows wrap when evaluating unwind sections

### LINKED ISSUE #112: CFI evaluation should have robust integer overflow and casting
We do a bunch of unchecked adds and casting between integer types with different ranges right now, and we should do a pass to make it more robust.

fitzgen:
Similar to #508: The DWARF and `.eh_frame` specs don't specify how to handle
overflows. So we will use wrapping.

Fixes #112


fitzgen:
> Is this enough to close #112 too?

Yes, I think so.

PR #511, opened 2020-05-18
Switch from Travis CI to GitHub Actions

fitzgen:
Technically, I already created an initial github actions install, because it won't run CI until you've merged it into the master branch and that involves the inevitable config tweaking: https://github.com/gimli-rs/gimli/blob/master/.github/workflows/rust.yml

Github actions gives us ten parallel jobs at the free, open source tier, while travis only gives us two now. I also removed a few of the redundant matrix combinations. The result is that CI is way faster now. I don't have actual numbers, but I'm pretty sure we were at *at least* a half hour before, and it is probably ten minutes or so now.


.travis.yml
```diff
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
```
philipc, line 1: `sudo` is deprecated now.

.travis.yml
```diff
@@ -27,39 +21,6 @@ before_script:
...
-    # Build a big-endian target.
-    - rust: stable
-      sudo: required
-      services: docker
-      env:
-        - TARGET=powerpc64-unknown-linux-gnu
-        - GIMLI_JOB=cross
+script: |
+  RUSTFLAGS="--cfg procmacro2_semver_exempt" cargo install cargo-tarpaulin
+  cargo tarpaulin --verbose --ciserver travis-ci --coveralls "$TRAVIS_JOB_ID"
```
philipc, line 26: Are we keeping this because of a github actions limitiation?

philipc:
FYI, I recently changed coveralls to only use the status API (instead of comments) to reduce the spam. It makes it less visible though, and even more so with this number of actions, but I think it's still ok? It's also configured to never fail, which may not be the best...

.travis.yml
fitzgen: We were using it for the coverage job before, and this just carries that over. I have no idea if it is actually required or not.

.travis.yml
fitzgen: Nope, just because I haven't looked into how to configure `cargo tarpaulin` and github actions together, nor added a token for them to talk to each other.

fitzgen:
> FYI, I recently changed coveralls to only use the status API (instead of comments) to reduce the spam. It makes it less visible though, and even more so with this number of actions, but I think it's still ok? It's also configured to never fail, which may not be the best...

I didn't really mind the spam, but I don't really care either way.

fitzgen:
I'm going to go ahead and merge this now, and leave the final migration of the coverage job from travis to github actions for later.

PR #512, opened 2020-05-19
Add fuzzing infrastructure

fitzgen:
This adds a few fuzz targets for various DWARF sections, and also integrates it with our CI, so that the fuzz targets are built in one job, and then each fuzz target is run in N parallel jobs for five minutes.

The CI integration bits will likely involve a few tries and force pushes >.<


fitzgen:
Ok, I think this is good to go now!

PR #515, opened 2020-06-25
Don't fail if unable to read the string for DW_AT_comp_dir or DW_AT_name

### LINKED ISSUE #514: Handle DW_TAG_partial_unit
I'm trying to use dump_syms which depends on symbolic which itself depends on gimli to dump breakpad symbols for iris_dri.so-20.1.1-1.fc32.x86_64.debug.
I get neither DW_TAG_subprogram nor DW_TAG_inlined_subroutine, all the units are either empty or erroneous.
I llvm-dwarf dump the file:
The first 54 CU looks like:
```
0x00000000: Compile Unit: length = 0x00000013 version = 0x0004 abbr_offset = 0x2842 addr_size = 0x08 (next unit at 0x00000017)

0x0000000b: DW_TAG_partial_unit

0x0000000c:   DW_TAG_imported_unit
                DW_AT_import	(<alt 0x2a6a57>)

0x00000011:   DW_TAG_imported_unit
                DW_AT_import	(<alt 0x2a6a73>)

0x00000016:   NULL
```
These ones look empty.
And all the next CU start with something like:
```
0x000000cf: Compile Unit: length = 0x0000001b version = 0x0004 abbr_offset = 0x15987 addr_size = 0x08 (next unit at 0x000000ee)

0x000000da: DW_TAG_partial_unit
              DW_AT_stmt_list	(0x00000000)
              DW_AT_comp_dir	()

0x000000e3:   DW_TAG_imported_unit
                DW_AT_import	(<alt 0xb>)

0x000000e8:   DW_TAG_imported_unit
                DW_AT_import	(<alt 0x2a6b29>)

0x000000ed:   NULL
```
they look like erroneous.
I greped on `TAG_partial_unit` on the gimli source and I got almost nothing so maybe it's because gimli doesn't handle partial unit.

jrmuizel:
This matches the behaviour of llvm-dwarfdump:
e.g. dwarf::toString(UnitDie.find(DW_AT_comp_dir)) just returns an
Optional and doesn't fail.

Fixes #514

### DIFF
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -608,11 +608,11 @@ impl<R: Reader> Unit<R> {
         }
 
         unit.name = match name {
-            Some(val) => Some(dwarf.attr_string(&unit, val)?),
+            Some(val) => dwarf.attr_string(&unit, val).ok(),
             None => None,
         };
         unit.comp_dir = match comp_dir {
-            Some(val) => Some(dwarf.attr_string(&unit, val)?),
+            Some(val) => dwarf.attr_string(&unit, val).ok(),
             None => None,
         };
         unit.line_program = match line_program_offset {


philipc:
I think this change is good, but I'm also curious why we are unable to read those strings. Is there another bug causing that?

jrmuizel:
I think offsets are outside of the file. We get an `Err` value: UnexpectedEof(ReaderOffsetId(4497503920))' and the file is only 112197928 bytes long. llvm-dwarfdump is also seems unable to read the strings.

philipc:
The value in `ReaderOffsetId` isn't an offset (it's a memory address), but if llvm-dwarfdump can't read them then that's fine.

PR #516, opened 2020-06-29
Some dwo support in dwarfdump

khuey:
This allows dumping a .dwo directly. It doesn't traverse the <regular debug info> -> dwo linkage though.


khuey:
Yes. As I said above "It doesn't traverse the -> dwo linkage though."

PR #517, opened 2020-07-02
Add dwo section names.

khuey:


philipc:
Should dwarfdump use this too?

khuey:
Sure, why not.

Issue #371, opened 2019-02-01
Handle more DWARF 5 unit types


philipc:
DWARF 5 allows the following unit types:

- DW_UT_compile
- DW_UT_type
- DW_UT_partial
- DW_UT_skeleton
- DW_UT_split_compile
- DW_UT_split_type

I propose we delete the existing `struct CompilationUnitHeader` and `struct TypeUnitHeader` definitions, and instead have a `unit_type: enum UnitType` field in `struct UnitHeader`. This will allow consumers to handle all of these units without needing to duplicate code for each unit type.

PR #518, opened 2020-07-02
Add Dwarf::attr_address which will convert addr and addrx forms to their proper values

khuey:
Just like attr_string and attr_ranges.


khuey:
Ah, right, this needs other changes in my queue. Let's try it again later.

philipc:
Regarding handling DWARF 5 unit types, my thought in #371 was to use an enum instead of a trait, but it's not something I've looked into much. I feel there is too much duplication for the current compilation units and type units, and enum may help avoid that.

PR #521, opened 2020-07-02
Support skeleton and split compilation units

khuey:
The gist of this is that in DWARF 5 the DWO id for skeleton/split units lives in the header rather than as an attribute on the CU, so it needs to be parsed and saved somewhere.  I chose not to revisit the decision to have a separate `CompilationUnitHeader` and `TypeUnitHeader`, and instead replaced `UnitHeader` with a trait that they both implement. What was previously `UnitHeader` is now `CommonUnitHeader` and used in many fewer places. Finally, `CompilationUnitHeader` grew a `CompilationUnitType` field that tells us whether we have a conventional, skeleton, or split unit (and what the DWO id is in the latter cases).

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -932,6 +932,10 @@ where
             UnitType::Type {
                 type_signature,
                 type_offset,
+            }
+            | UnitType::SplitType {
+                type_signature,
+                type_offset,
             } => {
                 write!(buf, "  signature        = ")?;
                 dump_type_signature(buf, type_signature)?;
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -927,6 +927,23 @@ where
             header.offset().as_debug_info_offset().unwrap().0,
         )?;
 
+        match header.type_() {
+            UnitType::Compilation | UnitType::Partial => (),
+            UnitType::Type {
+                type_signature,
+                type_offset,
+            } => {
+                write!(buf, "  signature        = ")?;
+                dump_type_signature(buf, type_signature)?;
+                writeln!(buf)?;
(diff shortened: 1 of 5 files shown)


philipc:
Does your application handle `.debug_types`? I'm not keen on this extra type parameter for `read::dwarf::Unit` and I'm wondering how well it works out in practice. Additionally, DWARF5 eliminates `.debug_types` and moves type units into `.debug_info`, so this solution doesn't seem to be able to handle that.

Either way, this is going to be a large breaking change, so I'll do a release first.

khuey:
No, I don't handle `.debug_types`. I've never actually seen it in the wild. gcc has a switch for it but I haven't seen it used.

philipc:
In order to handle type units in DWARF 5, type units need to be another variant in a `UnitType` enum. So there should be a `UnitHeadersIter` and a `UnitHeader` that can be used for both `.debug_info` and `.debug_types`. Offsets in these will need to be a `UnitSectionOffset`. I can implement this if you prefer.

khuey:
I'm happy to implement it but I'm not entirely sure what you're suggesting. Do you mean to combine `TypeUnitHeader` and `CompilationUnitHeader` (thus leaving type units as a variant in the `UnitType` enum) and their associated iterator machinery, even for DWARF 4?

philipc:
Yes. So `Dwarf::units` will return an iterator for all units in the `.debug_info` section, including DWARF 5 type units. And `Dwarf::type_units` will return an iterator for all units in the `.debug_types` section. These iterators will have the same type, they will simply be iterating over different section data.

khuey:
Alright this is rebased on top of #523, plus a few extras now that type units are mixed in.

PR #525, opened 2020-07-06
Pick up non-standard GNU extension attributes when looking for address and range list bases.

khuey:


khuey:
Woops pushed to the wrong branch there. It's back to the correct commit now.

PR #529, opened 2020-07-12
More split dwarf work

khuey:
The pre-DWARF5 GNU split dwarf extension differs from the standardized version in at least two ways:

1. The various new list sections do not have the headers that they have in the standardized version.
2. The .debug_loclists section can be present in a .dwo as .debug_loc.dwo, with a slightly different encoding format.

Additionally, the default values for the various base attributes are wrong in DWARF 5, where there is the section header (that is not present in the GNU split dwarf extension) that must be skipped.

To handle all of this, I fix the default base handling by keying off the DWARF version, then I refactor LocationLists and RangeLists because only one of the two sections is present at any given time. Which encoding to use for these lists is no longer keyed off the DWARF version but rather the section in use. That allows the DWARF version to be used to distinguish the pre-DWARF5 GNU split dwarf encoding for location lists.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -550,7 +550,10 @@ where
         reader: Default::default(),
     };
 
-    let dwarf = gimli::Dwarf::load(&mut load_section, |_| Ok(no_reader.clone())).unwrap();
+    let mut dwarf = gimli::Dwarf::load(&mut load_section, |_| Ok(no_reader.clone())).unwrap();
+    if flags.dwo {
+        dwarf.file_type = gimli::DwarfFileType::Dwo;
+    }
 
     let out = io::stdout();
     if flags.eh_frame {
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -381,13 +381,22 @@ impl<R: Reader> Dwarf<R> {
         unit: &Unit<R>,
         offset: LocationListsOffset<R::Offset>,
     ) -> Result<LocListIter<R>> {
-        self.locations.locations(
-            offset,
-            unit.encoding(),
-            unit.low_pc,
-            &self.debug_addr,
-            unit.addr_base,
-        )
+        match self.file_type {
+            DwarfFileType::Main => self.locations.locations(
+                offset,
+                unit.encoding(),
+                unit.low_pc,
+                &self.debug_addr,
+                unit.addr_base,
+            ),
+            DwarfFileType::Dwo => self.locations.locations_dwo(
+                offset,
+                unit.encoding(),
+                unit.low_pc,
+                &self.debug_addr,
+                unit.addr_base,
+            ),
+        }
     }
 
     /// Try to return an attribute value as a location list offset.
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -660,7 +660,9 @@ impl<R: Reader> Unit<R> {
     pub fn copy_relocated_attributes(&mut self, other: &Unit<R>) {
         self.low_pc = other.low_pc;
         self.addr_base = other.addr_base;
-        self.rnglists_base = other.rnglists_base;
+        if self.header.version() < 5 {
+            self.rnglists_base = other.rnglists_base;
+        }
     }
 }
 
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -3,8 +3,8 @@ use alloc::string::String;
 use crate::common::{
     DebugAddrBase, DebugAddrIndex, DebugInfoOffset, DebugLineStrOffset, DebugLocListsBase,
     DebugLocListsIndex, DebugRngListsBase, DebugRngListsIndex, DebugStrOffset, DebugStrOffsetsBase,
-    DebugStrOffsetsIndex, DebugTypesOffset, Encoding, LocationListsOffset, RangeListsOffset,
-    SectionId, UnitSectionOffset,
+    DebugStrOffsetsIndex, DebugTypesOffset, DwarfFileType, Encoding, LocationListsOffset,
+    RangeListsOffset, SectionId, UnitSectionOffset,
 };
 use crate::constants;
 use crate::read::{
@@ -14,23 +14,6 @@ use crate::read::{
     IncompleteLineProgram, LocListIter, LocationLists, Range, RangeLists, Reader, ReaderOffset,
     ReaderOffsetId, Result, RngListIter, Section, UnitHeader, UnitOffset,
 };
-
-/// The "type" of file with DWARF debugging information. This determines, among other things,
-/// which files DWARF sections should be loaded from.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum DwarfFileType {
-    /// A normal executable or object file.
-    Main,
-    /// A .dwo split DWARF file.
-    Dwo,
-    // TODO: Supplementary files, .dwps?
-}
-
-impl Default for DwarfFileType {
-    fn default() -> Self {
-        DwarfFileType::Main
-    }
-}
 
 /// All of the commonly used DWARF sections, and other common information.
 #[derive(Debug, Default)]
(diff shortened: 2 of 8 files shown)


philipc:
> Additionally, the default values for the various base attributes are wrong in DWARF 5, where there is the section header (that is not present in the GNU split dwarf extension) that must be skipped.

Why does this matter? Shouldn't those values be set by the attributes? From memory, the reason we need defaults for the GNU extensions is because they don't use those attributes.

> Since these sections are mutually exclusive, we can use an enum to represent them.

They aren't mutually exlusive. An executable can contain both DWARF 4 and DWARF 5 debuginfo since each object file can be compiled with different settings.

khuey:
> 
> 
> > Additionally, the default values for the various base attributes are wrong in DWARF 5, where there is the section header (that is not present in the GNU split dwarf extension) that must be skipped.
> 
> Why does this matter? Shouldn't those values be set by the attributes? From memory, the reason we need defaults for the GNU extensions is because they don't use those attributes.

The attributes are not always present. e.g. clang does not emit the attributes if the values are the defaults.

> > Since these sections are mutually exclusive, we can use an enum to represent them.
> 
> They aren't mutually exlusive. An executable can contain both DWARF 4 and DWARF 5 debuginfo since each object file can be compiled with different settings.

Ugh, that's obnoxious. Is that really something we care to support?

The alternative here is probably to add a new `CompatGnuDebugLocDwo` type and include it in `LocationLists`.

philipc:
> The attributes are not always present. e.g. clang does not emit the attributes if the values are the defaults.

From what I can tell, that's only for split DWARF, and it works because these values should be inherited from the the skeleton unit. So the correct fix is to inherit those values, instead of guessing where it will be. I can be convinced otherwise if you can show me code in a consumer (e.g. gdb or lldb) that calculates these values instead.

> Is that really something we care to support?

Yes. It probably occurs more due to objects being produced by different tools rather than different settings. There's less of a case for it in split DWARF, but we definitely need to support it otherwise. In split DWARF, I think it would be fine to treat the `.debug_loc.dwo` section the same as `.debug_loclists.dwo` (so the version check shouldn't be done for DWO files). I haven't investigated this in depth though.

philipc:
So the bug is that `RawLocListIter` shouldn't be using `encoding.version`. Instead it should remember if this was `LocationLists::debug_loc` or `LocationLists::debug_loclists`.

khuey:
> 
> 
> > The attributes are not always present. e.g. clang does not emit the attributes if the values are the defaults.
> 
> From what I can tell, that's only for split DWARF,

Right.

> and it works because these values should be inherited from the skeleton unit.

No. There are four sections with relevant base attributes: `.debug_addr`, `.debug_str_offsets`, `.debug_loclists`, and `.debug_rnglists`. Of those two (`.debug_str_offsets` and `.debug_loclists`) exist in the DWO where there's no need for a explicit base attribute because there's only one unit's worth of data. The `DW_AT_str_offsets_base` on a skeleton unit is controlling *only* for attributes that appear on the skeleton unit (e.g. `DW_AT_comp_dir`) and not for string attributes that appear on the split unit in the DWO. This is why I didn't add it to `copy_relocated_attributes`.

I thought clang sometimes omitted the `DW_AT_rnglists_base` but now that I'm looking at it again I think it only omits it for units with no ranges.

> So the correct fix is to inherit those values, instead of guessing where it will be. I can be convinced otherwise if you can show me code in a consumer (e.g. gdb or lldb) that calculates these values instead.

gdb doesn't actually support DWARF 5. lldb is doing similar calculations at https://github.com/llvm/llvm-project/blob/e808cab824488af137b62902e65dec3827b83b46/lldb/source/Plugins/SymbolFile/DWARF/DWARFUnit.cpp#L425. LLVM parses the section headers so it's adjusting the specified offsets down. We're not so we need to adjust the default offset up.

> > Is that really something we care to support?
> 
> Yes. It probably occurs more due to objects being produced by different tools rather than different settings. There's less of a case for it in split DWARF, but we definitely need to support it otherwise. In split DWARF, I think it would be fine to treat the `.debug_loc.dwo` section the same as `.debug_loclists.dwo` (so the version check shouldn't be done for DWO files). I haven't investigated this in depth though.

Yeah ok that's not super-terrible then.

khuey:
Table F.1 in the DWARF 5 spec is pretty clear about how the base attributes work. `DW_AT_rnglists_base` and `DW_AT_loclists_base` are never used in a split DWARF setup.

philipc:
Ok I can see that clang does calculate the header size in some cases.

Still trying to understand how what you have said matches this text from the standard:

> The following attributes are not part of a split full compilation unit entry but instead are
inherited (if present) from the corresponding skeleton compilation unit: DW_AT_low_pc,
DW_AT_high_pc, DW_AT_ranges, DW_AT_stmt_list, DW_AT_comp_dir,
DW_AT_str_offsets_base, DW_AT_addr_base and DW_AT_rnglists_base.


>Table F.1 in the DWARF 5 spec is pretty clear about how the base attributes work. DW_AT_rnglists_base and DW_AT_loclists_base are never used in a split DWARF setup.

This recently was discussed on the dwarf mailing list, and the conclusion is this is an error in the standard.

khuey:
> Still trying to understand how what you have said matches this text from the standard:
> 
> > The following attributes are not part of a split full compilation unit entry but instead are
> > inherited (if present) from the corresponding skeleton compilation unit: DW_AT_low_pc,
> > DW_AT_high_pc, DW_AT_ranges, DW_AT_stmt_list, DW_AT_comp_dir,
> > DW_AT_str_offsets_base, DW_AT_addr_base and DW_AT_rnglists_base.

The text is wrong. Having `DW_AT_rnglists_base` on a skeleton unit makes no sense because the range lists live in the DWO. And I discussed how the strings stuff works above.

> > Table F.1 in the DWARF 5 spec is pretty clear about how the base attributes work. DW_AT_rnglists_base and DW_AT_loclists_base are never used in a split DWARF setup.
> 
> This recently was discussed on the dwarf mailing list, and the conclusion is this is an error in the standard.

How, exactly?

khuey:
> So the bug is that `RawLocListIter` shouldn't be using `encoding.version`. Instead it should remember if this was `LocationLists::debug_loc` or `LocationLists::debug_loclists`.

That doesn't actually help because the same encoding is used to choose between `debug_loc` and `debug_loclists`.

Put another way, in the DWARF 4 GNU split dwarf case `LocationsList::raw_locations` needs to somehow know to choose the `debug_loclists` branch instead of the `debug_loc` branch.

philipc:
> That doesn't actually help because the same encoding is used to choose between debug_loc and debug_loclists.

Right, so that check needs changing too. The sections aren't mutually exclusive, but their use within a unit is. So I think which format to use needs to be derived from properties of the unit (its version) and the file (is it a DWO?). Does that sound right?

khuey:
> 
> 
> > That doesn't actually help because the same encoding is used to choose between debug_loc and debug_loclists.
> 
> Right, so that check needs changing too. The sections aren't mutually exclusive, but their use within a unit is. So I think which format to use needs to be derived from properties of the unit (its version) and the file (is it a DWO?). Does that sound right?

That's right, but currently nothing in gimli knows about DWARF v4 split dwarf files at all. I guess we could pass that information into `Dwarf` or something.

philipc:
I think your changes to the base attribute defaults are good, but I'll look at it a bit more later to make sure I understand.

khuey:
I think actually I should push the branching on the version into `default_from_encoding` and call it for both the v4 and v5 cases.

khuey:
Alright how about this version?

examples/dwarfdump.rs
```diff
@@ -365,7 +365,7 @@ struct Flags {
 impl Flags {
     fn section_name(&self, id: gimli::SectionId) -> Option<&'static str> {
         if self.dwo {
-            id.dwo_name()
+            id.dwo_names().get(0).map(|s| *s)
```
philipc, line 368: This will still only load one of the section names.

Rather than allowing multiple names per section, what if leave `.debug_loc.dwo` and `.debug_loclists.dwo` as separate sections, and choose which one to use in `raw_locations_dwo` based on the version?

src/read/dwarf.rs
```diff
@@ -376,13 +381,22 @@ impl<R: Reader> Dwarf<R> {
...
-        )
+        match self.file_type {
+            DwarfFileType::Main => self.locations.locations(
+                offset,
+                unit.encoding(),
+                unit.low_pc,
+                &self.debug_addr,
+                unit.addr_base,
+            ),
+            DwarfFileType::Dwo => self.locations.locations(
```
philipc, line 392:
```suggestion
            DwarfFileType::Dwo => self.locations.locations_dwo(
```

khuey: doh!

examples/dwarfdump.rs
khuey: Yeah ok that makes sense.

khuey:
Comments addressed.

PR #532, opened 2020-08-29
Add a name_to_register routine to convert names into register numbers

nbaksalyar:
This PR adds a converse of `register_name`, i.e. a function that converts a string register name into a register number. It's useful for debugger command-line interfaces.

Please let me know if this needs a test case.

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -19,6 +19,16 @@ macro_rules! registers {
                     _ => return None,
                 }
             }
+
+	    /// Converts a register name into a register number.
+	    pub fn name_to_register(value: &str) -> Option<Register> {
+		match value {
+                    $(
+                        $disp => Some(Self::$name),
+                    )+
+                    _ => return None,
+		}
+	    }
         }
     };
 }


src/arch.rs
```diff
@@ -19,6 +19,16 @@ macro_rules! registers {
                     _ => return None,
                 }
             }
+
+	    /// Converts a register name into a register number.
+	    pub fn name_to_register(value: &'static str) -> Option<Register> {
```
philipc, line 24:
```suggestion
	    pub fn name_to_register(name: &str) -> Option<Register> {
```

philipc (changes requested):
I suspect that this is not really what you want.

- do you want `RA` to be a register name?
- how are you going to handle `rax`/`eax`/`ax`/`ah`/`al`?
- how are you going to handle synonyms for ARM register names?

I haven't looked at what gdb/lldb do, but you may not even want to use the DWARF register numbers in user interface code. One of the other register numbering schemes might be better (e.g. ptrace or your assembler/disassembler).

nbaksalyar:
> I haven't looked at what gdb/lldb do, but you may not even want to use the DWARF register numbers in user interface code

Yes, more likely these are going to be used only as helper methods and not directly in the UI code.

Thanks for the review!

philipc (approved):
Thanks!

PR #539, opened 2020-12-19
.debug_aranges improvements

philipc:
These are for use in addr2line.


fitzgen (commented):
I was on vacation and not reading github emails for the last couple weeks, but this looks great!

PR #541, opened 2021-01-01
Support DW_FORM_addrx* in low_pc and high_pc

### LINKED ISSUE #540: Dwarf::{unit_ranges, die_ranges} doesn't handle DW_FORM_addrx
In DWARF 5, `DW_AT_low_pc` can be encoded using `DW_FORM_addrx` reference. For example:

```
void foo() {}
```

compiled with

```
$ clang temp.c -o temp.o -gdwarf-5 -c
```

results in

```
0x0000000c: DW_TAG_compile_unit [1] *
              DW_AT_producer [DW_FORM_strx1]    (indexed (00000000) string = "Ubuntu clang version 11.0.1-++20201103062553+ef4ffcafbb2-1~exp1~20201103053207.128")
              DW_AT_language [DW_FORM_data2]    (DW_LANG_C99)
              DW_AT_name [DW_FORM_strx1]        (indexed (00000001) string = "temp.c")
              DW_AT_str_offsets_base [DW_FORM_sec_offset]       (0x00000008)
              DW_AT_stmt_list [DW_FORM_sec_offset]      (0x00000000)
              DW_AT_comp_dir [DW_FORM_strx1]    (indexed (00000002) string = "/c/Users/me/Documents/balin")
              DW_AT_low_pc [DW_FORM_addrx]      (indexed (00000000) address = 0x0000000000000000 ".text")
              DW_AT_high_pc [DW_FORM_data4]     (0x00000006)
              DW_AT_addr_base [DW_FORM_sec_offset]      (0x00000008)

0x00000023:   DW_TAG_subprogram [2]  
                DW_AT_low_pc [DW_FORM_addrx]    (indexed (00000000) address = 0x0000000000000000 ".text")
                DW_AT_high_pc [DW_FORM_data4]   (0x00000006)
                DW_AT_frame_base [DW_FORM_exprloc]      (DW_OP_reg6 RBP)
                DW_AT_name [DW_FORM_strx1]      (indexed (00000003) string = "foo")
                DW_AT_decl_file [DW_FORM_data1] ("/c/Users/me/Documents/balin/temp.c")
                DW_AT_decl_line [DW_FORM_data1] (1)
                DW_AT_external [DW_FORM_flag_present]   (true)

0x0000002e:   NULL
```

Unfortunately, `Dwarf::die_ranges` (and, correspondingly, `Dwarf::unit_ranges` which uses the former under the hood) currently handles only absolute addresses, and return empty iterators for nodes in the output above.

P.S. If necessary, I can extract and write a minimal example to parse the file and call `unit_ranges` and `die_ranges` on those nodes, but was hoping that this description would be self-sufficient.

RReverser:
Fixes #540.

### DIFF
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -334,14 +334,19 @@ impl<R: Reader> Dwarf<R> {
         while let Some(attr) = attrs.next()? {
             match attr.name() {
                 constants::DW_AT_low_pc => {
-                    if let AttributeValue::Addr(val) = attr.value() {
-                        low_pc = Some(val);
-                    }
+                    low_pc = Some(
+                        self.attr_address(unit, attr.value())?
+                            .ok_or(Error::UnsupportedAttributeForm)?,
+                    );
                 }
                 constants::DW_AT_high_pc => match attr.value() {
-                    AttributeValue::Addr(val) => high_pc = Some(val),
                     AttributeValue::Udata(val) => size = Some(val),
-                    _ => return Err(Error::UnsupportedAttributeForm),
+                    attr => {
+                        high_pc = Some(
+                            self.attr_address(unit, attr)?
+                                .ok_or(Error::UnsupportedAttributeForm)?,
+                        );
+                    }
                 },
                 constants::DW_AT_ranges => {
                     if let Some(list) = self.attr_ranges(unit, attr.value())? {
@@ -563,6 +568,7 @@ impl<R: Reader> Unit<R> {
         let mut name = None;
         let mut comp_dir = None;
         let mut line_program_offset = None;
+        let mut low_pc_attr = None;
 
         {
             let mut cursor = unit.header.entries(&unit.abbreviations);
@@ -578,9 +584,7 @@ impl<R: Reader> Unit<R> {
                         comp_dir = Some(attr.value());
                     }
                     constants::DW_AT_low_pc => {
-                        if let AttributeValue::Addr(address) = attr.value() {
-                            unit.low_pc = address;
-                        }
+                        low_pc_attr = Some(attr.value());
                     }
                     constants::DW_AT_stmt_list => {
                         if let AttributeValue::DebugLineRef(offset) = attr.value() {
@@ -629,6 +633,11 @@ impl<R: Reader> Unit<R> {
             )?),
             None => None,
         };
+        if let Some(low_pc_attr) = low_pc_attr {
+            if let Some(addr) = dwarf.attr_address(&unit, low_pc_attr)? {
+                unit.low_pc = addr;
+            }
+        }
         Ok(unit)
     }
 


src/read/dwarf.rs
```diff
@@ -629,6 +633,11 @@ impl<R: Reader> Unit<R> {
             )?),
             None => None,
         };
+        if let Some(low_pc_attr) = low_pc_attr {
+            if let Some(addr) = dwarf.attr_address(&unit, low_pc_attr)? {
```
RReverser, line 637: Note: to be consistent, I followed example of `unit.{name,comp_dir,line_program}` above where invalid attribute form is silently ignored, but I'm not sure why it's done that way...

Wouldn't it be better to return explicit `UnsupportedAttributeForm` error in all those cases?

RReverser:
Other than the comment above, checked and this works as expected on my original DWARF example.

philipc (approved):
Thanks!

src/read/dwarf.rs
RReverser: Hmm, not sure either. I guess personally I would turn those fields into private OnceCells, and expose fallible getters as functions for them instead - that way, user could both handle an error, and not sacrifice any performance because variable would be cached once retrieved.

However, this would be a breaking API change, so probably not acceptable?

philipc: Not sure I properly understand what you mean. Using OnceCell, retrieving the variable would need to iterate the attributes again in order to initialize the OnceCell?

Breaking changes are fine if we are sure it is the correct thing to do (but for this case I'm not sure). We still don't support all of DWARF 5, and that is likely to require more breaking changes.

RReverser: > retrieving the variable would need to iterate the attributes again in order to initialize the OnceCell

Yeah, API-wise it would be equivalent of `DebuggingInformationEntry::attr(name)` which already does that, but here it would do it with caching. As you said yourself, many consumers are already fine with ignoring some fields or errors in them.

I think most consumers don't actually need to access _all_ of the `Unit` properties, and emitters usually don't emit many properties on Unit node, so iteration won't have any noticeable performance impact.

RReverser: As another alternative, you could keep current iteration, but store `Result`s in private fields, and clone them in such getters. But I really think that making field accessors lazy makes more sense for common scenarios, as it will reduce costs for parsing things user doesn't care about.

RReverser:
Could you please publish a patch release for this fix? For now I'm pointing to my Github fork with `[patch.crates-io]`, but I'd like to change that.

philipc:
I can't because there's already breaking changes on master, and in any case I'd prefer to wait a bit to be sure you don't need more fixes.

RReverser:
Ah okay. So far this is the only thing I ran into, I don't think I'll need anything else on Gimli side in the nearest future, but I guess makes sense to wait.

PR #549, opened 2021-02-07
use a more efficient method to copy [u8]

the8472:
I saw `clone_from_slice` showing up in some profiles and the assembly looks like a fairly dumb loop of `mov` instructions. `copy_from_slice` should optimize better since it calls `copy_nonoverlapping` (`memcpy`) internally.


philipc (changes requested):
Thanks! Could you change this for `EndianReader` too please?

philipc (approved):

PR #551, opened 2021-02-14
Change row and column numbers to be reported with NonZeroU64.

khuey:
This allows for more compact representation of their values (the return values of line() and column() should both be representable in a single u64 now) and makes it clear to users that Some(0) and ColumnType::Column(0) are never returned.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 --------------------------------------------------------------------------------
 
+## 0.24.0
+
+Not yet released
+
+### Breaking changes
+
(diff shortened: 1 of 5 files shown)


philipc (approved):
Thanks!

philipc:
If we're doing breaking changes here and size is a concern, then maybe we should change to `NonZeroU32` while we're at it? From memory, I don't think llvm or gcc use more than `u32` when emitting these. `addr2line` already does this conversion.

khuey:
I cared more about the representable but unused Some(0) case than the size. That's just a bonus.

Also registers for most people will be 64 bits these days so dropping to a u32 doesn't really save anything.

PR #554, opened 2021-04-29
read: replace Dwarf::debug_str_sup with Dwarf::sup

philipc:
This is simpler and allows us to avoid loading the supplementary
`.debug_str` twice.

Also split the supplementary section loading into `Dwarf::load_sup`,
which has the benefit that users don't have to pass a dummy section
loader when there is no supplementary object file.

### DIFF
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -1,4 +1,5 @@
 use alloc::string::String;
+use alloc::sync::Arc;
 
 use crate::common::{
     DebugAddrBase, DebugAddrIndex, DebugInfoOffset, DebugLineStrOffset, DebugLocListsBase,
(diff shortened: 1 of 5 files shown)


src/read/dwarf.rs
```diff
@@ -87,14 +89,25 @@ impl<T> Dwarf<T> {
...
     }
 
+    /// Load the DWARF sections from the supplementary object file.
+    ///
+    /// `section` operates the same as for `load`.
+    pub fn load_sup<F, E>(&mut self, section: F) -> core::result::Result<(), E>
+    where
+        F: FnMut(SectionId) -> core::result::Result<T, E>,
+    {
+        self.sup = Some(Arc::new(Self::load(section)?));
```
fitzgen, line 107: Should this assert that `self.sup.is_none()`? Or document that it will overwrite any already-loaded supplementary info? Not sure what the right choice is here. (But if we assert, that should also be documented.)

fitzgen (approved):
Nice clean up!

src/read/dwarf.rs
philipc: Sure, documented it.

PR #556, opened 2021-05-22
Add DebugLineStr::new.

khuey:
This seems to be missing for some reason.


philipc (approved):
This is fine, but typically you would use `Dwarf::load` or `Section::load` instead.

PR #557, opened 2021-06-04
Add `UnwindTable::into_current_row`

koute:
This will make it possible to drop the `UnwindTable` and return a reference to `UnwindTableRow` whose lifetime will be bound only to the lifetime of the `UninitializedUnwindContext`. This is still not perfect though since the `UnwindTable` constructor originally accepts the unwind context through a `&mut`, so you still won't be able to e.g. search through multiple sections and return a reference to `UnwindTableRow` for only one of them due to the limitations of Rust's current borrow checker without laundering the lifetime through `unsafe`, but it's better than what we have now.

(Some background on this change: I've been refactoring [this horrible method](https://github.com/koute/not-perf/blob/fcc8a5d558c2863a86568a05e13c82f97bf84c4a/nwind/src/frame_descriptions.rs#L356) from my profiler, and what I'd like to do is basically to look through multiple sections, and return the `UnwindTableRow` when I get a hit. I could just `clone` it, but it's over 6 kilobytes so if the compiler won't be able to elide that copy the performance will tank.)

An alternative to this would be to not store the `UnwindContext` inside of `UnwindTable` and just pass the unwind context manually each time we iterate over the rows instead of setting it once in the constructor. That would obviously be a more intrusive change than this one though, so I'm not sure if we actually want to go that way? (Especially since the new Polonius-based borrow checker fixes the issue I've described in the first paragraph.)

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -369,14 +369,14 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
     ///
     /// You must provide a function to get the associated CIE. See
     /// `PartialFrameDescriptionEntry::parse` for more information.
-    pub fn unwind_info_for_address<F>(
+    pub fn unwind_info_for_address<'ctx, F>(
         &self,
         frame: &EhFrame<R>,
         bases: &BaseAddresses,
-        ctx: &mut UninitializedUnwindContext<R>,
+        ctx: &'ctx mut UninitializedUnwindContext<R>,
         address: u64,
         get_cie: F,
-    ) -> Result<UnwindTableRow<R>>
+    ) -> Result<&'ctx UnwindTableRow<R>>
     where
         F: FnMut(
             &EhFrame<R>,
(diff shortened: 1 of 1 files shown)


philipc:
> An alternative to this would be to not store the UnwindContext inside of UnwindTable and just pass the unwind context manually each time we iterate over the rows instead of setting it once in the constructor.

Does this even avoid the lifetime problem? I experimented to get code like this in `unwind_info_for_address`:
```rust
        while let Some(row) = table.next_row(ctx)? {
            if row.contains(address) {
                return Ok(row);
            }
        }
```
but that gave a lifetime error.

I think the ideal API would be for `next_row` to return a reference with the context lifetime, but I'm not sure that's possible. Otherwise I think what you've done in this PR is fine. It would be good to change `unwind_info_for_address` to use this too though, since currently it is cloning the row.

koute:
> Does this even avoid the lifetime problem? I experimented to get code like this in `unwind_info_for_address`:

Currently it doesn't, mostly due to the borrow checker limitations that won't be fixed until Polonius lands. (If you turn on `-Zpolonius` it will compile.) But by splitting it like that you could call `ctx.row()` directly after advancing the row since the `ctx` wouldn't be stored inside of the `UnwindTable` anymore.

> I think the ideal API would be for `next_row` to return a reference with the context lifetime, but I'm not sure that's possible.

Yeah, I don't think that's currently possible.

> Otherwise I think what you've done in this PR is fine. It would be good to change `unwind_info_for_address` to use this too though, since currently it is cloning the row.

Sure, I'll change it.

koute:
Done.

philipc (approved):
Thanks!

PR #568, opened 2021-07-14
Fix ranges with the GNU DWO extension and other improvements

khuey:
This set of commits makes `dwarfdump` more useful for dumping a .dwo, by allowing me to provide the path of the "parent" binary, and fixes a nasty bug in the handling of ranges in the pre-DWARF5 GNU DWO extension where we weren't doing the behavior that's documented at https://github.com/bminor/binutils-gdb/blob/fac3b6a2e04f7feeefbf425c3798c34d134752d6/gdb/dwarf2/cu.h#L189

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -1757,7 +1757,7 @@ fn dump_loc_list<R: Reader, W: Write>(
     unit: &gimli::Unit<R>,
     dwarf: &gimli::Dwarf<R>,
 ) -> Result<()> {
-    let raw_locations = dwarf.locations.raw_locations(offset, unit.encoding())?;
+    let raw_locations = dwarf.raw_locations(unit, offset)?;
     let raw_locations: Vec<_> = raw_locations.collect()?;
     let mut locations = dwarf.locations(unit, offset)?;
     writeln!(
@@ -1885,7 +1885,7 @@ fn dump_range_list<R: Reader, W: Write>(
     unit: &gimli::Unit<R>,
     dwarf: &gimli::Dwarf<R>,
 ) -> Result<()> {
-    let raw_ranges = dwarf.ranges.raw_ranges(offset, unit.encoding())?;
+    let raw_ranges = dwarf.raw_ranges(unit, offset)?;
     let raw_ranges: Vec<_> = raw_ranges.collect()?;
     let mut ranges = dwarf.ranges(unit, offset)?;
     writeln!(
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -17,6 +17,7 @@ use std::iter::Iterator;
 use std::mem;
 use std::process;
 use std::result;
+use std::slice;
 use std::sync::{Condvar, Mutex};
 use typed_arena::Arena;
 
@@ -358,18 +359,9 @@ struct Flags {
     pubtypes: bool,
     aranges: bool,
     dwo: bool,
+    dwo_parent: Option<object::File<'static>>,
     raw: bool,
     match_units: Option<Regex>,
-}
-
-impl Flags {
-    fn section_name(&self, id: gimli::SectionId) -> Option<&'static str> {
-        if self.dwo {
-            id.dwo_name()
-        } else {
-            Some(id.name())
-        }
-    }
 }
 
 fn print_usage(opts: &getopts::Options) -> ! {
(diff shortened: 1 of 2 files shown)


src/read/dwarf.rs
```diff
@@ -297,6 +297,12 @@ impl<R: Reader> Dwarf<R> {
         unit: &Unit<R>,
         offset: RangeListsOffset<R::Offset>,
     ) -> Result<RngListIter<R>> {
+        let offset = if self.file_type == DwarfFileType::Dwo && unit.header.version() < 5 {
+            RangeListsOffset(offset.0.wrapping_add(unit.rnglists_base.0))
```
philipc, line 301: This means that `RangeListsOffset` has different meanings at different places in the code. While your changes look correct, it is hard to be sure, and ditto for users of this library. e.g. it looks like the `AttributeValue::RangeListsRef` support in `write/unit.rs` needs fixing too. I prefer if we introduce something like `RawRangeListsOffset` that is used in `AttributeValue`, and then convert that into a `RangeListsOffset`. Or do the conversion when parsing attributes but we may not have the required info there.

examples/dwarfdump.rs
```diff
@@ -647,7 +682,40 @@ where
...
+                .map(|unit_header| dwarf.unit(unit_header))
+                .filter_map(|unit| {
+                    let dwo_id = match unit.header.type_() {
+                        gimli::UnitType::Skeleton(dwo_id) => dwo_id,
+                        gimli::UnitType::Compilation => {
+                            let mut entries = unit.entries();
+                            let (_, this) = entries.next_dfs()?.unwrap();
+                            match this.attr_value(gimli::constants::DW_AT_GNU_dwo_id)? {
+                                Some(gimli::AttributeValue::DwoId(dwo_id)) => dwo_id,
+                                Some(_) => unreachable!(),
```
philipc, line 698: What makes this unreachable? (and same for the other use below)

examples/dwarfdump.rs
```diff
@@ -472,32 +470,43 @@ fn main() {
...
-                eprintln!("Failed to map file '{}': {}", sup_path, err);
+                eprintln!("Failed to map file '{}': {}", path, err);
                 process::exit(1);
             }
         };
-        match object::File::parse(&*sup_mmap) {
+        let mmap_ptr = mmap.as_ptr();
+        let mmap_len = mmap.len();
+        mem::forget(mmap);
+        match object::File::parse(unsafe { slice::from_raw_parts(mmap_ptr, mmap_len) }) {
```
philipc, line 491: Ok, but an arena would be nicer.

philipc (changes requested):
Good find for the bug fix.

It seems like the dwarfdump dwo changes include things that we eventually should support in the library itself somehow.

src/read/dwarf.rs
khuey: Yeah, the downside of a new type is that then everybody has to be aware of and handle this stuff that's only needed for a non-standard and obsolete extension.

I've never looked at the writing side of gimli but if it's at all possible I would just not support writing these non-standard GNU extensions.

examples/dwarfdump.rs
khuey: `dwoid!` only ever returns `AttributeValue::DwoId`

examples/dwarfdump.rs
khuey: idk what this means in practice.

khuey:
> It seems like the dwarfdump dwo changes include things that we eventually should support in the library itself somehow.

fwiw, in Pernosco when processing .dwos we `Dwarf::load` the .dwo file and then replace `debug_addr`, `debug_line`, `debug_line_str`, and `range_lists` (if and only if the .dwo does not have a `debug_rnglists`) with those sections from the parent. I feel like it's a bit difficult to push this into gimli since object loading lives entirely outside it. I suppose you could have a `Dwarf::load` variant that takes two functions?

examples/dwarfdump.rs
philipc: Right, but then it falls through to `self.value.clone()` for other forms.

src/read/dwarf.rs
philipc: That's a valid concern, but if it's important enough for you to handle it, then maybe others want to as well? It is uglier to use, but see what you think of https://github.com/philipc/gimli/commit/473e138d736bb6ed0b2515a5160e6f3385b99ac3

On the writing side, the only support required is for reading it when converting. It won't write it out in the same form.

examples/dwarfdump.rs
philipc: See https://github.com/philipc/gimli/commit/e6e7557a1198983494617f984ebf803af2d660da

philipc:
> I suppose you could have a Dwarf::load variant that takes two functions?

Or something you call after `Dwarf::load` that does the replacements, similar to `load_sup`. Not something that needs changing in this PR though.

philipc (approved):

PR #569, opened 2021-07-16
DWO cleanups

philipc:
Follow up to #568 
r? @khuey

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -613,23 +613,7 @@ where
             dwo_parent_units = match dwo_parent
                 .units()
                 .map(|unit_header| dwo_parent.unit(unit_header))
-                .filter_map(|unit| {
-                    let dwo_id = match unit.header.type_() {
-                        gimli::UnitType::Skeleton(dwo_id) => dwo_id,
-                        gimli::UnitType::Compilation => {
-                            let mut entries = unit.entries();
-                            let (_, this) = entries.next_dfs()?.unwrap();
-                            match this.attr_value(gimli::constants::DW_AT_GNU_dwo_id)? {
-                                Some(gimli::AttributeValue::DwoId(dwo_id)) => dwo_id,
-                                Some(_) => unreachable!(),
-                                None => return Ok(None),
-                            }
-                        }
-                        _ => return Ok(None),
-                    };
-
-                    Ok(Some((dwo_id, unit)))
-                })
+                .filter_map(|unit| Ok(unit.dwo_id.map(|dwo_id| (dwo_id, unit))))
                 .collect()
             {
                 Ok(units) => units,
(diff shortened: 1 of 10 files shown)


examples/dwarfdump.rs
```diff
@@ -602,32 +592,40 @@ where
...
-            });
-            dwarf.debug_addr = gimli::Section::load(load_dwo_parent_section.as_mut().unwrap())?;
-            dwarf.ranges = gimli::RangeLists::new(
-                gimli::Section::load(load_dwo_parent_section.as_mut().unwrap())?,
-                gimli::Section::load(load_dwo_parent_section.as_mut().unwrap())?,
-            );
+            };
+            let dwo_parent = gimli::Dwarf::load(&mut load_dwo_parent_section)?;
+            dwarf.debug_addr = dwo_parent.debug_addr.clone();
+            dwarf.ranges = dwo_parent.ranges.clone();
```
khuey, line 612: debug_line/debug_line_str really should be copied from the parent, and ranges should be copied if and only if the dwo doesn't have debug_rnglists (in other words if this is the GNU split dwarf extension and not the standard DWARF 5 version)

src/read/dwarf.rs
```diff
@@ -671,6 +684,11 @@ impl<R: Reader> Unit<R> {
                             unit.rnglists_base = base;
                         }
                     }
+                    constants::DW_AT_GNU_dwo_id => {
+                        if let AttributeValue::DwoId(dwo_id) = attr.value() {
+                            unit.dwo_id = Some(dwo_id);
```
khuey, line 689: Should we check to see if this is already present? DW_AT_GNU_dwo_id shouldn't really be present in DWARF 5 programs.

src/common.rs
```diff
@@ -136,6 +136,14 @@ pub struct DebugMacinfoOffset<T = usize>(pub T);
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct DebugMacroOffset<T = usize>(pub T);
 
+/// An offset into either the `.debug_ranges` section or the `.debug_rnglists` section,
+/// depending on the version of the unit the offset was contained in.
+///
+/// If this is a from a DWARF 4 DWO file, then it must additionally be offset by the
+/// value of `DW_AT_GNU_ranges_base`. You can use `Unit::ranges_offset_from_raw` to do this.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct RawRangeListsOffset<T = usize>(pub T);
```
khuey, line 145: My only concern about this is that it might be confused as being related to `raw_ranges`, but I don't have a better suggestion for the name.

khuey (approved):
generally lgtm

examples/dwarfdump.rs
philipc: Ok. This isn't something I changed in this PR is it? This sounds like something to be addressed as part of adding a method to `Dwarf` to do this.

src/read/dwarf.rs
philipc: Ok. It shouldn't matter, but it doesn't take much to add the check.

examples/dwarfdump.rs
khuey: Right, nothing changed here afaict.

khuey:
Can we (you :) ) do a 0.25 with this?

philipc:
I will in a few days after an `object` release.

philipc:
0.25 is released.

Issue #374, opened 2019-02-01
Add operations for mutate write::DebuggingInformationEntry


yurydelendik:
Are there operations for remove the child DIE or attributes? It will be useful to have such operations to mutate in-memory UnitTable forest.

philipc:
Not yet, but we definitely want them.

For DIEs, removing them just needs to update the `children` field of the parent, without actually deleting the child from the entries `Vec`. Attributes can simply be deleted.

One catch with that will be if a deleted attribute referenced something in another table (strings or range lists), then we'll still write those values out (deleting them from the other tables isn't easy). I don't have a solution for that yet.

PR #570, opened 2021-07-23
Add `delete_child` method to `write::DebuggingInformationEntry`

ayrtonm:
This adds a method for deleting DIEs by removing the ID from the `children` Vec as suggested in issue #374.

### DIFF
--- a/src/write/unit.rs
+++ b/src/write/unit.rs
@@ -545,6 +545,15 @@ impl DebuggingInformationEntry {
         self.children.iter()
     }
 
+    /// Delete a child entry and all of its children.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `id` is invalid.
+    pub fn delete_child(&mut self, id: UnitEntryId) {
+        self.children.retain(|&child| child != id);
+    }
+
     /// Return the type abbreviation for this DIE.
     fn abbreviation(&self, encoding: Encoding) -> Result<Abbreviation> {
         let mut attrs = Vec::new();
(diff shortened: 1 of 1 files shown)


src/write/unit.rs
```diff
@@ -545,6 +545,15 @@ impl DebuggingInformationEntry {
         self.children.iter()
     }
 
+    /// Delete a child entry and all of its children.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `id` is invalid.
```
philipc, line 552: I don't think this panics.

src/write/unit.rs
```diff
@@ -2809,6 +2818,18 @@ mod tests {
...
+            let child4 = add_child(unit, root, constants::DW_TAG_subprogram, "child4");
+
+            // This will not be deleted to check that deleting interleaved children works
+            add_child(unit, root, constants::DW_TAG_subprogram, "child5");
+
+            let child6 = add_child(unit, root, constants::DW_TAG_subprogram, "child6");
+            add_child(unit, child6, constants::DW_TAG_variable, "grandchild6");
+            unit.get_mut(root).delete_child(child4);
+            // Deleting `child6` deletes `grandchild6` as well
+            unit.get_mut(root).delete_child(child6);
```
philipc, line 2832: Please don't add it in this test. This test is specifically for checking the sibling attribute, and adding this makes it too hard to follow what's going on.

I'd even be fine without a test for this given the simplicity of the code it is testing.

philipc (changes requested):

src/write/unit.rs
ayrtonm: Oh that's right, I forgot to update the comment after changing the implementation. I'll fix it.

src/write/unit.rs
ayrtonm: That's fair, I can add a new test then. I'm hesitant to add this without a test since my previous implementation had a subtle bug when interleaving adds and deletes.

philipc (approved):

philipc:
Thanks!

PR #571, opened 2021-07-26
Release 0.25.0

philipc:


philipc:
fuzz failures seem like an upstream problem: https://github.com/rust-fuzz/cargo-fuzz/issues/270

PR #574, opened 2021-08-10
Add missing ARM and AArch64 register numbers

ayrtonm:
Register numbers were taked from [this table](https://developer.arm.com/documentation/ihi0040/c/?lang=en#dwarf-register-names) which is an updated version of the pdf linked in the comment on the `ARM` struct. For registers D0-D15 and S0-S31 I used the 2-byte register numbers (256-271) as recommended in the notes below the table. Left the TODO comment as some registers still need to be added.


src/arch.rs
```diff
@@ -57,6 +57,116 @@ registers!(Arm, {
...
+    WCGR0 = (104, "wCGR0"),
+    WCGR1 = (105, "wCGR1"),
+    WCGR2 = (106, "wCGR2"),
+    WCGR3 = (107, "wCGR3"),
+    WCGR4 = (108, "wCGR4"),
+    WCGR5 = (109, "wCGR5"),
+    WCGR6 = (110, "wCGR6"),
+    WCGR7 = (111, "wCGR7"),
+
+    ACC0 = (104, "ACC0"),
```
philipc, line 70: These overlapping ranges give compilation warnings.

Probably what we need to do as add a `registers_alias!` macro for the aliases. So then `Arm::register_name` will only give the primary name, but we still define constants for the aliases.

philipc (changes requested):
Can you also update the comment to `https://developer.arm.com/documentation/ihi0040/c/` please.

src/arch.rs
ayrtonm: Oh that's true, I only did `cargo doc` and missed that. That sounds reasonable. I guess D0-D15 should be the primary names for 256-271 and wCGR0-wCGR7 for 104-111.

ayrtonm: I ended up reusing the `registers` macro so `name_to_register` will work if given the name of an aliasing register. I also updated the comment on `register_name`.

src/arch.rs
```diff
@@ -35,7 +44,7 @@ macro_rules! registers {
 
 /// ARM architecture specific definitions.
 ///
-/// See [DWARF for the ARM Architecture](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0040b/IHI0040B_aadwarf.pdf).
+/// See [DWARF for the ARM Architecture](https://developer.arm.com/documentation/ihi0040/b/).
```
philipc, line 47:
```suggestion
/// See [DWARF for the ARM Architecture](https://developer.arm.com/documentation/ihi0040/c/).
```

src/arch.rs
```diff
@@ -57,6 +66,192 @@ registers!(Arm, {
...
+    S23 = (267, "S23"),
+    S24 = (268, "S24"),
+    S25 = (268, "S25"),
+    S26 = (269, "S26"),
+    S27 = (269, "S27"),
+    S28 = (270, "S28"),
+    S29 = (270, "S29"),
+    S30 = (271, "S30"),
+    S31 = (271, "S31"),
+
```
philipc, line 180:
```suggestion
```

philipc (changes requested):

philipc (approved):
Thanks!

PR #577, opened 2021-08-12
Finish adding ARM register numbers and aliases

ayrtonm:
Added the remaining ARM registers and aliases for R13-R15 listed [here](https://developer.arm.com/documentation/ihi0040/c/).

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -48,7 +48,6 @@ macro_rules! registers {
 #[derive(Debug, Clone, Copy)]
 pub struct Arm;
 
-// TODO: add more registers.
 registers!(Arm, {
     R0 = (0, "R0"),
     R1 = (1, "R1"),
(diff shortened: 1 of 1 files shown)


src/arch.rs
```diff
@@ -136,6 +170,10 @@ registers!(Arm, {
     D31 = (287, "D31"),
 },
 aliases {
+    SP = (13, "SP"),
+    LR = (14, "LR"),
+    PC = (15, "PC"),
```
philipc, line 175: I'm not familiar with ARM, but would it make sense for these to be the primary names? binutils objdump defaults to `-M reg-names-std` which uses these.

ayrtonm: I considered making them primary names since they're more common, but I wasn't sure if I should change the behavior of `register_name` so I made them aliases. I slightly prefer R13-R15 as aliases but either's fine with me.

philipc: Going to leave it since I'm not sure if this constitutes a breaking change and I probably want to do a minor release soon.

philipc (approved):

Issue #581, opened 2021-08-27
Use `gimli` without `alloc`


nbdd0121:
I am working on a pure Rust stack unwinder (https://github.com/nbdd0121/unwind). It's now fully functional on hosted platforms (with libc dependency), and my intention is for it to work on baremetal `#![no_std]` binaries as well. It's sort of working right now, but requires `alloc`, which is a big issue because allocator isn't always available or desirable for baremetal programs.

Currently my unwind crate uses two things that require allocation:
* `UninitializedUnwindContext` and
* `Evaluation`

One way to avoid having them performing allocation themselves is to ask the user to supply the backing memory. This creates usability issues for users that can use allocator, so I think an approach similar to [`managed` crate](https://docs.rs/managed/0.8.0/managed/) could be ideal. This approach is used by [`smoltcp` crate](https://docs.rs/smoltcp/0.7.5/smoltcp/) to avoid implicit allocations.

PR #582, opened 2021-08-28
Implement ArrayVec and use it for RegisterRuleMap and UnwindContext

nbdd0121:
Essentially undoes #497, but provide an in-crate implementation of a minimal ArrayVec instead of using a dependency.

It should aid readability and keeping all `unsafe` in one place should make auditing easier.

### DIFF
--- a/src/read/util.rs
+++ b/src/read/util.rs
@@ -1,0 +1,158 @@
+use core::fmt;
+use core::mem::MaybeUninit;
+use core::ops;
+use core::ptr;
+use core::slice;
+
+// Use a helper trait since const generics can't be used due to MSRV.
+// SAFETY: Implementer must not modify the content in storage.
+pub(crate) unsafe trait Array {
+    type Item;
+    type Storage;
+
+    fn new_storage() -> Self::Storage;
+    fn as_slice(storage: &Self::Storage) -> &[MaybeUninit<Self::Item>];
+    fn as_mut_slice(storage: &mut Self::Storage) -> &mut [MaybeUninit<Self::Item>];
+}
+
+macro_rules! impl_array {
+    () => {};
+    ($n:literal $($rest:tt)*) => {
+        // SAFETY: does not modify the content in storage.
+        unsafe impl<T> Array for [T; $n] {
+            type Item = T;
+            type Storage = [MaybeUninit<T>; $n];
+
+            fn new_storage() -> Self::Storage {
+                // SAFETY: An uninitialized `[MaybeUninit<_>; _]` is valid.
+                unsafe { MaybeUninit::uninit().assume_init() }
+            }
+
+            fn as_slice(storage: &Self::Storage) -> &[MaybeUninit<T>] {
+                storage
+            }
+
+            fn as_mut_slice(storage: &mut Self::Storage) -> &mut [MaybeUninit<T>] {
+                storage
+            }
+        }
+        impl_array!($($rest)*);
+    }
+}
+
+impl_array!(4 192);
+
+pub(crate) struct ArrayVec<A: Array> {
+    storage: A::Storage,
+    len: usize,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct CapacityFull;
+
+impl<A: Array> ArrayVec<A> {
+    pub fn new() -> Self {
+        Self {
+            storage: A::new_storage(),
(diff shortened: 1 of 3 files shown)


src/read/util.rs
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    pub fn swap_remove(&mut self, index: usize) -> A::Item {
+        // Bounds check.
+        let _ = &self[index];
+        // SAFETY: replace `self[index]` with the last element.
+        // This works even if `self[index]` is the last element.
+        unsafe {
+            let last = ptr::read(self.as_ptr().add(self.len - 1));
```
philipc, line 107: What if `self.len == 0`? I'd feel more comfortable if this function was implemented as `swap` + `pop`.

src/read/util.rs
```diff
@@ -0,0 +1,174 @@
...
+                storage
+            }
+        }
+        impl_array!($($rest)*);
+    }
+}
+
+impl_array!(4 192);
+
+pub struct ArrayVec<A: Array> {
```
philipc, line 42: Use `pub(crate)` for top-level items to make it clearer that we never re-export anything here.

philipc (commented):
I'm happy to merge this because I think this is heading in the right direction, but do you think we'll need to make this `ArrayVec` implementation public in order for you to choose between `Vec` and `ArrayVec` for #581? If so, maybe we can revert back to the `ArrayVec` crate as a dependency instead, but make it optional and change the default here to `Vec`, since I think that's actually a better choice for many users.

src/read/util.rs
```diff
@@ -0,0 +1,174 @@
+use core::fmt;
+use core::mem::MaybeUninit;
+use core::ops;
+use core::ptr;
+use core::slice;
+
+// Use a helper trait since const generics can't be used due to MSRV.
+// Implementer must ensure that `Storage` is `[MaybeUninit<Item>; CAPACITY]`
```
bjorn3, line 8: Please use

src/read/util.rs
nbdd0121: This is how std implements `Vec::swap_remove`.

nbdd0121:
> I'm happy to merge this because I think this is heading in the right direction, but do you think we'll need to make this `ArrayVec` implementation public in order for you to choose between `Vec` and `ArrayVec` for #581?

I am still experimenting different designs and I am not so sure about whether this should be public. I feel that this is a worthy change about readability/maintainability so I send a PR anyway.

> If so, maybe we can revert back to the `ArrayVec` crate as a dependency instead, but make it optional and change the default here to `Vec`, since I think that's actually a better choice for many users.

Wouldn't that break signal safety guarantee?

src/read/util.rs
nbdd0121: Added a comment explaining `0 <= index < len`

nbdd0121: Switched to `swap_remove` anyway because it removes the usage of `unsafe`.

src/read/util.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<A: Array> ops::Deref for ArrayVec<A> {
+    type Target = [A::Item];
+
+    fn deref(&self) -> &[A::Item] {
+        let slice = &A::as_slice(&self.storage)[..self.len];
```
philipc, line 113: We didn't previously bounds check this in `RegisterRuleMap::rules` or `RegisterRuleMap::rules_mut`. There is a performance regression for the benchmark `cfi::iterate_entries_evaluate_every_fde`, and removing this and the bounds check in `deref_mut` seems to fix that.

philipc:
> Wouldn't that break signal safety guarantee?

The default behaviour would, but users could use `ArrayVec` if they need that. I suspect that most current users of gimli don't need signal safety (the only other unwind implementation I'm aware of is unmaintained). There is already a problem for `Evaluation`, for which we don't support signal safety at all.

src/read/util.rs
philipc: What's the status of this PR? Do you think this bounds check is something that can be changed?

nbdd0121: Sorry I missed the comment. Will fix and push soon.

philipc (approved):
Thanks!

PR #592, opened 2021-09-17
Add EntriesRaw::skip_attributes.

### LINKED ISSUE #591: AttributeSpecification::size may be wrong for DW_FORM_ref_addr in version 2 if word size and address size differ
Compare:

https://github.com/gimli-rs/gimli/blob/d20becf924f623a501f24738c35558e6222e6946/src/read/unit.rs#L2102-L2110

and

https://github.com/gimli-rs/gimli/blob/d20becf924f623a501f24738c35558e6222e6946/src/read/abbrev.rs#L448-L451

I think `AttributeSpecification::size` needs to check the encoding version and use `address_size` for the size of `DW_FORM_ref_addr` if the version is 2.

mstange:
This makes it easier for consumers to skip over abbrevs that
they don't need. It might also be slightly faster than parsing
the unneeded attributes and throwing away their values, but
any speed benefits are unproven so far.

The performance of this is dominated by memory accesses: Either
reading the bytes from main memory into the cache, or even paging
the file contents in from disk. The skipping in skip_attributes
only saves memory accesses for fixed-size attributes, but variable-
length attributes are so common that we're never able to skip an
entire cache line.

This was an experiment to see if it could help addr2line performance.
It doesn't seem like this is the case, so this PR can be discarded. I'm
mostly putting it up for my own future reference.

This patch also fixes #591.


src/read/unit.rs
```diff
@@ -2208,6 +2209,77 @@ pub(crate) fn parse_attribute<'unit, R: Reader>(
...
+                constants::DW_FORM_udata
+                | constants::DW_FORM_ref_udata
+                | constants::DW_FORM_strx
+                | constants::DW_FORM_GNU_str_index
+                | constants::DW_FORM_addrx
+                | constants::DW_FORM_GNU_addr_index
+                | constants::DW_FORM_loclistx
+                | constants::DW_FORM_rnglistx => {
+                    let _ = input.read_uleb128()?;
+                }
```
philipc, line 2268: Not sure how well this optimises, maybe a `skip_uleb128` can be more efficient.

src/read/unit.rs
```diff
@@ -2208,6 +2209,77 @@ pub(crate) fn parse_attribute<'unit, R: Reader>(
     }
 }
 
+#[inline(always)]
```
philipc, line 2212: This is probably not appropriate here.

philipc (commented):
I'll experiment with this when I get time.

philipc:
I tried this out on `addr2line`, and it shows good improvements for its benchmarks, so I think this is worth merging after the previous comments have been addressed.

PR #594, opened 2021-10-07
Optimize for case where CIE contains 0 or 1 register rules

nbdd0121:
In most cases CIE contains no register rules at all, so the initial rules are default and don't have to be saved. (libgcc's unwinder even just assumes this is always the case and always restore undefined).


philipc:
I haven't reviewed closely, but I think you need to demonstrate that this is beneficial. The `cfi::iterate_entries_evaluate_every_fde` benchmark is slower.

PR #595, opened 2021-10-10
Allow custom storage for UnwindContext and Evaluation

nbdd0121:
Partially addresses #581 (alloc is still a dep, but everything unwind related doesn't require heap).

New traits `UnwindContextStorage` and `EvaluationStorage` is introduced to allow customisation of what the storage is. `UnwindContext`, `UnwindTable` and `Evaluation` will gain 1 additional generic parameter which is defaulted to `StoreOnHeap`.

The ctor with this generic param is called `new_in`, and the old `new` method will use the default param so type inference won't fail.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -1,7 +1,7 @@
 use alloc::vec::Vec;
 
 use core::cmp::{Ord, Ordering};
-use core::fmt::Debug;
+use core::fmt::{self, Debug};
 use core::iter::FromIterator;
 use core::mem;
 use core::num::Wrapping;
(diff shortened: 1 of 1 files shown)


philipc:
Once we do use const generics, would `ArrayVec<[T; N]>` become `ArrayVec<T, N>`? Given that, it doesn't make sense to have `ArrayVec<Vec<T>>`. I realize it will be more boilerplate, but what do you think of leaving `Array` and `ArrayVec` how they are, and adding a new `ArrayLike` that is implemented for `ArrayVec` and `Vec`, so that `Vec` can be used directly instead of via `ArrayVec`?

nbdd0121:
Wouldn't it require exposing ArrayVec to the user?

nbdd0121:
I think we can still do `ArrayVec<[T; N]>` with const generics; we don't need the macro anymore and can just do `impl<T, const N: usize> ArrayLike for [T; N]> {...}`

philipc:
Ok, that makes sense. So the comment on `ArrayLike` is wrong: the trait isn't needed due to lack of const generics, only the macro.

PR #596, opened 2021-10-12
Add read-core feature gate

### LINKED ISSUE #581: Use `gimli` without `alloc`

nbdd0121:
Fix #581

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -11,9 +11,9 @@ use super::util::{ArrayLike, ArrayVec};
 use crate::common::{DebugFrameOffset, EhFrameOffset, Encoding, Format, Register, SectionId};
 use crate::constants::{self, DwEhPe};
 use crate::endianity::Endianity;
-#[cfg(feature = "read")]
-use crate::read::StoreOnHeap;
-use crate::read::{EndianSlice, Error, Expression, Reader, ReaderOffset, Result, Section};
+use crate::read::{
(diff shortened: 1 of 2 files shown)


nbdd0121:
Bisection shows that it's changed in rust-lang/rust#76010, which fixes rust-lang/rust#75930. I could manually implement these traits, so perhaps we can bump MSRV to 1.48.

nbdd0121:
It seems that the checking is lazy if the bound depends on generic parameter:
```rust
trait Bar<U> {}

struct Foo<U, T: Bar<U> = ()>(U, T);
//^ this is lazy, since () may implement Bar<U> for some U, but not necessary all Us
```
and this is eagerly checked:
```rust
trait Bar {}

struct Foo<T: Bar = ()>(T);
//^ this is eager, so will raise an error
```

PR #603, opened 2021-11-03
Avoid use of unsafe for Vec storage

philipc:
This change is intended to be transparent to users. However, I did have to add a `T: Clone` bound for implementations of `ArrayLike`.

r? @nbdd0121


philipc:
I don't know how to fix the 1.42 build, so might have to leave this for later.

PR #604, opened 2021-11-09
Dedupe comp-dir/directory in line-table paths

Swatinem:
The same as https://github.com/gimli-rs/addr2line/pull/239, except updating the `simple_line` example.

Also includes a driveby fix for an unused import.


philipc (approved):

philipc:
Related `dwarfdump` fixes in #606.

PR #610, opened 2022-01-18
Some debug_frame -> eh_frame comment fixes.

mstange:
I think these were missed when this code was copied from the
debug_frame implementation to the eh_frame implementation.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -423,7 +423,7 @@ where
     Endian: Endianity,
 {
     /// Construct a new `EhFrame` instance from the data in the
-    /// `.debug_frame` section.
+    /// `.eh_frame` section.
     ///
     /// It is the caller's responsibility to read the section and present it as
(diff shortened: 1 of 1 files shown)


mstange:
The build failure seems unrelated to this PR. It complains about something in the indexmap dependency.

```
error[E0658]: use of unstable library feature 'vec_drain_as_slice': recently added
    --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/indexmap-1.8.0/src/map.rs:1182:30
```

philipc:
Yeah the failure is because indexmap bumps MSRV in a minor version. `object` had the same problem (https://github.com/gimli-rs/object/commit/4cb177ee948222eb30f5713b51bb527b8a504a7d) but I'm not sure that is the ideal fix... instead we want to test that it's possible to build with the MSRV, but not require that all dependencies are the latest version when doing so, but I don't know how to test that.

philipc (approved):

PR #618, opened 2022-04-07
Support DWARF2 for DW_OP_implicit_pointer

crzysdrs:
In DwarfV2, the address length encoding in DW_OP_implicit_pointer is the encoding address size, not the format word size.

[libdwarf/dwarf_loc#L539](https://github.com/avast/libdwarf/blob/7d4f63f44d44386744628ee1bca84afc43a56583/libdwarf/libdwarf/dwarf_loc.c#L539)

### DIFF
--- a/src/read/op.rs
+++ b/src/read/op.rs
@@ -705,10 +705,12 @@ where
             }
             constants::DW_OP_stack_value => Ok(Operation::StackValue),
             constants::DW_OP_implicit_pointer | constants::DW_OP_GNU_implicit_pointer => {
-                let value = if encoding.version != 2 {
-                    bytes.read_offset(encoding.format)?
+                let value = if encoding.version == 2 {
+                    bytes
+                        .read_address(encoding.address_size)
+                        .and_then(Offset::from_u64)?
                 } else {
-                    Offset::from_u64(bytes.read_address(encoding.address_size)?)?
+                    bytes.read_offset(encoding.format)?
                 };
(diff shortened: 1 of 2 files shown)


src/read/op.rs
```diff
@@ -705,7 +705,11 @@ where
             }
             constants::DW_OP_stack_value => Ok(Operation::StackValue),
             constants::DW_OP_implicit_pointer | constants::DW_OP_GNU_implicit_pointer => {
-                let value = bytes.read_offset(encoding.format)?;
+                let value = if encoding.version != 2 {
```
philipc, line 708: Please change the test to `==` and swap the branches. Same for the if expression for writing.

src/read/op.rs
```diff
@@ -705,7 +705,11 @@ where
...
+                    bytes.read_offset(encoding.format)?
+                } else {
+                    Offset::from_u64(bytes.read_address(encoding.address_size)?)?
```
philipc, line 711:
```suggestion
                    bytes.read_address(encoding.address_size).and_then(R::Offset::from_u64)?
```
Just for consistency with how the rest of the code is written.

philipc (changes requested):

src/read/op.rs
crzysdrs: Swapped

src/read/op.rs
crzysdrs: Adjusted for consistency.

philipc (approved):
Thanks!

PR #620, opened 2022-04-13
Replace fallible-iterator with tryiterator

khuey:
As we've discussed over email. Just looking for feedback here, I have to release tryiterator on crates.io before we merge this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ object = { version = "0.27.1", features = ["wasm"] }
 rayon = "1.0"
 regex = "1"
 test-assembler = "0.1.3"
+tryiterator = { git = "https://github.com/Pernosco/tryiterator" }
 typed-arena = "2"
 
 [features]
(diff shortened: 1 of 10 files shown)


philipc:
These changes are all look trivial, which is good.

However, it appears that something like `AttrsIter::len` will silently discard any errors. Do you have any way to address that? This was the original reason we started using `FallibleIterator`.

Have you sort feedback from the larger Rust community about `tryiterator`?

khuey:
> These changes are all look trivial, which is good.
> 
> However, it appears that something like `AttrsIter::count` will silently discard any errors. Do you have any way to address that? This was the original reason we started using `FallibleIterator`.

It would be straightforward to add a `TryIterator::count_ok`.

> Have you sort feedback from the larger Rust community about `tryiterator`?

I started to ask people for their opinions privately this week. I'm in no hurry to merge this.

philipc:
> It would be straightforward to add a TryIterator::count_ok.

Right, but it would need a lint to determine the cases where `count` was wrong.

khuey:
Another option that was suggested to me is to add an intentionally conflicting `TryIteratorExt::count` so that `iter.count()` won't work without qualifying which trait is being used.

PR #623, opened 2022-07-21
Shrink unsafe block

cactter:


src/read/util.rs
```diff
@@ -146,8 +146,8 @@ impl<A: ArrayLike> ArrayVec<A> {
         }
 
         // SAFETY: storage[index] is filled later.
+        let p = unsafe { storage.as_mut_ptr().add(index) };
         unsafe {
-            let p = storage.as_mut_ptr().add(index);
             core::ptr::copy(p as *const _, p.add(1), self.len - index);
```
philipc, line 151: I'm not convinced this is shrinking it in a meaningful way. It appears to me that all it has done is move a temporary assignment outside the unsafe block, while making the code harder to read.

PR #624, opened 2022-08-22
Add LoongArch register defs

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:
https://loongson.github.io/LoongArch-Documentation/README-EN.html


philipc:
I see the list of registers in the documentation, but I didn't see anywhere that documents that DWARF values for these registers.

zhaixiaojuan:
The mapping relationship between registers and DWARF is consistent with the definition in gcc.
https://github.com/gcc-mirror/gcc/blob/master/gcc/config/loongarch/loongarch.h#374


Loongarch registers | Loongarch dwarf2 regnum
-- | --
$r0-$r31 | 0-31
$f0-$f31 | 32-63
$fcc0-$fcc7 | 64-71

philipc:
Including `$` in the register names seems unusual to me. Does this mean that when using AT&T assembly syntax you end up referring to `$r1` as `%$r1`?

zhaixiaojuan:
The assembly format of LoongArch64 is different from that of AT&T and does not use %. 
The following is the assembly code example of LoongArch64:

  `addi.d  $sp, $sp, -16(0xff0)`
  `st.d  $fp, $sp, 8(0x8)`
  `addi.d   $fp, $sp, 16(0x10)`
  `pcaddu12i   $a0, 7(0x7)`
  `addi.d   $a0, $a0, 1772(0x6ec)`
  `pcaddu12i   $t0, 7(0x7)`

PR #627, opened 2022-09-15
Make Abbreviations refcounted and add `Unit::with_abbreviations` constructor

### LINKED ISSUE #626: Add a way to share parsed `Abbreviations` between `Unit`s
I think I’m hitting an interesting pathological case with a massive 5.5G DWARF file:

The file has ~280k Units, and ~1.5k abbrevs. It looks like those abbrevs are shared by all the units (they all have abbrev_offset = 0).

That means that `Unit::new` is re-parsing the same ~1.5k abbrevs over and over again, holding them in memory and freeing those in the end.

A way to share the `Abbreviations` between `Unit`s would be nice to avoid this duplicated work.

Swatinem:
`Abbreviations` are now reference counted internally making them cheap to `Clone`.

The new `Unit` constructor also allows parsing the `Abbreviations` independently from constructing the Unit. Combined with the reference counting, the `Abbreviations` can also be shared efficiently between `Unit`s.

This should fix #626.

### DIFF
--- a/src/read/abbrev.rs
+++ b/src/read/abbrev.rs
@@ -1,6 +1,7 @@
 //! Functions for parsing DWARF debugging abbreviations.
 
 use alloc::collections::btree_map;
+use alloc::sync::Arc;
 use alloc::vec::Vec;
 use core::convert::TryFrom;
 use core::fmt::{self, Debug};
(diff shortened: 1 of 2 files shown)


Swatinem:
One drawback is that there is no way to validate that the API user actually passed the correct `Abbreviations`.

I wonder if it would be enough to call that out in the docs to "be careful".

philipc:
Thanks for the PR. However, I prefer a solution where existing users get the benefit automatically. I think this is as simple as adding an `AbbreviationsCache` to `Dwarf`, and changing `Unit::new` to use it.

#628 has an initial implementation of this idea, but it doesn't work in `no_std` due to the use of `Mutex`. I'm not sure how to fix that, but first let me know if this approach is suitable for you.

philipc:
Merged #628 instead.

PR #630, opened 2022-10-26
skip tombstone when iter rangelist

### LINKED ISSUE #629: Invalid ranges read from wasm file
I run dwarfdump agains a wasm file, and get invalid address ranges compare to llvm's impl. 
Those entries with 0xfffffffe are invalid ones. Seems valid entries are mixed with invalid ones, is it something known?
```
UNIT<.debug_info+0x00000a0e>: length = 0x2cc2f, format = Dwarf32, version = 4, address_size = 4, abbrev_offset = 0x389
< 0><0x0000000b>  DW_TAG_compile_unit
                    DW_AT_producer              clang LLVM (rustc version 1.66.0-nightly (b8b5caee0 2022-10-16))
                    DW_AT_language              DW_LANG_Rust
                    DW_AT_name                  library/std/src/lib.rs/@/std.65018fcd-cgu.0
                    DW_AT_stmt_list             <.debug_line+0x0000043c>
                    DW_AT_comp_dir              /rustc/b8b5caee04116c7383eb1c6470fcf15c437a60d4
                    DW_AT_GNU_pubnames          yes
                    DW_AT_low_pc                0x00000000
                    DW_AT_ranges                <rnglist at .debug_ranges+0x0000fcc8 with 785 entries>
                        [ 0] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 1] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 2] <address pair low-off: 0x000002b9 addr 0x000002b9 high-off: 0x000002c6 addr 0x000002c6>
                        [ 3] <address pair low-off: 0x000002c7 addr 0x000002c7 high-off: 0x000002d4 addr 0x000002d4>
                        [ 4] <address pair low-off: 0x000002d5 addr 0x000002d5 high-off: 0x000002e1 addr 0x000002e1>
                        [ 5] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 6] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 7] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 8] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [ 9] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
                        [10] <address pair low-off: 0xfffffffe addr 0xfffffffe high-off: 0xfffffffe addr 0xfffffffe>
```

llvm dwarfdump's output

```
0x00000a0e: Compile Unit: length = 0x0002cc2f, format = DWARF32, version = 0x0004, abbr_offset = 0x0389, addr_size = 0x04 (next unit at 0x0002d641)

0x00000a19: DW_TAG_compile_unit
              DW_AT_producer    ("clang LLVM (rustc version 1.66.0-nightly (b8b5caee0 2022-10-16))")
              DW_AT_language    (DW_LANG_Rust)
              DW_AT_name        ("library/std/src/lib.rs/@/std.65018fcd-cgu.0")
              DW_AT_stmt_list   (0x0000043c)
              DW_AT_comp_dir    ("/rustc/b8b5caee04116c7383eb1c6470fcf15c437a60d4")
              DW_AT_GNU_pubnames        (true)
              DW_AT_low_pc      (0x00000000)
              DW_AT_ranges      (0x0000fcc8
                 [0x000002b9, 0x000002c6)
                 [0x000002c7, 0x000002d4)
                 [0x000002d5, 0x000002e1)
                 [0x000002e2, 0x000002f1)
                 [0x000002f2, 0x00000335)
                 [0x00000336, 0x0000034a)
                 [0x0000034c, 0x000004db)
                 [0x00000714, 0x00000859)
                 [0x00000944, 0x000009b5)
```

shuoli84:
fixes #629 
I add the check and skip logic between raw range => range, all tests passed, but dwarfdump tool is broken, because it expects one to one relationship between raw range and range. 
Any idea on how to fix this?

### DIFF
--- a/src/read/rnglists.rs
+++ b/src/read/rnglists.rs
@@ -483,18 +483,8 @@ impl<R: Reader> RngListIter<R> {
 
     /// Advance the iterator to the next range.
     pub fn next(&mut self) -> Result<Option<Range>> {
-        let tombstone = range_address_tombstone(self.raw.encoding.address_size);
-
-        macro_rules! try_tombstone {
-            ($e: expr) => {{
-                let val = $e;
-                if tombstone == val {
-                    continue;
-                } else {
-                    val
-                }
-            }};
-        }
+        let tombstone =
+            range_address_tombstone(self.raw.encoding.address_size, self.raw.encoding.version);
 
         loop {
             let raw_range = match self.raw.next()? {
@@ -504,44 +494,46 @@ impl<R: Reader> RngListIter<R> {
 
             let range = match raw_range {
                 RawRngListEntry::BaseAddress { addr } => {
-                    self.base_address = try_tombstone!(addr);
+                    self.base_address = addr;
                     continue;
                 }
                 RawRngListEntry::BaseAddressx { addr } => {
-                    self.base_address = try_tombstone!(self.get_address(addr)?);
+                    self.base_address = self.get_address(addr)?;
                     continue;
                 }
                 RawRngListEntry::StartxEndx { begin, end } => {
-                    let begin = try_tombstone!(self.get_address(begin)?);
-                    let end = try_tombstone!(self.get_address(end)?);
+                    let begin = self.get_address(begin)?;
+                    let end = self.get_address(end)?;
                     Range { begin, end }
                 }
                 RawRngListEntry::StartxLength { begin, length } => {
-                    let begin = try_tombstone!(self.get_address(begin)?);
+                    let begin = self.get_address(begin)?;
+                    if begin == tombstone {
+                        continue;
+                    }
                     let end = begin + length;
                     Range { begin, end }
                 }
                 RawRngListEntry::AddressOrOffsetPair { begin, end }
                 | RawRngListEntry::OffsetPair { begin, end } => {
-                    let mut range = Range {
-                        begin: try_tombstone!(begin),
-                        end: try_tombstone!(end),
-                    };
-                    range.add_base_address(
-                        try_tombstone!(self.base_address),
-                        self.raw.encoding.address_size,
-                    );
+                    if self.base_address == tombstone || begin == tombstone {
+                        continue;
+                    }
+
+                    let mut range = Range { begin, end };
+                    range.add_base_address(self.base_address, self.raw.encoding.address_size);
                     range
                 }
-                RawRngListEntry::StartEnd { begin, end } => Range {
-                    begin: try_tombstone!(begin),
-                    end: try_tombstone!(end),
-                },
+                RawRngListEntry::StartEnd { begin, end } => Range { begin, end },
                 RawRngListEntry::StartLength { begin, length } => Range {
-                    begin: try_tombstone!(begin),
+                    begin,
                     end: begin + length,
                 },
             };
+
+            if range.begin == tombstone {
+                continue;
+            }
 
             if range.begin > range.end {
                 self.raw.input.empty();
@@ -623,10 +615,15 @@ impl Range {
 }
 
 /// compute address tombstone for range
-const fn range_address_tombstone(address_size: u8) -> u64 {
-    // llvm's code use address_tombstone - 1 as range's tombstone
-    // for more info, check llvm's DWARFDebugRangeList::getAbsoluteRanges code
-    crate::compute_tombstone_address(address_size) - 1
+#[inline(always)]
+const fn range_address_tombstone(address_size: u8, version: u16) -> u64 {
+    if version == 4 {
+        // llvm's dwarf v4 code use address_tombstone - 1 as range's tombstone
+        // for more info, check llvm's DWARFDebugRangeList::getAbsoluteRanges code
+        crate::compute_tombstone_address(address_size) - 1
+    } else {
+        crate::compute_tombstone_address(address_size)
+    }
 }
 
 #[cfg(test)]
@@ -640,11 +637,12 @@ mod tests {
     #[test]
     fn test_rnglists_32() {
         const ADDRESS_SIZE: u8 = 4;
-        const TOMBSTONE: u64 = range_address_tombstone(ADDRESS_SIZE);
+        const VERSION: u16 = 5;
+        const TOMBSTONE: u64 = range_address_tombstone(ADDRESS_SIZE, VERSION);
 
         let encoding = Encoding {
             format: Format::Dwarf32,
-            version: 5,
+            version: VERSION,
             address_size: ADDRESS_SIZE,
         };
         let section = Section::with_endian(Endian::Little)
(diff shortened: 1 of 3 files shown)


shuoli84:
@philipc I fired a pr here, pls take a look?
There are different ways to solve the problem, my considerations:
1. add a flag on Range struct to indicate it is tombstone. But it will increase range size, which is over kill because tombstone is a rare case.
2. skip RawRange, which is wrong, RawRange is something raw I think?
3. keep RawRange, skips Range, which we have both the low level raw info and correct user facing behavior. 
So I implemented 3.

src/read/rnglists.rs
```diff
@@ -491,32 +504,41 @@ impl<R: Reader> RngListIter<R> {
 
             let range = match raw_range {
                 RawRngListEntry::BaseAddress { addr } => {
-                    self.base_address = addr;
+                    self.base_address = try_tombstone!(addr);
```
philipc, line 507: A tombstone for a base address must not be ignored. Instead, all offsets using that base address are also tombstones. That is, change `add_base_address` to handle this for base addresses.

src/read/rnglists.rs
```diff
@@ -491,32 +504,41 @@ impl<R: Reader> RngListIter<R> {
...
                 RawRngListEntry::BaseAddressx { addr } => {
-                    self.base_address = self.get_address(addr)?;
+                    self.base_address = try_tombstone!(self.get_address(addr)?);
                     continue;
                 }
                 RawRngListEntry::StartxEndx { begin, end } => {
-                    let begin = self.get_address(begin)?;
-                    let end = self.get_address(end)?;
+                    let begin = try_tombstone!(self.get_address(begin)?);
+                    let end = try_tombstone!(self.get_address(end)?);
```
philipc, line 516: Instead of checking this for each kind of entry, do a single check below at the same place we check `if range.begin > range.end`.

philipc (changes requested):
I've noted a couple of things that needed fixing. However, I'm not sure yet that this is the approach we want to take. Another option would be to add a method to `Range` that checks if it is a tombstone.

For dwarfdump, there already isn't a one to one correspondence between `RawRange` and `Range`, so this would become an additional check. Having it as a method on `Range` might make that easier to implement.

I would also like to match LLVM's handling for DWARF v5 at the same time, since this will eventually become the DWARF v6 support. Trying to implement that might give a better feel for what the best approach is.

shuoli84:
> I've noted a couple of things that needed fixing. However, I'm not sure yet that this is the approach we want to take. Another option would be to add a method to `Range` that checks if it is a tombstone.
>
> For dwarfdump, there already isn't a one to one correspondence between `RawRange` and `Range`, so this would become an additional check. Having it as a method on `Range` might make that easier to implement.

I think check against address is better compare to provide a method on `Range`, `Range` maybe created by Address calculation, if we didn't check against the Address, then tombstone + offset/length may produce valid address?
Or we do both, check against address, if it is tombstone, then create a Tombstone `Range`?

> I would also like to match LLVM's handling for DWARF v5 at the same time, since this will eventually become the DWARF v6 support. Trying to implement that might give a better feel for what the best approach is.

Do you mean how llvm handling tombstone address or something else? llvm skipped the tombstone range while parsing, I think it is same as what I did in this pr. ref: https://github.com/llvm/llvm-project/blob/2bdfececef4330b3a6489cdb67c57eb771d5f9e4/llvm/lib/DebugInfo/DWARF/DWARFDebugRangeList.cpp#L87

philipc:
> Range maybe created by Address calculation, if we didn't check against the Address, then tombstone + offset/length may produce valid address?

This should be handled in `add_base_address`.

> Do you mean how llvm handling tombstone address or something else?

I mean the DWARF v5 handling: https://github.com/llvm/llvm-project/blob/2bdfececef4330b3a6489cdb67c57eb771d5f9e4/llvm/lib/DebugInfo/DWARF/DWARFDebugRnglists.cpp#L95

shuoli84:
> > Range maybe created by Address calculation, if we didn't check against the Address, then tombstone + offset/length may produce valid address?
> 
> This should be handled in `add_base_address`.
> 
> > Do you mean how llvm handling tombstone address or something else?
> 
> I mean the DWARF v5 handling: https://github.com/llvm/llvm-project/blob/2bdfececef4330b3a6489cdb67c57eb771d5f9e4/llvm/lib/DebugInfo/DWARF/DWARFDebugRnglists.cpp#L95

Thanks, I didn't aware of this 🤣 .. So what I implemented is actually the dwarfv4 logic, and the link you provided is the v5 one? Is it a good idea to implement both? I think v4 is quite common.

philipc:
> Is it a good idea to implement both? I think v4 is quite common.

Yes, this crate aims to support both.

src/read/rnglists.rs
```diff
@@ -504,45 +494,47 @@ impl<R: Reader> RngListIter<R> {
...
                 RawRngListEntry::StartxEndx { begin, end } => {
-                    let begin = try_tombstone!(self.get_address(begin)?);
-                    let end = try_tombstone!(self.get_address(end)?);
+                    let begin = self.get_address(begin)?;
+                    let end = self.get_address(end)?;
                     Range { begin, end }
                 }
                 RawRngListEntry::StartxLength { begin, length } => {
-                    let begin = try_tombstone!(self.get_address(begin)?);
+                    let begin = self.get_address(begin)?;
```
shuoli84, line 510: check before begin + length, if begin is tombstone, then it will overflow

shuoli84:
@philipc I've updated the logic to handle both v4 and v5. 
Per dwarfdump. I think it is better to untie the `RawRange` and `Range`, maybe add a flag to control which set to show. I understand if add we add a method `is_tombstone` will make dwarfdump's logic easier, but that will leak the tombstone to end user. Also llvm's implementation just ignored the tombstone.

shuoli84:
I made a simple change to print ranges first, then raw ranges second, seems not bad.

```
                    DW_AT_ranges                <rnglist at .debug_ranges+0x00000078 with 2 entries>
                        ranges:
                        [ 0] [00000004, 0000013b)
                        [ 1] [0000013c, 00000194)
                        raw ranges:
                        [ 0] <address pair low-off: 0x00000004 high-off: 0x0000013b>
                        [ 1] <address pair low-off: 0x0000013c high-off: 0x00000194>
```

philipc:
Thanks for working on this. I'm going to fix some things and experiment with alternatives. I'll open a new PR and add you as coauthor.

PR #632, opened 2022-11-20
Fix atomic ordering in LazyArc

philipc:
Fixes build error:
error: compare_exchange's failure ordering may not be stronger than the success ordering of `Release`


bjorn3:
This has been allowed since https://github.com/rust-lang/rust/pull/98383 which landed in rustc 1.64. Maybe add a comment to that effect?

PR #635, opened 2023-01-06
Add support for XCOFF

bzEq:
XCOFF has different names for debug sections. This PR adds support for XCOFF which is the default object file format on AIX. Tests on AIX server all pass.
```
rg  '(Running)|(passed)' test.log 
2:     Running unittests src/lib.rs (target/debug/deps/gimli-f009eada21901b94)
464:test result: ok. 458 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.32s
466:     Running tests/convert_self.rs (target/debug/deps/convert_self-6975f7e844e5c08a)
472:test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.04s
474:     Running tests/parse_self.rs (target/debug/deps/parse_self-86a99c3c60cd3793)
488:test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.62s
490:     Running benches/bench.rs (target/debug/deps/bench-74d0874835d90a7c)
517:test result: ok. 23 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.88s
519:     Running unittests examples/dwarf-validate.rs (target/debug/examples/dwarf_validate-9cebfaaf145a12ef)
523:test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
525:     Running unittests examples/dwarfdump.rs (target/debug/examples/dwarfdump-dfec48f8bc4e68ab)
529:test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
531:     Running unittests examples/simple.rs (target/debug/examples/simple-bfa04b1b769853fc)
535:test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
537:     Running unittests examples/simple_line.rs (target/debug/examples/simple_line-826c47cfd577f13f)
541:test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -291,6 +291,7 @@ pub enum SectionId {
 
 impl SectionId {
     /// Returns the ELF section name for this kind.
+    #[cfg(not(target_os = "aix"))]
     pub fn name(self) -> &'static str {
         match self {
             SectionId::DebugAbbrev => ".debug_abbrev",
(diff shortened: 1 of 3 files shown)


src/common.rs
```diff
@@ -318,6 +319,35 @@ impl SectionId {
         }
     }
 
+    /// Returns the XCOFF section name for this kind.
+    #[cfg(target_os = "aix")]
```
philipc, line 323: This library supports parsing non-native binaries, so we can't change the meaning of a method like this (changing it from ELF name to XCOFF name based on target). Instead, we need to keep the `name` method unchanged, and add a new method such as `xcoff_name` that returns the XCOFF name.

src/common.rs
```diff
@@ -318,6 +319,35 @@ impl SectionId {
...
+    pub fn name(self) -> &'static str {
+        match self {
+            SectionId::DebugAbbrev => ".dwabrev",
+            SectionId::DebugAddr => panic!("XCOFF doesn't have debug address section"),
```
philipc, line 327: We can't panic like this. Instead, the `xcoff_name` method should return `Option<&'static str>`.

philipc (changes requested):

philipc:
We don't have file format specific tests currently, and don't really need to because gimli is meant to be independent of the object file format. We do test the dwarfdump example on linux and macos executables in github actions, but I doubt there is a way to do the same for AIX.

Before sending a PR for addr2line, it'd be better to get the dwarfdump example working in this crate too. Maybe it did with your original changes, but it won't now. We'll need to choose section names based on the file format.

philipc (approved):
Thanks!

philipc:
While I'm fine with adding `xcoff_name` as in this PR, I'm not sure that it's actually what we want to use in addr2line and dwarfdump. It would be better if the object crate handled this kind of logic, and `Object::section_by_name` currently already does that for Mach-O. I think it would be okay to change `section_by_name` for XCOFF too by doing string comparisons (and document this). Ideally we'd have something like `Object::dwarf_section_by_id`, but the problem with that is getting a common ID definition from somewhere, because gimli can't depend on object and vice versa.

PR #637, opened 2023-01-15
Add a couple helper methods for working with split DWARF

khuey:
These will be useful for addr2line.

### DIFF
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -996,6 +996,25 @@ impl<R: Reader> Unit<R> {
             self.rnglists_base = other.rnglists_base;
         }
     }
+
+    /// Find the dwo name (if any) for this unit, automatically handling the differences
+    /// between the standardized DWARF 5 split DWARF format and the pre-DWARF 5 GNU
+    /// extension.
+    ///
+    /// The returned value is relative to this unit's `comp_dir`.
+    pub fn dwo_name(&self) -> Result<Option<AttributeValue<R>>> {
+        let mut entries = self.entries();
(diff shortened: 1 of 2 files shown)


src/read/dwarf.rs
```diff
@@ -569,6 +569,20 @@ impl<R: Reader> Dwarf<R> {
         }
         err.description().into()
     }
+
+    /// Assuming `self` was loaded from a .dwo, take the appropriate
+    /// sections from `parent` (which contains the skeleton unit for this
+    /// dwo) and return a suitable instance of `Dwarf`.
+    pub fn make_dwo(mut self, parent: &Dwarf<R>) -> Dwarf<R> {
```
philipc, line 576:
```suggestion
    pub fn make_dwo(&mut self, parent: &Dwarf<R>) {
```

Also relax the requirement on `R` to be `R: Clone`.

philipc (changes requested):
Thanks!

Can you also change the dwarfdump example to use `make_dwo`.

philipc:
The coverage failure in CI can be ignored.

src/read/dwarf.rs
khuey: Done.

philipc: How about the suggested change to be `&mut`? Since we're only modifying a few fields, that makes more sense to me, but I'm open to being persuaded.

khuey: Ah I didn't actually look at the code :)

Sure, I don't care either way.

philipc (approved):

PR #649, opened 2023-05-03
read: add impl fallible-iterator for OperationIter

tamird:


tamird:
Looks like CI is failing because `oxide_miniz` is using [`slice::fill`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.fill) and [`u32::MAX`](https://doc.rust-lang.org/stable/std/primitive.u32.html#associatedconstant.MAX) which were stabilized in 1.50.0 and 1.43.0 respectively.

It also seems to be using a loop in a constant context which was stabilized in 1.46.0 [citation](https://github.com/rust-lang/rust/blob/9b99ff76b28605ad2e2c49e369721ae400e6a665/compiler/rustc_feature/src/accepted.rs#L123C1-L124).

tamird:
Hopefully https://github.com/gimli-rs/gimli/pull/651 fixes the build.

philipc (approved):

PR #650, opened 2023-05-03
Use cargo-hack to test all feature combinations

tamird:
See individual commits.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -10,11 +10,11 @@ jobs:
   build:
     strategy:
       matrix:
-        os: ["ubuntu-latest"]
-        rust_channel: ["stable", "beta", "nightly", "1.42.0"]
+        os: ['ubuntu-latest']
+        rust_channel: ['stable', 'beta', 'nightly', '1.42.0']
         include:
-          - rust_channel: "stable"
-            os: "macOS-latest"
+          - rust_channel: stable
+            os: macOS-latest
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v3
@@ -61,7 +61,7 @@ jobs:
           rustup install nightly
           rustup default nightly
       - name: Install `cargo fuzz`
-        run: cargo install cargo-fuzz --vers "^0.11.0"
+        run: cargo install cargo-fuzz --vers '^0.11.0'
       - run: cargo fuzz build -Oa
       - uses: actions/upload-artifact@v3
         with:
@@ -75,8 +75,8 @@ jobs:
   run_fuzz_targets:
     strategy:
       matrix:
-        fuzz_target: ["debug_abbrev", "debug_aranges", "debug_info", "debug_line", "eh_frame", "eh_frame_hdr"]
-    name: "Run `${{matrix.fuzz_target}}` fuzz target"
+        fuzz_target: ['debug_abbrev', 'debug_aranges', 'debug_info', 'debug_line', 'eh_frame', 'eh_frame_hdr']
+    name: Run `${{matrix.fuzz_target}}` fuzz target
     needs: build_fuzz_targets
     runs-on: ubuntu-latest
     steps:
@@ -90,7 +90,7 @@ jobs:
         with:
           name: fuzz-targets
         # Note: -max_total_time=300 == 300 seconds == 5 minutes.
-      - name: "Run `${{matrix.fuzz_target}}` fuzz target"
+      - name: Run `${{matrix.fuzz_target}}` fuzz target
         run: |
           mkdir ${{matrix.fuzz_target}}_artifacts
           chmod +x ./fuzz-targets/${{matrix.fuzz_target}}
@@ -137,9 +137,9 @@ jobs:
       matrix:
         target:
           # A 32-bit target.
-          - "i686-unknown-linux-gnu"
+          - i686-unknown-linux-gnu
           # A big-endian target
-          - "mips64-unknown-linux-gnuabi64"
+          - mips64-unknown-linux-gnuabi64
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -181,7 +181,7 @@ jobs:
         uses: coverallsapp/github-action@master
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          path-to-lcov: './lcov.info'
+          path-to-lcov: ./lcov.info
 
   doc:
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,12 +18,10 @@ jobs:
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v3
-    - name: Install rustup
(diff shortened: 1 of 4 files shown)


philipc (changes requested):
Please don't use actions-rs. It is unmaintained.

tamird:
Thanks, I didn't realize that. Done.

.github/workflows/rust.yml
```diff
@@ -108,26 +104,28 @@ jobs:
...
+        with:
+          toolchain: stable
+      - uses: taiki-e/install-action@v2
+        with:
+        tool: cargo-hack
+      # Ensure gimli can be built with all feature combinations.
+      - run: cargo hack check --feature-powerset --exclude-features rustc-dep-of-std
+      # Ensure gimli tests pass with all feature combinations.
+      #
+      # TODO: get all tests compiling with just read-core, then remove this and s/check/test/ above.
```
philipc, line 118: This isn't a TODO because we'll never do that. It's not worth the effort and noise.
Instead, always exclude read-core, and keep the existing `cargo check --no-default-features --features read-core`.

.github/workflows/rust.yml
```diff
@@ -108,26 +104,28 @@ jobs:
...
-      - run: cargo test --no-default-features --features read,endian-reader
-      - run: cargo test --no-default-features --features read,endian-reader,std
-      - run: cargo test --no-default-features --features write
+      - name: Install rust stable
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+      - uses: taiki-e/install-action@v2
+        with:
+        tool: cargo-hack
```
philipc, line 113: Inconsistent indentation.

.github/workflows/rust.yml
```diff
@@ -141,23 +139,25 @@ jobs:
...
-      - run: cargo install cross
-      - run: rustup target add ${{matrix.target}}
+      - name: Install rust stable
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: ${{matrix.target}}
+      - uses: taiki-e/install-action@v2
+          with:
+          tool: cross
```
philipc, line 149: Inconsistent indentation.

.github/workflows/rust.yml
```diff
@@ -52,12 +50,10 @@ jobs:
...
-      - name: Install rustup
-        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile=minimal
-      - name: Install nightly rust
-        run: |
-          rustup install nightly
-          rustup default nightly
+      - name: Install rust nightly
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
```
philipc, line 56:
```suggestion
      - uses: dtolnay/rust-toolchain@nightly
```

Same for other places that use a fixed toolchain.

philipc (changes requested):

.github/workflows/rust.yml
tamird: Done.

.github/workflows/rust.yml
tamird: Done.

.github/workflows/rust.yml
tamird: Done.

.github/workflows/rust.yml
tamird:
```
cargo check --no-default-features --features read-core
```
is already covered by the `cargo hack check` above on line 115.

I've added some more invocations to make sure all supported combinations are compiled and tested.

.github/workflows/rust.yml
```diff
@@ -10,20 +10,18 @@ jobs:
   build:
     strategy:
       matrix:
-        os: ["ubuntu-latest"]
-        rust_channel: ["stable", "beta", "nightly", "1.42.0"]
+        os: ['ubuntu-latest']
+        rust_channel: ['stable', 'beta', 'nightly', '1.42.0']
```
philipc, line 14: Why do the different quotes matter?

tamird: They don't - this was just for consistency and it is in its own commit to clarify the intent.

.github/workflows/rust.yml
philipc: I think you're making it too complicated. As I see it, all we need are these two lines:

```
cargo check --no-default-features --features read-core
cargo hack test  --feature-powerset --exclude-features rustc-dep-of-std,read-core
```

tamird: I think you're right; that would produce the same coverage. I was thinking that being that cargo check is faster, this arrangement would produce faster iteration for folks working on this crate. 

Would you prefer your setup?

tamird: Went with your  formulation.

.github/workflows/rust.yml
philipc: This is really only a cosmetic change. Consistency isn't a great reason here. There are at least three different contexts for quotes in this file: YAML strings, github actions expressions, and shell commands, and in all cases there are reasons why you might need different quotes.

The downside of this sort of change is that it is noise in the git history, and takes time to review. I'll keep it now that it has been done though.

.github/workflows/rust.yml
philipc: This still needs to be changed to what I asked for.

tamird: The difference between what's here and your suggestion is that this runs the tests with read-core, which your suggested shape did not. Do you agree?

tamird: I'm sure you're right, but this is encoding knowledge at a distance. The Cargo.toml can change without this configuration changing (and ideally this config is resilient to such changes, erring on the side of failing loudly rather than passing silently).

tamird: Turns out it wasn't that hard to get tests compiling everywhere, so I just did that and simplified this all the way down.

tamird: Hold up, heavy handed on the regex.

tamird: OK, this is ready for another look. Apologies for the churn.

philipc: > The Cargo.toml can change without this configuration changing

No, `read-core` by definition will always be a subset of `read`.

> Turns out it wasn't that hard to get tests compiling everywhere

I'm not interested in this change. It's another hurdle that developers now have to satisfy without providing any value. Testing `read-core` by itself doesn't accomplish anything useful. Please revert this commit back to just the cargo hack changes that I asked for.

> #![deny(warnings)]

Never ever do this. It means we fail to build with new compiler versions if they add new warnings.

philipc: Also, please don't resolve conversations. It makes it harder for me to keep track of the state of the review. I'll resolve them when I'm satisfied with the changes.

tamird: Done.

src/write/unit.rs
```diff
@@ -2,9 +2,11 @@ use alloc::vec::Vec;
 use std::ops::{Deref, DerefMut};
 use std::{slice, usize};
 
+#[cfg(feature = "read")]
+use crate::common::DwoId;
```
philipc, line 6: I think this can be moved into `mod convert`.

philipc (approved):
Thanks!

PR #651, opened 2023-05-03
Bump MSRV to 1.51.0

tamird:
See individual commit messages.

cc @philipc since MSRV was effectively bumped in 059c553fe73abba2b52b56983c32b7f33484629d.


philipc:
`object` is a dev dependency only, so it doesn't affect MSRV. It does make it hard to test the MSRV though.

tamird:
Sent https://github.com/gimli-rs/gimli/pull/652 as an alternative solution.

philipc:
I've merged #653 instead. While we don't really need to support something as old as 1.42.0, I don't want to change it until we have other breaking changes.

PR #652, opened 2023-05-04
Don't run tests or examples at MSRV

tamird:
See individual commit messages.

Alternative to #651.

r? philipc


philipc:
Thanks for the PR. I've gone with a simpler fix in #653.

PR #653, opened 2023-05-05
Fix CI for MSRV

philipc:


tamird:
Can you help me understand how this sidesteps the object crate having MSRV of 1.51.0?

philipc:
I knew it would be possible to do this because it used to work, and nothing changed in this crate since then. So the only changes were in external crates, which we can avoid by pinning versions.

As to why it used to work, I think that 1.42.0 is old enough that cargo ignores the setting instead of giving an error. I haven't checked though.

So this is all a hack, but it's only temporary, and not worth spending more effort on at this time.

PR #655, opened 2023-05-15
Reader::to_{slice,string} -> s/Cow/&//

tamird:
The implementations of these methods always return references; formalize
this in the API. The implementations of `Reader::to_slice` are all
infallible; formalize this as well.

### DIFF
--- a/examples/dwarfdump.rs
+++ b/examples/dwarfdump.rs
@@ -328,12 +328,12 @@ impl<'a, R: gimli::Reader<Offset = usize>> gimli::Reader for Relocate<'a, R> {
     }
 
     #[inline]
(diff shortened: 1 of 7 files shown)


bjorn3:
A custom reader which reads fragments of a file on the fly would have to return `Cow::Owned`, right?

tamird:
I don't see why; can you help me understand?

What you're describing sounds mutable, but these methods take `&self`.

philipc:
@bjorn3 is correct. Using `Cow` was an intentional design decision to allow a `Reader` implementation like that. It would use interior mutability. As such, we won't be making this change.

PR #656, opened 2023-05-15
Prevent hidden re-export to be displayed in documentation

GuillaumeGomez:
I'm working on https://github.com/rust-lang/rust/pull/109697 and it'll make re-export of doc hidden items visible.

Just one thing I'm wondering: is it intended this way or do you want it to be visible in the documentation? If the latter, I'll just close this PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,7 +62,9 @@ pub mod constants;
 pub use crate::constants::*;
 
 mod endianity;
-pub use crate::endianity::{BigEndian, Endianity, LittleEndian, NativeEndian, RunTimeEndian};
+pub use crate::endianity::{BigEndian, Endianity, LittleEndian, RunTimeEndian};
+#[doc(hidden)]
+pub use crate::endianity::NativeEndian;
 
 pub mod leb128;
 


src/lib.rs
```diff
@@ -62,7 +62,9 @@ pub mod constants;
 pub use crate::constants::*;
 
 mod endianity;
-pub use crate::endianity::{BigEndian, Endianity, LittleEndian, NativeEndian, RunTimeEndian};
+pub use crate::endianity::{BigEndian, Endianity, LittleEndian, RunTimeEndian};
+#[doc(hidden)]
+pub use crate::endianity::NativeEndian;
```
philipc, line 67: Note that there are two items named `NativeEndian`, and we want documentation for one of them. I'll have to investigate if the reason for having two items still exists.

philipc: We do still need both the type alias and the constant. If there will be no way to only have documentation for the type alias (as we have currently), then we should have documentation for both. If we need documentation for both, then we'll have to add a documentation comment to the constant.

Also, we have `#![deny(missing_docs)]` in this crate. Will https://github.com/rust-lang/rust/pull/109697 cause the missing documentation on the constant to be a compile error in existing versions of this crate? That would be undesirable I think.

philipc: A hack to keep the current behaviour is to move the `NativeEndian` items into lib.rs.

GuillaumeGomez: The change introduced by the rustdoc PR is about the re-export of `#[doc(hidden)]` items. Currently the re-exports are hidden but this PR will make them appear in documentation (inlined directly at the re-export location).

GuillaumeGomez: Since `NativeEndian` is not currently documented, I thought it was voluntary so I added `#[doc(hidden)]` on the re-export as well. However, we can currently add `#[doc(inline)]` to have the same behaviour as what will be done after this PR is merged.

philipc: I think you've missed my point. There are two items named `NativeEndian`, and only one of them has `#[doc(hidden)]`:
https://github.com/gimli-rs/gimli/blob/1fe9657a8407250f55201924fb826dac7d0bb4d7/src/endianity.rs#L240-L247

GuillaumeGomez: Oh I see, thanks for the clarification. Well, the issue is visible, I'll let you fix however you see more fit. ;)

philipc: Can you answer this question:

Also, we have #![deny(missing_docs)] in this crate. Will https://github.com/rust-lang/rust/pull/109697 cause the missing documentation on the constant to be a compile error in existing versions of this crate?

philipc: I guess I'll also have to dig into https://github.com/rust-lang/rust/pull/109697 to see why these change was done, but to me, needing to use hacks in this crate to get the previous behaviour doesn't sound good either.

GuillaumeGomez: With the `deny`, yes, it'll make a compilation error.

GuillaumeGomez: > I guess I'll also have to dig into [rust-lang/rust#109697](https://github.com/rust-lang/rust/pull/109697) to see why these change was done, but to me, needing to use hacks in this crate to get the previous behaviour doesn't sound good either.

It's a clarification of behaviours. For now, re-exporting `#[doc(hidden)]` isn't really defined, so this PR is adding the missing pieces so it's finally documented and expected.

philipc: Since the behaviour of glob re-exports won't be changing, I'll switch to use that instead in #657 .

PR #659, opened 2023-05-23
Update dependencies

tamird:
See individual commits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ compiler_builtins = { version = "0.1.2", optional = true }
 [dev-dependencies]
 crossbeam = "0.8"
 getopts = "0.2"
-memmap2 = "0.5.5"
+memmap2 = "0.6.1"
 num_cpus = "1"
 object = { version = "0.30.0", features = ["wasm"] }
 rayon = "1.0"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "./README.md"
 repository = "https://github.com/gimli-rs/gimli"
 
 [dependencies]
-fallible-iterator = { version = "0.2.0", default-features = false, optional = true }
+fallible-iterator = { version = "0.3.0", default-features = false, optional = true }
```
philipc, line 15: Do you need something from this version? This is part of the public API, so I prefer to leave this until we have other breaking changes.

examples/dwarfdump.rs
```diff
@@ -497,7 +497,7 @@ fn main() {
                 process::exit(1);
             }
         };
-        let mmap_ref = (*arena_mmap.alloc(mmap)).borrow();
+        let mmap_ref = &*arena_mmap.alloc(mmap);
```
philipc, line 500: These changes are good, but it seems like this is a false positive for the rust compiler warning?

Cargo.toml
tamird: At one point I needed https://github.com/sfackler/rust-fallible-iterator/pull/34, but it seems I don't anymore. Removed this bump.

examples/dwarfdump.rs
tamird: Yeah, I think you're right. I've changed this a bit to simplify.

philipc: Ah nice, even better.

PR #676, opened 2023-08-14
read: add Evaluation::value_result

### LINKED ISSUE #675: DW_AT_frame_base wrong calculation
First of all - thanks to awesome library.
I ran into a problem while trying to get function argument data from libc.
Raw dwarf looks like this:

```
< 1><0x000010a3>    DW_TAG_subprogram
                      DW_AT_external              yes(1)
                      DW_AT_name                  __clock_nanosleep
                      DW_AT_decl_file             0x00000001 ./time/../sysdeps/unix/sysv/linux/clock_nanosleep.c
                      DW_AT_decl_line             0x0000003c
                      DW_AT_decl_column           0x00000001
                      DW_AT_linkage_name          __GI___clock_nanosleep
                      DW_AT_prototyped            yes(1)
                      DW_AT_type                  <0x0000007c>
                      DW_AT_low_pc                0x000e57a0
                      DW_AT_high_pc               <offset-from-lowpc> 478 <highpc: 0x000e597e>
                      DW_AT_frame_base            len 0x0001: 0x9c: 
                          DW_OP_call_frame_cfa
                      DW_AT_call_all_calls        yes(1)
                      DW_AT_sibling               <0x00001659>
< 2><0x000010c9>      DW_TAG_formal_parameter
                        DW_AT_name                  clock_id
                        DW_AT_decl_file             0x00000001
                        DW_AT_decl_line             0x0000003c
                        DW_AT_decl_column           0x00000025
                        DW_AT_type                  <0x00000101>
                        DW_AT_location              0x000d3a73
      .debug_loclists offset  : 0x000d3a73
      <debug_loclists offset 0x000d3a73 with 8 entries follows>
   [ 0]<start,end            0x000e57a0 0x000e57de>
                            DW_OP_reg5
   [ 1]<start,end            0x000e57de 0x000e57fd>
                            DW_OP_reg5
   [ 2]<start,end            0x000e5832 0x000e583f>
                            DW_OP_reg5
   [ 3]<start,end            0x000e583f 0x000e5841>
                            DW_OP_entry_value 0x00000001 contents 0x55
                            DW_OP_stack_value
   [ 4]<start,end            0x000e5841 0x000e5850>
                            DW_OP_reg5
   [ 5]<start,end            0x000e5850 0x000e5875>
                            DW_OP_fbreg -116
   [ 6]<start,end            0x000e58c7 0x000e58db>
                            DW_OP_reg5
   [ 7]<end-of-list>
``` 

So i try to get value of argument clock_id. Now step by step:
1) my current pc is a E5868 so for get value of clock_id we need to eval this expr:
 DW_OP_fbreg -116
2) We need a frame base go to this expr from above:
                      DW_AT_frame_base            len 0x0001: 0x9c: 
                          DW_OP_call_frame_cfa
3) Next, my program calculate CFA and call [`Evaluation::resume_with_frame_base`]
4) On next evaluation iteration  [`Evaluation::evaluate`] returns EvaluationResult::Complete (it's still ok)
5) [`Evaluation::result`] return a vec of Piece (with singe Piece)
6) (And at this step i got a wrong result) piece.location contains a [`Location::Address`] but a [`Location::Value`] expected.
I'm sure what I'm saying (that result must be a  [`Location::Value`]) because I checked the results (in other words, subtracted 116 from the cfa and got the correct value of the variable).

I'm look at sources (https://github.com/gimli-rs/gimli/blob/master/src/read/op.rs#L1977C14-L1977C14) and have a question, why in this situation value interpret as address? maybe it's better to add a new enumerator for such a case (for example [`Location::Raw`])?

philipc:
DWARF expressions can be used to either compute a value or a location description, depending on context.
    
Currently a value computation will result in a location description that contains one `Piece` whose `location` is set to `Location::Address`, and this address is the computed value.
    
`Evaluation::value_result` makes it easier to obtain this value, supports results that are typed values, and checks there were no operations that only valid for location descriptions.

Closes #675

### DIFF
--- a/src/read/op.rs
+++ b/src/read/op.rs
@@ -1077,10 +1077,10 @@ impl<R: Reader> EvaluationStorage<R> for StoreOnHeap {
 ///
 /// # Examples
 /// ```rust,no_run
-/// use gimli::{EndianSlice, Evaluation, EvaluationResult, Format, LittleEndian, Value};
-/// # let bytecode = EndianSlice::new(&[], LittleEndian);
+/// use gimli::{Evaluation, EvaluationResult, Expression};
+/// # let bytecode = gimli::EndianSlice::new(&[], gimli::LittleEndian);
 /// # let encoding = unimplemented!();
-/// # let get_register_value = |_, _| Value::Generic(42);
+/// # let get_register_value = |_, _| gimli::Value::Generic(42);
 /// # let get_frame_base = || 0xdeadbeef;
 ///
 /// let mut eval = Evaluation::new(bytecode, encoding);
@@ -1126,6 +1126,7 @@ pub struct Evaluation<R: Reader, S: EvaluationStorage<R> = StoreOnHeap> {
     // is stored here while evaluating the subroutine.
     expression_stack: ArrayVec<S::ExpressionStack>,
 
+    value_result: Option<Value>,
     result: ArrayVec<S::Result>,
 }
 
@@ -1175,6 +1176,7 @@ impl<R: Reader, S: EvaluationStorage<R>> Evaluation<R, S> {
             stack: Default::default(),
             expression_stack: Default::default(),
             pc,
+            value_result: None,
             result: Default::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


philipc:
@khuey You've probably used the DWARF expression support the most. Do you think this change would be useful?

khuey:
I'm not really convinced this is worth the complexity. What does this solve besides the user's issue in #675? Is the typed value stuff separable from this?

Another issue that the spec does not always match what happens in practice. DW_AT_vtable_elem_location, for example, is treated as a *value* (at least as of clang 16), not a location, despite what the spec claims. There might be others (especially among the weird Fortran attributes).

philipc:
I'll defer to your judgement. If this hasn't been a problem for you then there's no point changing it.

> What does this solve besides the user's issue in https://github.com/gimli-rs/gimli/issues/675?

It encodes some more knowledge in gimli, without requiring every user to duplicate that knowledge.

> Is the typed value stuff separable from this?

We could probably add the `pub fn value_result(&self) -> Value` without needing to add `ExpressionKind`.

> Another issue that the spec does not always match what happens in practice. 

We would change it to do whatever happens in practice then.

philipc:
> Is the typed value stuff separable from this?

This is trivial, so I've changed the PR to just do that.

PR #680, opened 2023-09-13
Add SVE register definitions to AArch64

geowor01:
Also adds a test to check all AArch64 register codes are assigned.

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -822,11 +822,13 @@ registers!(X86_64, {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


philipc:
I wasn't sure what the beta designation meant for these. I assume these are all in common usage now then?

geowor01:
It is indeed still in beta, but the specification is complete and these things aren't expected to change. LLVM already includes the SVE registers in it's DWARF output, so I think it would make sense for readers to be able to interpret these register numbers.

PR #690, opened 2024-01-14
feat(arch): add MIPS register defs

SK83RJOSH:
Add MIPS register definitions, these are similar to LoongArch and map [1-to-1](https://github.com/gcc-mirror/gcc/blob/48448055fb70862ff3914f8a714ff5c4128e6ced/gcc/config/mips/mips.h#L1570C1-L1571C1) with DWARF registers. Adding these so they don't need to redeclared downstream in the unwinding crate. 🙂


philipc (changes requested):
Sorry for taking so long to review, I overlooked this PR.

src/arch.rs
```diff
@@ -509,6 +509,116 @@ aliases {
...
+    T8 = (24, "$t8"),
+    T9 = (25, "$t9"),
+    K0 = (26, "$k0"),
+    K1 = (27, "$k1"),
+    GP = (28, "$gp"),
+    SP = (29, "$sp"),
+    FP = (30, "$fp"),
+    RA = (31, "$ra"),
+
+    S8 = (29, "$s8")
```
philipc, line 619:
```suggestion
    S8 = (30, "$s8")
```
Based on https://wiki.osdev.org/MIPS_Overview

philipc (approved):

PR #691, opened 2024-01-29
Add PowerPC 64-bit DWARF register mappings

tmcgilchrist:
Using the definitions from https://openpowerfoundation.org/specifications/64bitelfabi/ with a small test to check the register mappings. I followed the pattern from https://github.com/gimli-rs/gimli/pull/574 .

Are there other tests required for these changes? I am testing using my own Rust project that inspects ELF binaries and extracts debug information from them.

The Power32 bindings should be quite similar according to http://refspecs.linux-foundation.org/elf/elfspec_ppc.pdf and https://github.com/torvalds/linux/blob/master/arch/powerpc/perf/perf_regs.c

### DIFF
--- a/src/arch.rs
+++ b/src/arch.rs
@@ -820,6 +820,131 @@ registers!(X86_64, {
     K7 = (125, "k7"),
 });
 
+/// PowerPC 64bit
+///
+/// See [64-bit ELF ABI Specification for OpenPOWER Architecture](https://openpowerfoundation.org/specifications/64bitelfabi/).
+#[derive(Debug, Clone, Copy)]
+pub struct PowerPc64;
+
+registers!(PowerPc64, {
+    R0 = (0, "r0"),
+    R1 = (1, "r1"),
+    R2 = (2, "r2"),
+    R3 = (3, "r3"),
+    R4 = (4, "r4"),
+    R5 = (5, "r5"),
+    R6 = (6, "r6"),
+    R7 = (7, "r7"),
+    R8 = (8, "r8"),
+    R9 = (9, "r9"),
+    R10 = (10, "r10"),
+    R11 = (11, "r11"),
+    R12 = (12, "r12"),
+    R13 = (13, "r13"),
+    R14 = (14, "r14"),
+    R15 = (15, "r15"),
+    R16 = (16, "r16"),
+    R17 = (17, "r17"),
+    R18 = (18, "r18"),
+    R19 = (19, "r19"),
+    R20 = (20, "r20"),
+    R21 = (21, "r21"),
+    R22 = (22, "r22"),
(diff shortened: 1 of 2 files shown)


src/arch.rs
```diff
@@ -820,6 +820,131 @@ registers!(X86_64, {
...
 });
 
+/// PowerPC 64bit
+///
+/// See [64-bit ELF ABI Specification for OpenPOWER Architecture](https://openpowerfoundation.org/specifications/64bitelfabi/).
+#[derive(Debug, Clone, Copy)]
+pub struct PowerPc64;
+
+registers!(PowerPc64, {
+    R0 = (0, "R0"),
```
philipc, line 830: For the string name, is it more common to use lowercase or uppercase? The linked document uses lowercase.

src/arch.rs
```diff
@@ -820,6 +820,131 @@ registers!(X86_64, {
...
+
+    CR0 = (68, "CR0"),
+    CR1 = (69, "CR1"),
+    CR2 = (70, "CR2"),
+    CR3 = (71, "CR3"),
+    CR4 = (72, "CR4"),
+    CR5 = (73, "CR5"),
+    CR6 = (74, "CR6"),
+    CR7 = (75, "CR7"),
+    XR = (76, "XER"),
```
philipc, line 907: Should this be:
```suggestion
    XER = (76, "XER"),
```

philipc (commented):
That's fine for the testing. The only better test we could do is check the dwarfdump output on a powerpc binary, but we don't do that sort of thing currently, so as long as you've tested that manually it's fine.

src/arch.rs
tmcgilchrist: Lowercase names would be more common, I'll change to those.

src/arch.rs
```diff
@@ -820,6 +820,131 @@ registers!(X86_64, {
...
+    VR26 = (103, "vr26"),
+    VR27 = (104, "vr27"),
+    VR28 = (105, "vr28"),
+    VR29 = (106, "vr29"),
+    VR30 = (107, "vr30"),
+    VR31 = (108, "vr31"),
+
+    VSCR = (110, "vscr"),
+    TFHAR = (114, "tfhar"),
+    FTIAR = (115, "tfiar"),
```
philipc, line 944:
```suggestion
    TFIAR = (115, "tfiar"),
```

src/arch.rs
tmcgilchrist: Yes it should be.

src/arch.rs
tmcgilchrist: Fixed that one too, hopefully I've got all the strings lined up correctly.

philipc (approved):
Thanks!

PR #692, opened 2024-01-29
Fix some clippy warnings

Enselic:
This PR makes these commands return no warnings (except "unknown lint" from 1.60 for  `clippy::bool_to_int_with_if` and `clippy::derive_partial_eq_without_eq`).

```sh
cargo +1.75 clippy --lib    # current stable version
cargo +1.65 clippy --lib
cargo +1.60 clippy --lib --no-default-features --features read
```

In the case of `#![allow(clippy::unnecessary_cast)]` the current code is by my estimation clearer and more symmetric than the code clippy would want, so I ended up allowing unnecessary casts.

### DIFF
--- a/src/read/aranges.rs
+++ b/src/read/aranges.rs
@@ -182,7 +182,7 @@ where
             .and_then(|x| x.checked_add(segment_size))
             .ok_or(Error::InvalidAddressRange)?;
         if tuple_length == 0 {
-            return Err(Error::InvalidAddressRange)?;
+            return Err(Error::InvalidAddressRange);
         }
         let padding = if header_length % tuple_length == 0 {
             0
--- a/src/read/endian_reader.rs
+++ b/src/read/endian_reader.rs
@@ -393,8 +393,8 @@ where
 
     #[inline]
     fn offset_from(&self, base: &EndianReader<Endian, T>) -> usize {
-        let base_ptr = base.bytes().as_ptr() as *const u8 as usize;
-        let ptr = self.bytes().as_ptr() as *const u8 as usize;
+        let base_ptr = base.bytes().as_ptr() as usize;
+        let ptr = self.bytes().as_ptr() as usize;
         debug_assert!(base_ptr <= ptr);
         debug_assert!(ptr + self.bytes().len() <= base_ptr + base.bytes().len());
         ptr - base_ptr
--- a/src/read/endian_slice.rs
+++ b/src/read/endian_slice.rs
@@ -68,8 +68,8 @@ where
     /// of the given slice.
     #[inline]
     pub fn offset_from(&self, base: EndianSlice<'input, Endian>) -> usize {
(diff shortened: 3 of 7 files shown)


philipc:
> In the case of #![allow(clippy::unnecessary_cast)] the current code is by my estimation clearer and more symmetric than the code clippy would want, so I ended up allowing unnecessary casts.

Do you mean these changes? They seem okay to me at first glance.

```diff
diff --git a/src/read/endian_reader.rs b/src/read/endian_reader.rs
index 8852b380..ef927e52 100644
--- a/src/read/endian_reader.rs
+++ b/src/read/endian_reader.rs
@@ -393,8 +393,8 @@ where
 
     #[inline]
     fn offset_from(&self, base: &EndianReader<Endian, T>) -> usize {
-        let base_ptr = base.bytes().as_ptr() as *const u8 as usize;
-        let ptr = self.bytes().as_ptr() as *const u8 as usize;
+        let base_ptr = base.bytes().as_ptr() as usize;
+        let ptr = self.bytes().as_ptr() as usize;
         debug_assert!(base_ptr <= ptr);
         debug_assert!(ptr + self.bytes().len() <= base_ptr + base.bytes().len());
         ptr - base_ptr
diff --git a/src/read/endian_slice.rs b/src/read/endian_slice.rs
index 0db28dac..b327a84f 100644
--- a/src/read/endian_slice.rs
+++ b/src/read/endian_slice.rs
@@ -68,8 +68,8 @@ where
     /// of the given slice.
     #[inline]
     pub fn offset_from(&self, base: EndianSlice<'input, Endian>) -> usize {
-        let base_ptr = base.slice.as_ptr() as *const u8 as usize;
-        let ptr = self.slice.as_ptr() as *const u8 as usize;
+        let base_ptr = base.slice.as_ptr() as usize;
+        let ptr = self.slice.as_ptr() as usize;
         debug_assert!(base_ptr <= ptr);
         debug_assert!(ptr + self.slice.len() <= base_ptr + base.slice.len());
         ptr - base_ptr
```

src/read/dwarf.rs
```diff
@@ -1046,7 +1046,7 @@ impl<R: Reader> Unit<R> {
     /// The returned value is relative to this unit's `comp_dir`.
     pub fn dwo_name(&self) -> Result<Option<AttributeValue<R>>> {
         let mut entries = self.entries();
-        if let None = entries.next_entry()? {
+        if entries.next_entry()?.is_none() {
```
philipc, line 1049: I think I prefer:
```rust
        entries.next_entry()?;
        let entry = entries.current().ok_or(Error::MissingUnitDie)?;
```
This is more than a clippy fix, but it matches how we handle this case elsewhere in the file.

Enselic:
> Do you mean these changes? They seem okay to me at first glance.

Ok, let's go with those then. I pushed a commit for that now.

philipc (approved):

PR #695, opened 2024-02-21
Fix clippy warnings

GnomedDev:
RA flagged all these warnings, so I just fixed all of them.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -3938,7 +3938,7 @@ mod tests {
         let start = Label::new();
         let end = Label::new();
 
-        let augmentation = Some("replicant");
+        let augmentation = "replicant";
         let expected_rest = [1, 2, 3];
 
         let kind = debug_frame_le();
@@ -3951,7 +3951,7 @@ mod tests {
             // Version
             .D8(4)
             // Augmentation
-            .append_bytes(augmentation.unwrap().as_bytes())
+            .append_bytes(augmentation.as_bytes())
             // Null terminator
             .D8(0)
             // Extra augmented data that we can't understand.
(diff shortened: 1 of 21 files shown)


src/read/unit.rs
```diff
@@ -4207,11 +4207,7 @@ mod tests {
 
     #[test]
     fn test_attribute_udata_sdata_value() {
-        let tests: &[(
-            AttributeValue<EndianSlice<LittleEndian>>,
-            Option<u64>,
-            Option<i64>,
-        )] = &[
+        let tests: &[(AttributeValue<EndianSlice<LittleEndian>>, _, _)] = &[
```
philipc, line 4210: I think this is better left how it is. I assume this is `clippy::type_complexity`, but hiding complexity doesn't make it better.

src/write/op.rs
```diff
@@ -1100,417 +1100,417 @@ mod tests {
...
-                                read::Operation::WasmGlobal { index: 1000 },
-                            ),
-                            (
-                                &|x| x.op_wasm_stack(1000),
-                                Operation::WasmStack(1000),
-                                read::Operation::WasmStack { index: 1000 },
-                            ),
-                        ];
+                    #[allow(clippy::type_complexity)]
+                    let operations: &[(&dyn Fn(&mut Expression), _, _)] = &[
```
philipc, line 1104: I assume this change was an attempt to fix `clippy::type_complexity`? I prefer not to do it.

src/read/cfi.rs
```diff
@@ -6718,15 +6718,16 @@ mod tests {
         let section = EndianSlice::new(&section, LittleEndian);
 
         let mut offset = None;
-        match parse_fde(
+        let result = parse_fde(
```
philipc, line 6721: Which warning is this?

GnomedDev: clippy::blocks_in_conditions

philipc (approved):

PR #700, opened 2024-03-11
Update UnwindContextStorage comments after #687

mstange:
PR #687 switched `StoreOnHeap` to use a `Box` with a fixed-size array. This means that, by default, `UnwindContext` will not allocate during unwind, so the primary motivation for providing a custom `UnwindContextStorage` has changed.

This PR updates the docs.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -1872,7 +1872,8 @@ on the heap using [`Box`]. This is the default storage type parameter for [`Unwi
 You may want to supply your own storage type for one of the following reasons:
 
   1. In rare cases you may run into failed unwinds due to the fixed stack size
-     used by [`StoreOnHeap`], so you may want to try a Vec-based stack instead which
+     used by [`StoreOnHeap`], so you may want to try a larger `Box`. If denial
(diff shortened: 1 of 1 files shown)


src/read/cfi.rs
```diff
@@ -1867,12 +1867,22 @@ impl<R: Reader> FrameDescriptionEntry<R> {
...
     doc = "
 Normally you would only need to use [`StoreOnHeap`], which places the stack
-on the heap using [`Vec`]. This is the default storage type parameter for [`UnwindContext`].
+on the heap using [`Box`]. This is the default storage type parameter for [`UnwindContext`].
+
+You may want to supply your own storage type for one of the following reasons:
+
+  1. In rare cases you may run into failed unwinds due to the fixed stack size
+     used by [`StoreOnHeap`], so you may want to try a Vec-based stack instead which
+     can grow as needed.
```
philipc, line 1876:
```suggestion
  1. In rare cases you may run into failed unwinds due to the fixed stack size
     used by [`StoreOnHeap`], so you may want to try a larger `Box`. If denial
     of service is not a concern, then you could also to try a Vec-based stack which
     can grow as needed.
```

Issue #701, opened 2024-03-11
UnwindContext allocation cannot be reused across multiple unwinds with changing data lifetimes


mstange:
The docs for [`UnwindContext`](https://docs.rs/gimli/latest/gimli/read/struct.UnwindContext.html) say:

"To avoid re-allocating the context multiple times when evaluating multiple CFI programs, it can be reused."

However, reusing the context can be tricky: If your unwind data is stored in a slice with a limited lifetime, and you use `EndianSlice<'mylifetime>` as the `Reader`, then the unwind context cannot outlive that lifetime.

Example program (does not compile):

```rust
use gimli::{BaseAddresses, EhFrame, EndianSlice, NativeEndian, UnwindContext, UnwindSection};
use std::collections::HashMap;

#[derive(Default)]
struct Unwinder {
    eh_frame_data_for_module: HashMap<String, Vec<u8>>,
    unwind_context: UnwindContext<EndianSlice<'static, NativeEndian>>,
}

impl Unwinder {
    fn add_module(&mut self, module_name: impl Into<String>, eh_frame_data: Vec<u8>) {
        self.eh_frame_data_for_module
            .insert(module_name.into(), eh_frame_data);
    }
    fn remove_module(&mut self, module_name: &str) {
        self.eh_frame_data_for_module.remove(module_name);
    }
    fn unwind_in_module(&mut self, module_name: &str, address: u64) {
        let eh_frame_data = self.eh_frame_data_for_module.get(module_name).unwrap();
        let eh_frame = EhFrame::new(&eh_frame_data, NativeEndian);
        let bases = BaseAddresses::default();
        let _unwind_info = eh_frame
            .unwind_info_for_address(
                &bases,
                &mut self.unwind_context, // ERROR: can't be used due to eh_frame_data's lifetime
                address,
                EhFrame::cie_from_offset,
            )
            .unwrap();
    }
}

fn main() {
    let mut unwinder = Unwinder::default();
    unwinder.add_module("lib1.so", vec![0, 1, 2, 3]);
    unwinder.unwind_in_module("lib1.so", 0x1234);
    unwinder.add_module("lib2.so", vec![0, 1, 2, 3]);
    unwinder.unwind_in_module("lib2.so", 0x5678);
}
```

The two remedies I know of are:

 1. If you do not own the data, then you have to create a new `UnwindContext` every time. Ideally you have some scope inside which you can do as many batched unwinds as possible, so that you can at least reuse the context to some degree.
[...]

PR #702, opened 2024-03-12
Proof of concept: Allow recycling UnwindContext allocation

mstange:
I was trying to give the idea from #701 a try. This implementation works, but it uses GATs so it raises the MSRV to 1.65. It's also somewhat messy in a few places - I'm particularly unhappy with the `Option<>` in `ArrayVec` which I had to add because you can't destructure a type which implements Drop.

If addressing #701 requires Rust 1.65 anyway, and thus should be considered a breaking change, then there might be simpler options on the table. If we can eliminate or simplify the `ArrayLike` trait, it might open up simpler options.


philipc:
Raising MSRV is fine. I'm all for anything that can simplify this. Reducing the amount of unsafe here would be great.

philipc:
Closed by #703

PR #703, opened 2024-03-12
read/cfi: replace Expression with offset/length

### LINKED ISSUE #701: UnwindContext allocation cannot be reused across multiple unwinds with changing data lifetimes

philipc:
This avoids the need to store `Reader` in the `UnwindContext`, which simplifies lifetime management.

Closes #701

### DIFF
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -959,10 +959,7 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                         factored_offset
                     )?;
                 }
-                DefCfaExpression {
-                    offset: _,
-                    length: _,
-                } => {
+                DefCfaExpression { expression: _ } => {
                     writeln!(w, "                DW_CFA_def_cfa_expression (...)")?;
                 }
                 Undefined { register } => {
@@ -1036,8 +1033,7 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                 }
                 Expression {
                     register,
-                    offset: _,
-                    length: _,
+                    expression: _,
                 } => {
                     writeln!(
                         w,
@@ -1047,8 +1043,7 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                 }
                 ValExpression {
                     register,
-                    offset: _,
-                    length: _,
+                    expression: _,
                 } => {
                     writeln!(
                         w,
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -959,7 +959,10 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                         factored_offset
                     )?;
                 }
-                DefCfaExpression { expression: _ } => {
+                DefCfaExpression {
+                    offset: _,
+                    length: _,
+                } => {
                     writeln!(w, "                DW_CFA_def_cfa_expression (...)")?;
                 }
                 Undefined { register } => {
@@ -1033,7 +1036,8 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                 }
                 Expression {
                     register,
-                    expression: _,
+                    offset: _,
+                    length: _,
                 } => {
                     writeln!(
                         w,
@@ -1043,7 +1047,8 @@ fn dump_cfi_instructions<R: Reader, W: Write>(
                 }
                 ValExpression {
                     register,
-                    expression: _,
+                    offset: _,
+                    length: _,
                 } => {
                     writeln!(
                         w,
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -459,14 +459,14 @@ impl<'a, R: Reader + 'a> EhHdrTable<'a, R> {
     ///
     /// You must provide a function to get the associated CIE. See
     /// `PartialFrameDescriptionEntry::parse` for more information.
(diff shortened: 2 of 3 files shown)


philipc:
Maybe adding a type for the offset/length pair will make this easier to convert to.

mstange:
I like this solution a lot more than what I proposed!

philipc:
I've added the `UnwindExpression` type. Can you check if this meets your needs? I tried it in `unwinding` and it was a simple change, but `framehop` looks more involved.

mstange:
> Can you check if this meets your needs?

It meets my needs perfectly, thanks so much!

I have the framehop changes on a branch here: https://github.com/mstange/framehop/commits/gimli-issue-701/

> I tried it in `unwinding` and it was a simple change, but `framehop` looks more involved.

I agree, it was quite a bit more involved than I expected.

mstange:
I'd be interested in a release with this fix.

PR #704, opened 2024-03-28
Fix encoding of .eh_frame for small LSDA encoding

bjorn3:


bjorn3:
Without this an LSDA encoded as `DW_EH_PE_pcrel | DW_EH_PE_sdata4` would cause the `.eh_frame` section to be corrupt. I found this issue while implementing unwinding support for cg_clif.

philipc (changes requested):
Thanks!

Can you add a test for this in `test_frame_table` (or change the test there to use `DW_EH_PE_sdata4`).

bjorn3:
Done

philipc (approved):

PR #712, opened 2024-04-25
Fix Op::ImplicitPointer::size() to match Op::ImplicitPointer::write()

crzysdrs:
The `size` function  did not mirror the corresponding size adjustment based on dwarf version 2 found in `write` causing an assertion to fail due to a mismatch between the expected location and the actual write location.

```
thread '<unnamed>' panicked at /home/crzysdrs/proj/gimli/src/write/unit.rs:620:9:
assertion `left == right` failed
  left: DebugInfoOffset(1324)
 right: DebugInfoOffset(1328)
 ```

This looks like an oversight on my part when contributing #618.

### DIFF
--- a/src/write/op.rs
+++ b/src/write/op.rs
@@ -1077,7 +1077,7 @@ mod tests {
     #[test]
     #[allow(clippy::type_complexity)]
     fn test_operation() {
-        for version in [3, 4, 5] {
+        for version in [2, 3, 4, 5] {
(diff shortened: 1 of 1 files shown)


philipc (changes requested):
Can you change `test_operation` to include version 2? I think it would have found this bug.

philipc (approved):
Thanks!

PR #713, opened 2024-05-02
Fix 'hidden lifetime parameters in types are deprecated'

Enselic:
I am messing around with Rust's standard library and gimli. In my particular setup, Rust's build system tries to build gimli in a way that can be simulated like this:

```
cargo rustc --  -Wrust_2018_idioms -Wunused_lifetimes -Dwarnings
```

When build that way, the master branch fails with a bunch of errors like this:

```
error: hidden lifetime parameters in types are deprecated
```

This PR fixes that.

### DIFF
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -28,7 +28,7 @@ pub enum Error {
 
 impl fmt::Display for Error {
     #[inline]
-    fn fmt(&self, f: &mut fmt::Formatter) -> ::std::result::Result<(), fmt::Error> {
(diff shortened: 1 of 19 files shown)


philipc (changes requested):
I'm still not convinced this is useful for parameters (which is the bulk of this PR), but I guess that ship has sailed.

Since you've gone to the effort of changing this, I assume  that you want this to continue to work, so https://github.com/gimli-rs/gimli/blob/c49f64df0920f89c20eda47f76ff3d3655ad0326/src/lib.rs#L27 should be changed to help check that.

Also it appears that other targets such as tests need fixing still.

Enselic:
Thanks for review. I enabled it as a warning and updated tests too.

philipc (approved):

PR #716, opened 2024-05-24
Introduce new enum for `.dwo` section IDs

jameysharp:
Keeping this type distinct from the `SectionId` enum ensures that the compiler will report an error if someone adds incomplete support for new separate-debug sections.

As discussed in: https://github.com/gimli-rs/gimli/pull/714#discussion_r1611164319

The existing `SectionId::dwo_name` method did not match the list of sections actually supported elsewhere in gimli. I've added the one section ID that was missing (`.debug_macinfo.dwo`), but I have not removed the section IDs which are only present there: `.debug_str.dwo`, `.debug_cu_index`, and `.debug_tu_index`.

### DIFF
--- a/src/read/index.rs
+++ b/src/read/index.rs
@@ -129,7 +129,7 @@ pub struct UnitIndex<R: Reader> {
     hash_ids: R,
     hash_rows: R,
     // Only `section_count` values are valid.
-    sections: [DwoSectionId; SECTION_COUNT_MAX as usize],
+    sections: [IndexSectionId; SECTION_COUNT_MAX as usize],
     offsets: R,
     sizes: R,
 }
@@ -144,7 +144,7 @@ impl<R: Reader> UnitIndex<R> {
                 slot_count: 0,
                 hash_ids: input.clone(),
                 hash_rows: input.clone(),
-                sections: [DwoSectionId::DebugAbbrev; SECTION_COUNT_MAX as usize],
+                sections: [IndexSectionId::DebugAbbrev; SECTION_COUNT_MAX as usize],
                 offsets: input.clone(),
                 sizes: input.clone(),
             });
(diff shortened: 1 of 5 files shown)


philipc (changes requested):
Sorry I made a mistake with the name, and your suggestion of `IndexSectionId` is better (since the DWO and DWP contain other sections that can't appear in the index).

jameysharp:
No problem; done. Almost everything I know about DWARF today is stuff I learned this week, so you could easily convince me either way. I'm always happy to take advice.

src/read/index.rs
```diff
@@ -330,13 +330,57 @@ impl<'index, R: Reader> Iterator for UnitIndexSectionIterator<'index, R> {
...
     /// The section kind.
-    pub section: SectionId,
+    pub section: IndexSectionId,
     /// The base offset of the unit's contribution to the section.
     pub offset: u32,
     /// The size of the unit's contribution to the section.
     pub size: u32,
 }
 
+/// Section kinds which are permitted in a `.dwo` file.
```
philipc, line 340:
```suggestion
/// Section kinds which are permitted in a `.dwp` index.
```

philipc (approved):

philipc:
Do you have plans for more gimli work? If not I'll do a release soon (along with object and addr2line).

jameysharp:
I don't have anything more planned right now. I kind of wanted to improve the string formatting of the `Error` type, both to include the data associated with the various `Unknown` and `Unsupported` variants, and also ideally to attach a `ReaderOffsetId` to every error, not just `UnexpectedEof`. But I don't think I actually have the energy for either of those projects at the moment.

I intend to dig into the support for DWARF that's in Wasmtime and Cranelift, so it's possible I'll find more things I'd like from gimli along the way. But for now, I think a new release would be great. Thanks for your help getting these changes merged so promptly!

PR #728, opened 2024-06-27
Add support for DW_LNCT_LLVM_source

### LINKED ISSUE #431: Support DW_LNCT_LLVM_source

Mrmaxmeier:
Hi,

this is a follow-up on #432 (fixes #431).
The discussion in #432 is still relevant here: The accepted proposal explicitly deals with missing source attributes now by specifying that all provided embedded source attributes should be newline-terminated. 

I don't think there's a clean way to implement this though. (=> I've added `fn source_attr` and `fn source` in this PR)

Let me know if you'd be okay with merging support for this feature without waiting for DWARF v6 :slightly_smiling_face:
(I'd like to use this to simplify tests in `rustc` and wire up support in `symbolic`.)

### DIFF
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1688,9 +1688,13 @@ where
         match self.source {
             Some(ref source) => {
                 let res = unit.attr_string(source.clone())?;
-                // If the string is empty (missing mandatory line ending), we'll
-                // return Ok(None)
-                Ok((!res.is_empty()).then_some(res))
+                // Empty source strings (missing mandatory line ending) indicate
+                // that no source was provided for this file
+                if res.is_empty() {
+                    Ok(None)
+                } else {
+                    Ok(Some(res))
+                }
             }
             None => Ok(None),
         }
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1224,16 +1224,6 @@ where
             .any(|x| x.content_type == constants::DW_LNCT_MD5)
     }
 
-    /// Return true if the file debug information entry contains the source code.
-    pub fn file_has_source(&self) -> bool {
-        self.file_name_entry_format.iter().any(|x| {
-            matches!(
-                x.content_type,
-                constants::DW_LNCT_source | constants::DW_LNCT_LLVM_source
-            )
-        })
-    }
-
     /// Get the list of source files that appear in this header's line program.
     pub fn file_names(&self) -> &[FileEntry<R, Offset>] {
         &self.file_names[..]
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1684,7 +1684,7 @@ where
 
     /// The source code of this file. (UTF-8 source text string with "\n" line
     /// endings)
-    pub fn source(&self, unit: &crate::UnitRef<R>) -> crate::Result<Option<R>> {
+    pub fn source(&self, unit: &crate::UnitRef<'_, R>) -> crate::Result<Option<R>> {
         match self.source {
             Some(ref source) => {
                 let res = unit.attr_string(source.clone())?;
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1671,9 +1671,29 @@ where
         &self.md5
     }
 
-    /// FIXME
-    pub fn source(&self) -> Option<AttributeValue<R, Offset>> {
+    /// The source code of this file. (UTF-8 source text string with "\n" line
+    /// endings). Note that this may return an empty attribute that indicates
+    /// that no source code is available. The `FileEntry::source` function
+    /// handles this automatically.
+    ///
+    /// This is an accepted proposal for DWARFv6 (Issue 180201.1: DWARF and
+    /// source text embedding)
+    pub fn source_attr(&self) -> Option<AttributeValue<R, Offset>> {
         self.source.clone()
+    }
+
+    /// The source code of this file. (UTF-8 source text string with "\n" line
+    /// endings)
+    pub fn source(&self, unit: &crate::UnitRef<R>) -> crate::Result<Option<R>> {
+        match self.source {
+            Some(ref source) => {
+                let res = unit.attr_string(source.clone())?;
+                // If the string is empty (missing mandatory line ending), we'll
+                // return Ok(None)
+                Ok((!res.is_empty()).then_some(res))
+            }
+            None => Ok(None),
+        }
     }
 }
 
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1380,6 +1380,7 @@ where
                 timestamp: 0,
                 size: 0,
                 md5: [0; 16],
+                source: None,
             });
 
             file_name_entry_format = Vec::new();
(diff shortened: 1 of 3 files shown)


philipc:
> The accepted proposal explicitly deals with missing source attributes now by specifying that all provided embedded source attributes should be newline-terminated.

The accepted proposal is https://dwarfstd.org/issues/180201.1.html, but that isn't the definitive version of what ends up in the standard. A further proposal modified this: https://dwarfstd.org/issues/211108.2.html. It would be better to refer to the draft standards (https://snapshots.sourceware.org/dwarfstd/dwarf-spec/) to see the current status, but I'm reluctant to implement that here before it is released (or close to it).

So, don't implement `DW_LNCT_source` for now. Adding `DW_LNCT_LLVM_source` is fine though.

src/write/line.rs
```diff
@@ -1052,6 +1093,12 @@ mod convert {
                         timestamp: from_file.timestamp(),
                         size: from_file.size(),
                         md5: *from_file.md5(),
+                        source: match from_file.source_attr() {
+                            Some(source) => {
+                                Some(LineString::from(source, dwarf, line_strings, strings)?)
+                            }
+                            None => None,
+                        },
```
philipc, line 1101: This also needs to set `program.file_has_source` above where `program.file_has_md5` is set.

src/read/line.rs
```diff
@@ -1667,6 +1670,35 @@ where
...
+    ///
+    /// This is an accepted proposal for DWARFv6 (Issue 180201.1: DWARF and
+    /// source text embedding)
+    pub fn source_attr(&self) -> Option<AttributeValue<R, Offset>> {
+        self.source.clone()
+    }
+
+    /// The source code of this file. (UTF-8 source text string with "\n" line
+    /// endings)
+    pub fn source(&self, unit: &crate::UnitRef<'_, R>) -> crate::Result<Option<R>> {
```
philipc, line 1687: I can see why you've done this, but I prefer to avoid dependencies on `UnitRef` or `Unit` here. Currently nothing outside of the `read::dwarf` module requires them. I'd especially like to avoid this since the DWARF 6 support won't need it. I'd be fine with simply documenting that if the DWARF version is 5 then empty may mean not present, and let the user deal with it.

src/read/line.rs
```diff
@@ -1667,6 +1670,35 @@ where
...
     }
+
+    /// The source code of this file. (UTF-8 source text string with "\n" line
+    /// endings). Note that this may return an empty attribute that indicates
+    /// that no source code is available. The `FileEntry::source` function
+    /// handles this automatically.
+    ///
+    /// This is an accepted proposal for DWARFv6 (Issue 180201.1: DWARF and
+    /// source text embedding)
+    pub fn source_attr(&self) -> Option<AttributeValue<R, Offset>> {
```
philipc, line 1681:
```suggestion
    pub fn source(&self) -> Option<AttributeValue<R, Offset>> {
```

philipc (changes requested):

Mrmaxmeier:
> So, don't implement `DW_LNCT_source` for now. Adding `DW_LNCT_LLVM_source` is fine though.

I'm still calling the attribute `source` at the moment. Should it explicitly be `llvm_source` instead?

src/write/line.rs
```diff
@@ -632,6 +645,20 @@ impl LineProgram {
...
                 }
+                if self.file_has_source {
+                    // Note: An empty DW_LNCT_source is interpreted as missing
+                    // source code. Included source code should always be
+                    // terminated by a "\n" line ending.
+                    let empty_str = LineString::String(Vec::new());
+                    let source = info.source.as_ref().unwrap_or(&empty_str);
+                    source.write(
+                        w,
+                        constants::DW_FORM_string,
```
philipc, line 656: Ideally users should be able to use `DW_FORM_line_strp` too, but that has the problem of needing to have added an empty string already, and doing that won't be necessary for DWARF 6. So for now, let's just document that only `LineString::String` is supported for `source`.

src/read/line.rs
```diff
@@ -1224,6 +1224,13 @@ where
             .any(|x| x.content_type == constants::DW_LNCT_MD5)
     }
 
+    /// Return true if the file name entry format contains an source field.
```
philipc, line 1227:
```suggestion
    /// Return true if the file name entry format contains a source field.
```

src/write/line.rs
```diff
@@ -67,6 +67,12 @@ pub struct LineProgram {
     /// For version 5, this controls whether to emit `DW_LNCT_MD5`.
     pub file_has_md5: bool,
 
+    /// True if the file entries have embedded source code.
+    ///
+    /// For version <= 4, this is ignored.
+    /// For version 5, this controls whether to emit `DW_LNCT_source`.
```
philipc, line 73:
```suggestion
    /// For version 5, this controls whether to emit `DW_LNCT_LLVM_source`.
```

src/write/line.rs
```diff
@@ -632,6 +645,20 @@ impl LineProgram {
                 if self.file_has_md5 {
                     w.write(&info.md5)?;
                 }
+                if self.file_has_source {
+                    // Note: An empty DW_LNCT_source is interpreted as missing
```
philipc, line 649:
```suggestion
                    // Note: An empty DW_LNCT_LLVM_source is interpreted as missing
```

src/write/line.rs
```diff
@@ -1190,6 +1238,13 @@ mod tests {
                             program.file_has_md5 = true;
                         }
 
+                        // Note: Embedded source code is an accepted extension
+                        // that will become part of DWARF v6. We're using it for
+                        // v5 here anyways.
```
philipc, line 1243:
```suggestion
                        // that will become part of DWARF v6. We're using the LLVM extension
                        // here for v5.
```

philipc (changes requested):
> I'm still calling the attribute source at the moment. Should it explicitly be llvm_source instead?

No, leave it as `source`, it will be used for both in future.

src/write/line.rs
Mrmaxmeier: Added a note to the `write::FileInfo` field. :+1:

philipc (approved):
Thanks!

Mrmaxmeier:
Hi, can you cut a new release for crates.io? :slightly_smiling_face:
(This PR includes a breaking change: `FileInfo` used to be `Copy`)
Thanks!

philipc:
Released 0.31.0

PR #731, opened 2024-07-04
read/line: check for overflow when advancing the address

philipc:
Addresses in line sequences should only increase.
The `addr2line` crate relies on this for its binary search.


src/read/line.rs
```diff
@@ -800,21 +800,21 @@ impl LineRow {
         &mut self,
         instruction: LineInstruction<R>,
         program: &mut Program,
-    ) -> bool
```
philipc, line 803: Note: this is a breaking change.

PR #733, opened 2024-07-05
read: use address size when checking address overflow

philipc:


philipc:
Address size validation was needed to fix the [fuzz error found in CI](https://github.com/gimli-rs/gimli/actions/runs/9803484125/job/27069738694).

PR #735, opened 2024-07-08
read: add support for custom address types

### LINKED ISSUE #409: Add `Reader::Address` associated type
This should support the following uses:
- better support for relocated addresses (eg when using function sections, which have every function symbol at address 0)
- reuse of the same address type for conversion when both reading and writing
- optional support for segmented addresses (#357)
- using the native address size
- better type checking of our address handling

I wanted this when doing the relocated address support, but didn't do it because of how ugly the `Reader::Offset` support had been. However, I don't think that's so bad now after the default offset was changed in #392.

Any thoughts on this @fitzgen ?

philipc:
Add the `Reader::Address` associated type, which implements the `ReaderAddress` trait. This trait is implemented for u64 to provide the existing behaviour of supporting multiple address sizes, as well as for `Address32` and `Address64` which support a single address size.

The dwarfdump example now provides its own address type, which allows it to show symbols associated with relocated addresses in relocatable object files. For example:
```
    DW_AT_low_pc                main+0x0
    DW_AT_ranges                <rnglist at .debug_rnglists+0x0000000c>
        [ 0] <start-length main+0x0, 0x17> [main+0x0, main+0x17]

    DW_LNE_set_address to main+0x0
```

Closes #409


philipc:
This is a breaking change, and adds significant complexity. While I think the dwarfdump support is neat, I'm not completely convinced it is worth it, so I'm going to leave this as a draft for while.

To do:
- investigate how much this affects downstream crates
- investigate how this integrates with the write support (are the `convert_address` callbacks still needed?)

PR #745, opened 2024-08-22
write: add `LineConvert`

philipc:
This allow reuse of the implementation of `LineProgram::from` as part of a more complex transformation.

Operation of `LineProgram::from` is mostly unchanged. The one difference is that it now uses the string form chosen by `LineString::new` instead of copying the form of the input.


philipc:
I've experimented with converting the transformations in [Wasmtime](https://github.com/philipc/wasmtime/commit/e9210fe09dc2bf70ede7e0aa9fc3ba3d22e75613) and [walrus](https://github.com/philipc/walrus/commit/05f9bac1aa670be016edabcb4b15cf5267ab2e76) to use this.

fitzgen:
Nice! Will take a look later today!

src/write/line.rs
```diff
@@ -1003,177 +1018,378 @@ mod convert {
...
+        pub length: u64,
+        /// The rows in the sequence.
+        ///
+        /// The `LineRow::address` fields are set to an offset from `Self::start`.
+        pub rows: Vec<LineRow>,
+    }
+
+    /// The state for the conversion of a line number program.
+    #[derive(Debug)]
+    pub struct LineConvert<'a, R: Reader> {
```
fitzgen, line 1105: Is this intended for use outside of `LineProgram::from`? Because from a user-perspective, these docs don't tell me much about why I'd use this or how.

src/write/line.rs
```diff
@@ -1003,177 +1018,378 @@ mod convert {
...
+            self.program.in_sequence()
+        }
+
+        /// Set the next row and call [`LineProgram::generate_row`] for the converted program.
+        pub fn generate_row(&mut self, row: LineRow) {
+            *self.program.row() = row;
+            self.program.generate_row();
+        }
+
+        /// Return the program and a mapping from source file index to `FileId`.
```
fitzgen, line 1390: Maybe mention the subtelties with `files[0]` for dwarf <=4 here, and how a default entry is added.

fitzgen (approved):
Overall, LGTM. Thanks a ton for digging into this stuff!

I have to admit that it has been a while since I was deep in DWARF, so I'm not 100% up-to-date with all the details. My biggest piece of feedback is simply that of a downstream user: I think these APIs could benefit from a bit more documentation explaining what they are, why you'd use them, and how.

I also quickly scanned over the changes in Wasmtime and Walrus, and things also look good there, definitely an improvement!

philipc:
Thanks for the review. I've expanded the doc and added some examples.

I'll let this stew for a bit. I probably need to add some more edge case testing, and I need to figure out if we should be supporting `DW_LNE_set_address` in the middle of a sequence.

PR #749, opened 2024-09-02
arch: added missing MIPS HI/LO register defs

Tazdevil971:
As per title, this pull request adds missing HI/LO register definitions, those registers are used in older MIPS CPU (pre R6).

Register DWARF numbers come directly from [LLVM](https://github.com/llvm/llvm-project/blob/ecc9aece72989461167ba39541dce51032c0d76d/llvm/lib/Target/Mips/MipsRegisterInfo.td#L183)


philipc (commented):
The rest of the MIPS register definitions have a `$` prefix. I assume these should too.

Tazdevil971:
Thanks for the speedy reply! I updated the patch to now include the `$` prefix. My original thought was that since the HI/LO registers are special, and cannot be referenced or named directly in assembly (they can only be accessed though special `mflo`/`mfhi`/`mtlo`/`mthi` instructions), it would have been better to separate them from the rest.

But looking at other registers definitions no one does that, also I agree that it looks better like this.

philipc (approved):
Thanks!

PR #750, opened 2024-09-03
read: improve handling of 0 for tombstones

philipc:
Where possible, ignore these instead of returning them or returning an error. This often isn't possible because 0 is a valid address, but we can handle it for `DW_LNE_set_address` in the middle of a line sequence, and for address pairs.

This fixes one of the problems reported in https://github.com/gimli-rs/addr2line/issues/326.

### DIFF
--- a/src/read/aranges.rs
+++ b/src/read/aranges.rs
@@ -290,6 +290,8 @@ impl<R: Reader> ArangeEntryIter<R> {
     #[doc(hidden)]
     pub fn convert_raw(&self, mut entry: ArangeEntry) -> Result<Option<ArangeEntry>> {
         // Skip tombstone entries.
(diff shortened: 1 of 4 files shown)


marxin:
I've just noticed this PR when I was working on https://github.com/davidlattimore/wild/issues/37.
@philipc Can you please provide a link to the DWARF standard where are the tombstone values defined?

philipc:
It's actually only DWARF 6 that will specify this. The issue is https://dwarfstd.org/issues/200609.1.html, and you can also check the draft DWARF 6 standard. lld tried changing to -1 but it broke too many DWARF consumers, so it changed to 0, with a plan to try -1 again sometime in the future. https://reviews.llvm.org/D84825 contains a lot of relevant discussion.

marxin:
Thank you for the useful pointers!

PR #752, opened 2024-09-27
read: use const generics in `src/read/util.rs`

nbdd0121:
The MSRV (1.60) has been bumped past the minimum required for const generics (1.51).


philipc (approved):
Thanks!

PR #758, opened 2025-04-14
typo in UnknownCallFrameInstruction error

psifertex:
Sorry for the relatively trivial PR but we [ran into](https://github.com/Vector35/binaryninja-api/issues/6605#issuecomment-2802910329) this in a project that uses gimli downstream so I happened to notice it.

Appreciate the great library!


philipc (approved):
Thanks!

PR #759, opened 2025-05-01
Add support for reading items from .debug_macinfo

DanielT:
The section .debug_macinfo was specified in DWARF standards v2, v3 and v4. It was replaced by .debug_macro in DWARF v5. The content of .debug_macinfo is represented through the enum DebugMacInfoItem. It has variants for each specified type:
- macro defined
- macro undefined
- file includd
- end of include
- vendor extension
- end of unit

This change does not implement writing of .debug_macinfo.

### DIFF
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -1142,6 +1142,7 @@ fn dump_entries<R: Reader, W: Write>(
     flags: &Flags,
 ) -> Result<()> {
     let mut spaces_buf = String::new();
+    let mut deferred_macinfo = Vec::new();
 
     let mut entries = unit.entries_raw(None)?;
     while !entries.is_empty() {
@@ -1186,9 +1187,23 @@ fn dump_entries<R: Reader, W: Write>(
                         writeln_error(w, unit.dwarf, err, "Failed to dump attribute value")?
                     }
                 };
+                // dump_attr_value only prints the offset for the macro info attribute.
+                // The content is too long to print inline, so store the offset to print later.
+                if attr.name() == gimli::DW_AT_macro_info {
+                    if let gimli::AttributeValue::DebugMacinfoRef(offset) = attr.value() {
+                        deferred_macinfo.push(offset);
+                    }
+                }
             }
         }
     }
+
+    for offset in deferred_macinfo {
+        writeln!(w)?;
+        writeln!(w, "Macros <.debug_macinfo+0x{:08x}>", offset.0)?;
+        dump_macinfo_list(w, unit, offset)?;
+    }
+
     Ok(())
 }
 
@@ -2263,6 +2278,76 @@ fn dump_addr<R: Reader, W: Write>(w: &mut W, debug_addr: &gimli::DebugAddr<R>) -
             )?
         }
         writeln!(w, "]",)?;
+    }
+    Ok(())
+}
+
+fn dump_macinfo_list<R: Reader, W: Write>(
+    w: &mut W,
+    unit: gimli::UnitRef<'_, R>,
+    offset: gimli::DebugMacinfoOffset,
+) -> Result<()> {
+    let mut indent = 2; // base indent is 2 spaces
+    let mut macinfo_iter = unit.macinfo(offset)?;
+    while let Some(macinfo) = macinfo_iter.next()? {
+        match macinfo {
+            gimli::DebugMacInfoItem::StartFile { .. } => {
+                // print the item first, then indent
+                write!(w, "{:indent$}", "", indent = indent)?;
+                dump_macinfo(w, unit, macinfo)?;
+                indent += 2;
+            }
+            gimli::DebugMacInfoItem::EndFile => {
+                // unindent first, then print the item
+                indent -= 2;
+                write!(w, "{:indent$}", "", indent = indent)?;
+                dump_macinfo(w, unit, macinfo)?;
+            }
+            _ => {
+                // no indentation change
+                write!(w, "{:indent$}", "", indent = indent)?;
+                dump_macinfo(w, unit, macinfo)?;
+            }
+        }
+    }
+    Ok(())
+}
+
+fn dump_macinfo<R: Reader, W: Write>(
+    w: &mut W,
+    unit: gimli::UnitRef<'_, R>,
+    macinfo: gimli::DebugMacInfoItem<R>,
+) -> Result<()> {
+    match macinfo {
+        gimli::DebugMacInfoItem::Define { line, text } => {
+            writeln!(
+                w,
+                "DW_MACINFO_define - lineno: {line}, macro: {}",
+                text.to_string_lossy()?
+            )?;
+        }
+        gimli::DebugMacInfoItem::Undef { line, name } => {
+            writeln!(
+                w,
+                "DW_MACINFO_undef - lineno: {line}, macro: {}",
+                name.to_string_lossy()?
+            )?;
+        }
+        gimli::DebugMacInfoItem::StartFile { line, file } => {
+            write!(w, "DW_MACINFO_start_file - lineno: {line}, file: ")?;
+            dump_file_index(w, file, unit)?;
+            writeln!(w)?;
+        }
+        gimli::DebugMacInfoItem::EndFile => {
+            writeln!(w, "DW_MACINFO_end_file")?;
+        }
+        gimli::DebugMacInfoItem::VendorExt { numeric, string } => {
+            writeln!(
+                w,
+                "DW_MACINFO_vendor_ext - number: {numeric}, string: {}",
+                string.to_string_lossy()?
+            )?;
+        }
     }
     Ok(())
 }
--- a/src/read/macros.rs
+++ b/src/read/macros.rs
@@ -1,0 +1,271 @@
+use core::fmt::Debug;
+
+use crate::common::{DebugMacinfoOffset, SectionId};
+use crate::endianity::Endianity;
+use crate::read::{EndianSlice, Reader, Section};
+use crate::{constants, DwMacInfo, Error, Result};
+
+/// The raw contents of the `.debug_macinfo` section.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct DebugMacInfo<R> {
+    pub(crate) section: R,
+}
+
+impl<'input, Endian> DebugMacInfo<EndianSlice<'input, Endian>>
+where
+    Endian: Endianity,
+{
+    /// Construct a new `DebugMacInfo` instance from the data in the `.debug_macinfo`
+    /// section.
+    ///
+    /// It is the caller's responsibility to read the `.debug_macinfo` section and
+    /// present it as a `&[u8]` slice. That means using some ELF loader on
+    /// Linux, a Mach-O loader on macOS, etc.
+    ///
+    /// ```
+    /// use gimli::{DebugMacInfo, LittleEndian};
+    ///
+    /// # let buf = [1, 0, 95, 95, 83, 84, 68, 67, 95, 95, 32, 49, 0];
+    /// # let read_section_somehow = || &buf;
+    /// let debug_str = DebugMacInfo::new(read_section_somehow(), LittleEndian);
+    /// ```
+    pub fn new(macinfo_section: &'input [u8], endian: Endian) -> Self {
+        Self::from(EndianSlice::new(macinfo_section, endian))
+    }
+}
+
+impl<R: Reader> DebugMacInfo<R> {
+    /// Look up a macro reference the `.debug_macinfo` section by DebugMacinfoOffset.
+    ///
+    /// A macinfo offset points to a list of macro information entries in the `.debug_macinfo` section.
+    /// To handle this, the function returns an iterator.
+    ///
+    /// ```
+    /// use gimli::{DebugMacInfo, DebugMacinfoOffset, LittleEndian};
+    ///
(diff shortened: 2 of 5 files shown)


philipc:
Thanks for working on this. I haven't reviewed your changes, but I wanted to do a quick comparision with the llvm output so I tried running dwarfdump on a file here and I get output like:
```
...
                    DW_AT_high_pc               <offset-from-lowpc>348
                    DW_AT_macro_info            Macro info: Start include file on line 0:  /object/testfiles/dwarf/base.cpp< 1><0x0000002e>    DW_TAG_variable
                      DW_AT_name                  sarr
...
```

For reference, llvm-dwarfdump simply gives the offset and lists the macinfo contents elsewhere:
```
...
              DW_AT_high_pc     (0x000000000000128c)
              DW_AT_macro_info  (0x00000000)

0x0000002e:   DW_TAG_variable
                DW_AT_name      ("sarr")
...

.debug_macinfo contents:
0x00000000:
DW_MACINFO_start_file - lineno: 0 filenum: 1
  DW_MACINFO_define - lineno: 1 macro: SIZE 10000
DW_MACINFO_end_file
DW_MACINFO_define - lineno: 0 macro: __llvm__ 1
DW_MACINFO_define - lineno: 0 macro: __clang__ 1
...
```

Note that in addition to missing a line break, the gimli output only gives one line of the macinfo data, whereas there are many more lines that would be useful to see. So I think the way llvm-dwarfdump does it is better. It's also better to simply list the constants and arguments (such as `DW_MACINFO_start_file`) instead of converting it to a description (such as "Start include file").

DanielT:
I just updated my branch:
Your comment made me realize that I had misunderstood the spec, and was missing most of the macinfo entries.
The DW_AT_macro_info attribute actually points to a list of items which is terminated with an end of unit item.

As a result, i changed `get_macinfo(offset)` to return an iterator over the items found starting at the `offset`.
I also modified dwarfdump to use this iterator to print all of the items. Unlike llvm-dwarfdump, they are printed inline with the DIEs, but the formatting should be generally improved compared to my first version..

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+        offset: DebugMacinfoOffset<R::Offset>,
+    ) -> impl Iterator<Item = DebugMacInfoItem<R>> {
+        let mut input = self.section.clone();
+        if let Ok(_) = input.skip(offset.0) {
+            DebugMacInfoIterator {
+                input_reader: input,
+                finished: false,
+            }
+        } else {
+            // If the offset is invalid, the iterator is immediately finished.
```
philipc, line 67: Return the error instead of an empty iterator.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+            section: macinfo_section,
+        }
+    }
+}
+
+/// an Entry in the `.debug_macinfo` section.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DebugMacInfoItem<R> {
+    /// The end of the unit.
+    EndOfUnit,
```
philipc, line 112:
```suggestion
```
This is never created, and shouldn't be, so can delete this.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+    /// A macro definition.
+    Define {
+        /// The line number where the macro is defined.
+        line_number: u64,
+        /// The text of the macro: a null-terminated string.
```
philipc, line 117: I think we should make it clear that the text is the macro name, arguments, and value.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+        macro_text: R,
+    },
+    /// A macro undefinition.
+    Undef {
+        /// The line number where the macro is undefined.
+        line_number: u64,
+        /// The text of the macro: a null-terminated string.
```
philipc, line 124: Similarly this is just the macro name

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+impl<R: Reader> Iterator for DebugMacInfoIterator<R> {
+    type Item = DebugMacInfoItem<R>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // make sure the iterator never returns Some(x) after returning None
+        if self.finished {
+            return None;
+        }
+        // Read the next entry from the input reader and return it as a DebugMacInfoItem.
+        let macinfo_type = DwMacInfo(self.input_reader.read_u8().ok()?);
```
philipc, line 162: Please make this a fallible iterator instead. Never do `.ok()?`.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+        /// a null-terminated string, whose meaning is vendor specific.
+        string_constant: R,
+    },
+}
+
+/// Iterator over the entries in the `.debug_macinfo` section.
+///
+/// the iterator is not public and does not appear in the API - callers only get an `impl Iterator<Item = DebugMacInfoItem<R>>`
+struct DebugMacInfoIterator<R: Reader> {
+    input_reader: R,
```
philipc, line 149:
```suggestion
    input: R,
```

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+    finished: bool,
```
philipc, line 150:
```suggestion
```
Use `input.is_empty()` and `input.empty()` instead.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+                let string_constant = self.input_reader.read_null_terminated_slice().ok()?;
+                Some(DebugMacInfoItem::VendorExt {
+                    numeric_constant,
+                    string_constant,
+                })
+            }
+            _ => {
+                // unknown macinfo type; return None, since Err is not an option here
+                self.finished = true;
+                return None;
```
philipc, line 210: Change to return an error.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+        let debug_macinfo = DebugMacInfo::from(EndianSlice::new(&section, LittleEndian));
+
+        let offset = position.value().unwrap() as usize;
+
+        let mut iter = debug_macinfo.get_macinfo(DebugMacinfoOffset(offset));
+
+        // Test getting macinfo entries
+        let entry = iter.next().unwrap();
+        assert!(
+            matches!(entry, DebugMacInfoItem::Define { line_number, macro_text } if line_number == 0 && macro_text.to_string_lossy() == "__STDC__ 1")
```
philipc, line 262:
```suggestion
            matches!(entry, DebugMacInfoItem::Define { line_number: 0, macro_text } if macro_text.slice() == b"__STDC__ 1")
```
Similarly for others.

src/read/macinfo.rs
```diff
@@ -0,0 +1,290 @@
...
+        file_number_index: u64,
+    },
+    /// The end of a file.
+    EndFile,
+    /// A vendor-specific extension.
+    VendorExt {
+        /// a numeric constant (LEB128), whose meaning is vendor specific.
+        numeric_constant: u64,
+        /// a null-terminated string, whose meaning is vendor specific.
+        string_constant: R,
```
philipc, line 141:
```suggestion
        /// A numeric constant whose meaning is vendor specific.
        numeric: u64,
        /// A string whose meaning is vendor specific.
        string: R,
```
Field names are needlessly verbose. The DWARF encoding (LEB128 and null-terminated) is irrelevant here. Capitalize the start of comments.

Similarly for other variants.

crates/examples/src/bin/dwarfdump.rs
```diff
@@ -1313,6 +1313,7 @@ fn dump_attr_value<R: Reader, W: Write>(
         }
         gimli::AttributeValue::DebugMacinfoRef(offset) => {
             writeln!(w, "<.debug_macinfo+0x{:08x}>", offset.0)?;
+            dump_macinfo_list(w, unit, offset)?;
```
philipc, line 1316: I still think that macros can be too verbose to be listed inline like this.

philipc (changes requested):

src/read/mod.rs
```diff
@@ -252,6 +252,11 @@ pub use self::rnglists::*;
 mod str;
 pub use self::str::*;
 
+#[cfg(feature = "read")]
+mod macinfo;
+#[cfg(feature = "read")]
+pub use self::macinfo::*;
```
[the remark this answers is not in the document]
DanielT: `macro` is actually a [reserved keyword](https://doc.rust-lang.org/reference/keywords.html#reserved-keywords) in rust.

How about `debug_macro` or `macros` instead?

src/read/macinfo.rs
DanielT: All uses of fallible iterator depend on the "fallible-iterator" feature.
If I make this iterator a FallibleIterator, then reading the content of .debug_macinfo as a whole would have to depend on this feature. I avoided this dependency by invalidating the iterator on error, but if you think making it fallible is better then I'll change it.

src/read/mod.rs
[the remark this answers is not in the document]
DanielT: renamed to `macros`

DanielT:
New update - as far as I can tell I've addressed all of your comments.

The module is renamed to `macros` since `macro` is reserved.
I also switched to a FallibleIterator as you requested. This resulted in the addition of several `#[cfg(feature = "fallible-iterator")]` guards, so .debug_macinfo is not usable without this feature. That doesn't bother me much though, since I've never disabled this anyway.

philipc:
The style used in this crate is to always expose a `next` method on the struct that isn't part of any trait implementation, and then use the `fallible-iterator` feature to control the existence of only the `FallibleIterator` implementation. The `next` method in that trait implementation is simply a one liner to call the other `next` method (they both have the same signature).

DanielT:
Oh, I see. In that case, what is the benefit of the fallible-iterator crate?

philipc:
The `FallibleIterator` trait provides many more methods that may be useful for working with the iterator, such as for chaining adaptors. I personally rarely use them, but occasionally they can make code simpler.

DanielT:
In case you missed it - my last push yesterday changed the iterator implementation, so fallible-iterator is no longer required.

src/read/macros.rs
```diff
@@ -0,0 +1,289 @@
...
+        line_number: u64,
+        /// The name of the macro without the definition.
+        macro_name: R,
+    },
+    /// The start of a file.
+    StartFile {
+        /// Line number of the source file on which the inclusion macro directive occurred.
+        line_number: u64,
+        /// An index into the line number table of the compilation unit.
+        file_number_index: u64,
```
philipc, line 125:
```suggestion
        line: u64,
        /// The text of the macro: The name of the macro followed immediately by any formal
        /// parameters including the surrounding parentheses, followed by the macro definition.
        text: R,
    },
    /// A macro undefinition.
    Undef {
        /// The line number where the macro is undefined.
        line: u64,
        /// The name of the macro without the definition.
        name: R,
    },
    /// The start of a file.
    StartFile {
        /// Line number of the source file on which the inclusion macro directive occurred.
        line: u64,
        /// An index into the line number table of the compilation unit.
        file: u64,
```

src/read/macros.rs
```diff
@@ -0,0 +1,289 @@
...
+                // two operands: a constant (LEB128) and a null terminated string, whose meaning is vendor specific
+                let numeric_constant = self.input_reader.read_uleb128()?;
+                let string_constant = self.input_reader.read_null_terminated_slice()?;
+                Ok(Some(DebugMacInfoItem::VendorExt {
+                    numeric: numeric_constant,
+                    string: string_constant,
+                }))
+            }
+            _ => {
+                // unknown macinfo type; return None, since Err is not an option here
```
philipc, line 194:
```suggestion
```

philipc (changes requested):

DanielT:
Done!

crates/examples/src/bin/dwarfdump.rs
```diff
@@ -2266,3 +2281,85 @@ fn dump_addr<R: Reader, W: Write>(w: &mut W, debug_addr: &gimli::DebugAddr<R>) -
...
+
+fn dump_macinfo<R: Reader, W: Write>(
+    w: &mut W,
+    unit: gimli::UnitRef<'_, R>,
+    macinfo: gimli::DebugMacInfoItem<R>,
+) -> Result<()> {
+    match macinfo {
+        gimli::DebugMacInfoItem::Define {
+            line: line_number,
+            text: macro_text,
```
philipc, line 2324:
```suggestion
            line,
            text,
```

Similar elsewhere.

philipc (changes requested):

DanielT:
>            text: macro_text,
Oof. That's the result of trying to automatically rename through rust-language-server, and I didn't check the diff afterwards.
Sorry for that!

It's fixed now.

src/read/macros.rs
```diff
@@ -0,0 +1,283 @@
...
+                // found the end of the unit, return None to stop the iteration
+                self.input.empty();
+                Ok(None)
+            }
+            constants::DW_MACINFO_define => {
+                let line_number = self.input.read_uleb128()?;
+                let macro_text = self.input.read_null_terminated_slice()?;
+                Ok(Some(DebugMacInfoItem::Define {
+                    line: line_number,
+                    text: macro_text,
```
philipc, line 158: Some more here still.

philipc (approved):
Thanks!

PR #762, opened 2025-05-07
Fix clippy warnings

DanielT:
- lifetimes could be omitted or written as '_ in several cases
- x.len() >= 1 was changed to !x.is_empty()


philipc:
The `is_empty()` change is fine, but I disagree with the lifetimes and prefer to add a clippy attribute to allow it instead.

DanielT:
Ok, no problem.
Is this a lint you never care about, in which case I would disable it in lib.rs for the whole crate.
Alternatively, I could annotate the individual lines with `#[allow(clippy::needless_lifetimes)]`

philipc:
Disable it in lib.rs please.

PR #763, opened 2025-05-11
Add support for reading macros from .debug_macro

DanielT:
.debug_macro replaces .debug_macinfo from Dwarf 5 onward.

The implementation is analogous to the code for .debug_macinfo, execpt that more cases need to be handled.

User defined debug_macro opcodes are not supported by this change, and an error would be triggered if one was encountered.
I might be wrong, but I think no one is using this anyway. In any case, it's not useful to me, so I decided not to implement this part of the spec. If turns out that someone _does_ use this after all, it can be added later.

One additional note regarding dwarfdump: it only dumps the .debug_macro units that are directly referenced from compilation units. This is the usual case.
With the new format, macro unit can reference other macro units using `DW_MACRO_import`, but these imported units are not dumped. Doing this would be a whole project, de-duplicating units, sorting units, etc., and I actually have no elf files that use `DW_MACRO_import`, so I wouldn't be able to test it.

### DIFF
--- a/src/read/macros.rs
+++ b/src/read/macros.rs
@@ -1,9 +1,10 @@
-use core::fmt::Debug;
-
 use crate::common::{DebugMacinfoOffset, SectionId};
 use crate::endianity::Endianity;
-use crate::read::{EndianSlice, Reader, Section};
-use crate::{constants, DwMacInfo, Error, Result};
+use crate::read::{EndianSlice, Reader, ReaderOffset, Section, UnitRef};
+use crate::{
+    constants, DebugLineOffset, DebugMacroOffset, DebugStrOffset, DebugStrOffsetsIndex, DwMacInfo,
+    DwMacro, Error, Format, Result,
+};
 
 /// The raw contents of the `.debug_macinfo` section.
 #[derive(Debug, Default, Clone, Copy)]
@@ -55,13 +56,10 @@ impl<R: Reader> DebugMacInfo<R> {
     /// }
     /// # Ok(()) }
     /// ```
-    pub fn get_macinfo(
-        &self,
-        offset: DebugMacinfoOffset<R::Offset>,
-    ) -> Result<DebugMacInfoIterator<R>> {
+    pub fn get_macinfo(&self, offset: DebugMacinfoOffset<R::Offset>) -> Result<MacInfoIter<R>> {
         let mut input = self.section.clone();
         input.skip(offset.0)?;
-        Ok(DebugMacInfoIterator { input })
+        Ok(MacInfoIter { input })
     }
 }
 
@@ -97,52 +95,16 @@ impl<R> From<R> for DebugMacInfo<R> {
     }
 }
 
-/// an Entry in the `.debug_macinfo` section.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum DebugMacInfoItem<R> {
-    /// A macro definition.
-    Define {
-        /// The line number where the macro is defined.
-        line: u64,
-        /// The text of the macro: The name of the macro followed immediately by any formal
-        /// parameters including the surrounding parentheses, followed by the macro definition.
-        text: R,
-    },
-    /// A macro undefinition.
-    Undef {
-        /// The line number where the macro is undefined.
-        line: u64,
-        /// The name of the macro without the definition.
-        name: R,
-    },
-    /// The start of a file.
-    StartFile {
-        /// Line number of the source file on which the inclusion macro directive occurred.
-        line: u64,
-        /// An index into the line number table of the compilation unit.
-        file: u64,
-    },
-    /// The end of the current included file.
-    EndFile,
-    /// A vendor-specific extension.
-    VendorExt {
-        /// A numeric constant, whose meaning is vendor specific.
-        numeric: u64,
-        /// A string whose meaning is vendor specific.
-        string: R,
-    },
-}
-
 /// Iterator over the entries in the `.debug_macinfo` section.
 #[derive(Clone, Debug)]
-pub struct DebugMacInfoIterator<R: Reader> {
+pub struct MacInfoIter<R: Reader> {
     input: R,
 }
 
-impl<R: Reader> DebugMacInfoIterator<R> {
+impl<R: Reader> MacInfoIter<R> {
     /// Advance the iterator to the next entry in the `.debug_macinfo` section.
-    pub fn next(&mut self) -> Result<Option<DebugMacInfoItem<R>>> {
-        // Read the next entry from the input reader and return it as a DebugMacInfoItem.
+    pub fn next(&mut self) -> Result<Option<MacroEntry<R>>> {
+        // Read the next entry from the input reader and return it as a DebugMacroItem.
         let macinfo_type = DwMacInfo(self.input.read_u8()?);
         match macinfo_type {
             constants::DW_MACINFO_null => {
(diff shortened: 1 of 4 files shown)


src/read/macros.rs
```diff
@@ -194,10 +195,351 @@ impl<R: Reader> fallible_iterator::FallibleIterator for DebugMacInfoIterator<R>
...
+            Format::Dwarf32
+        } else {
+            Format::Dwarf64
+        }
+    }
+}
+
+/// an Entry in the `.debug_macro` section.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DebugMacroItem<R, Offset = <R as Reader>::Offset>
```
philipc, line 336: Is it possible to share an `enum` with `.debug_macinfo`? That would make it simpler for callers to handle both sections.

src/read/macros.rs
```diff
@@ -194,10 +195,351 @@ impl<R: Reader> fallible_iterator::FallibleIterator for DebugMacInfoIterator<R>
...
+    R: Reader<Offset = Offset>,
+    Offset: ReaderOffset,
+{
+    /// A macro definition.
+    Define {
+        /// The line number where the macro is defined.
+        line: u64,
+        /// The text of the macro: The name of the macro followed immediately by any formal
+        /// parameters including the surrounding parentheses, followed by the macro definition.
+        text: R,
```
philipc, line 347: Instead of having `Define`, `DefineStrp`, `DefineStrx`, and `DefineSup`, can we combine them into a single `Define` variant that has `text: MacroString`, where `MacroString` is an enum for the various ways that a string can be encoded. Similarly for `Undef`.

src/read/macros.rs
DanielT: I've combined the enums as you proposed. Both iterators return a DebugMacroItem now.
In the combined enum, VendorExt is only used for .debug_macinfo, and Include / IncludeSup are only used by .debug_macro

DanielT:
Updated as requested

src/read/macros.rs
```diff
@@ -97,23 +98,267 @@ impl<R> From<R> for DebugMacInfo<R> {
...
+    fn parse(input: &mut R) -> Result<Self> {
+        let version = input.read_u16()?;
+        let flags = input.read_u8()?;
+        let format = if flags & Self::OFFSET_SIZE_FLAG == 0 {
+            Format::Dwarf32
+        } else {
+            Format::Dwarf64
+        };
+        let debug_line_offset = if flags & Self::DEBUG_LINE_OFFSET_FLAG != 0 {
+            DebugStrOffsetsBase(input.read_offset(format)?)
```
philipc, line 280: The standard says:
```
3. debug_line_offset
An offset in the .debug_line section of the beginning of the line number
information in the containing compilation, encoded as a 4-byte offset for a
32-bit DWARF format macro section and an 8-byte offset for a 64-bit DWARF
format macro section.
```

So treating this as a `DebugStrOffsetsBase` seems wrong to me.

I expect that this offset is meant to be used as a convenient way to find the line number table containing the file names for the start/end file entries.

src/read/macros.rs
```diff
@@ -97,23 +98,267 @@ impl<R> From<R> for DebugMacInfo<R> {
...
+                unit.string(str_offset)
+            }
+            MacroString::Supplementary(offset) => unit.sup_string(*offset),
+        }
+    }
+}
+
+/// an Entry in the `.debug_macro` section.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DebugMacroItem<R, Offset = <R as Reader>::Offset>
```
philipc, line 343:
```suggestion
pub enum MacroEntry<R, Offset = <R as Reader>::Offset>
```

src/read/macros.rs
```diff
@@ -97,23 +98,267 @@ impl<R> From<R> for DebugMacInfo<R> {
     }
 }
 
-/// an Entry in the `.debug_macinfo` section.
+/// Iterator over the entries in the `.debug_macinfo` section.
+#[derive(Clone, Debug)]
+pub struct DebugMacInfoIterator<R: Reader> {
```
philipc, line 103:
```suggestion
pub struct MacInfoIter<R: Reader> {
```

src/read/macros.rs
```diff
@@ -135,69 +390,140 @@ pub enum DebugMacInfoItem<R> {
 
 /// Iterator over the entries in the `.debug_macinfo` section.
```
philipc, line 391:
```suggestion
/// Iterator over the entries in the `.debug_macro` section.
```

src/read/macros.rs
```diff
@@ -135,69 +390,140 @@ pub enum DebugMacInfoItem<R> {
...
 #[derive(Clone, Debug)]
-pub struct DebugMacInfoIterator<R: Reader> {
+pub struct DebugMacroIterator<R: Reader> {
```
philipc, line 393:
```suggestion
pub struct MacroIter<R: Reader> {
```

This is more consistent with other naming in gimli.

src/read/macros.rs
```diff
@@ -97,23 +98,267 @@ impl<R> From<R> for DebugMacInfo<R> {
...
+    pub fn string(&self, unit: UnitRef<'_, R>) -> Result<R> {
+        match self {
+            MacroString::Direct(s) => Ok(s.clone()),
+            MacroString::StringPointer(offset) => unit.string(*offset),
+            MacroString::IndirectStringPointer(base, index) => {
+                let str_offset = unit.dwarf.debug_str_offsets.get_str_offset(
+                    unit.header.format(),
+                    *base,
+                    *index,
+                )?;
```
philipc, line 333:
```suggestion
                let str_offset = unit.str_offset(*index)?;
```
It's a bit unclear in the standard, but my expectation is that it should be this.

philipc (changes requested):

src/read/macros.rs
DanielT: I have no practical examples to test this against, but I'm going with the description of DW_FORM_strx:
```text
As an indirect offset into the string table using an index into a table of
offsets contained in the .debug_str_offsets section of the object file
(DW_FORM_strx, DW_FORM_strx1, DW_FORM_strx2,
DW_FORM_strx3 and DW_FORM_strx4). The representation of a
DW_FORM_strx value is an unsigned LEB128 value, which is
interpreted as a zero-based index into an array of offsets in the
.debug_str_offsets section. [...]
```

So if the operand of `DW_MACRO_define_strx` is of the type `DW_FORM_strx`, then a `DebugStrOffsetsBase` is needed for that value to make sense.

This is not something I feel strongly about though, so if you confirm that you think your reading is more likely to be correct, then I'm happy to go with that.

DanielT:
I've renamed the items as suggested, but I've left the handling of the _strx macros unchanged in this version

philipc:
I think my requested _strx changes are correct.  Here's a test file from the llvm project:
[debug-macro-multi-cu-strx.o.gz](https://github.com/user-attachments/files/20200051/debug-macro-multi-cu-strx.o.gz)
([source code](https://github.com/llvm/llvm-project/blob/main/llvm/test/DebugInfo/X86/debug-macro-multi-cu-strx.s))

This PR currently does not give the same macro strings as llvm-dwarfdump for that test file.

DanielT:
I've changed the code as you proposed and the output of dwarfdump on the LLVM testcase looks correct.

src/read/macros.rs
philipc: This still needs to be changed to `DebugLineOffset`.

DanielT:
updated: the header field is a DebugLineOffset instead of a DebugStrOffsetsBase

philipc (approved):

PR #764, opened 2025-05-29
Remove #[non_exhaustive] from read::CallFrameInstruction

gtsiam:
Using `#[non_exhaustive]` disallows users of the library from exhaustively matching the enum. This is nice for semver, but terrible for library users who are now in the awkward position of having to remember to update their exhaustive match when updating gimli.

This is less than ideal. I would know, as I now have to deal with it.

PS: `RegisterRule` is also highly suspect, but as it hasn't yet created problems (for me at least), so I'll leave it be for now.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -2958,7 +2958,6 @@ impl<T: ReaderOffset> RegisterRule<T> {
 
 /// A parsed call frame instruction.
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[non_exhaustive]
 pub enum CallFrameInstruction<T: ReaderOffset> {
     // 6.4.2.1 Row Creation Methods
     /// > 1. DW_CFA_set_loc


bjorn3:
The `non_exhaustive_omitted_patterns` lint exists for precisely this reason. Unfortunately it is still unstable though.

src/write/cfi.rs
```diff
@@ -378,7 +378,6 @@ impl FrameDescriptionEntry {
 ///
 /// This may be a CFA definition, a register rule, or some other directive.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-#[non_exhaustive]
```
bjorn3, line 381: Why remove it on the writer side too? You should only need to create `write::CallFrameInstruction` values as user of gimli, not match on them.

philipc (approved):
Thanks!

Yes, this is fine. There are many other similar enums for which this crate doesn't have this attribute, so it doesn't make sense to single this out (and I would be fine with `RegisterRule` and `write::CallFrameInstruction` if someone wanted it).

There are also a few other enums where do we have `non_exhaustive` but I think those ones make sense to have.

Issue #765, opened 2025-06-03
Non-zero directory index on associated compilation file for Dwarf5 Line Program


crzysdrs:
There is an implicit assumption that LinePrograms will only have compilation directories that are the zeroeth index. When converting LinePrograms expect this to be true resulting in the error on the following line: 

https://github.com/gimli-rs/gimli/blob/c2e0ec162477435e3d130db3095afad5581970a9/src/write/line.rs#L1028

I've got a workaround of removing the check for the zero index (avoiding the error) and instead using it to produce the comp_dir though I suspect this may not be the entirely correct solution.

[cpp_test.zip](https://github.com/user-attachments/files/20560260/cpp_test.zip)

Gimli version 28.1.

PR #767, opened 2025-06-03
Remove check for compilation directory matching first file directory.

crzysdrs:
Fix for issue #765.

Use the directory index zero, which is the correct compilation directory and ignore the compilation file provided index.

In the testcase attached to the issue, the directory that the compilation was invoked in was `/`. The file itself was provided as an absolute path `/usr/src/lfr-test/gdb//fibonacci.c`. It's not strictly the case that the compilation file will have a directory that matches the compilation directory.

Simplified Compilation instruction invoked from `/`:
```
gcc  -gdwarf-5 -fPIC -c -o /tmp/fibonacci.o /usr/src/lfr-test/gdb//fibonacci.c
``` 

However, it does appear that the compilation directory being the first entry in the directories list assumption is upheld.

### DIFF
--- a/src/write/line.rs
+++ b/src/write/line.rs
@@ -1024,9 +1024,6 @@ mod convert {
 
                 let comp_name = match from_header.file(0) {
                     Some(comp_file) => {
-                        if comp_file.directory_index() != 0 {
-                            return Err(ConvertError::InvalidDirectoryIndex);
-                        }
                         LineString::from(comp_file.path_name(), dwarf, line_strings, strings)?
                     }
                     None => LineString::new(&[][..], encoding, line_strings),


philipc:
This change is what I tested in https://github.com/gimli-rs/gimli/issues/765#issuecomment-2933550771 and I don't think it is correct. If you have a source file in `/tmp/test` and run the gcc command in `/tmp/test/build`, gcc produces these files:
```
include_directories[  0] = "/tmp/test/build"
include_directories[  1] = ".."
file_names[  0]:
           name: "fibonacci.c"
      dir_index: 1
file_names[  1]:
           name: "fibonacci.c"
      dir_index: 1
```

If I then pass that DWARF through gimli with this PR, it results in these files:
```
include_directories[  0] = "/tmp/test/build"
include_directories[  1] = ".."
file_names[  0]:
           name: "fibonacci.c"
      dir_index: 0
file_names[  1]:
           name: "fibonacci.c"
      dir_index: 1
```

That first file_names entry is saying that the primary source file is "/tmp/test/build/fibonacci.c", which is wrong.

What we need is an API that lets you specify a separate directories for the working directory and primary source file directory. Currently `LineProgram::new` assumes these are identical, which is too limiting.

crzysdrs:
Yeah I figured this wasn't the best solution.

PR #775, opened 2025-06-19
Add constants obsolete after DWARF 1.1

stravant:
* The DWARF 1.1 specification includes some constants which are only relevant in that spec, and became obsolete in the later revisions of the spec.

* Currently the gimli does not include these constants because since DWARF 1.1 is a different structure and rarely relevant it does not support parsing of DWARF 1.1.

* However DWARF 1.1 is relevant to the decomp community. Since decomp tooling wants to depend on gimli anyways for DWARF 2+, it would be nice if it could also at least use gimli data structures for DWARF 1.1 rather than having to write something bespoke. Including the constants unique to DWARF 1.1 helps with that.

* This is unlikely to cause much confusion about what is supported since anyone who runs into DWARF 1.1 in practice is probably already deep in the weeds.

* The added values can be cross-referenced here in Python elftools to confirm that they are correct: https://github.com/eliben/pyelftools/blob/0c3283b3c872a0424bf4017a95ecee14aa6f17b8/elftools/dwarf/enums.py

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -209,6 +209,13 @@ dw!(
 DwTag(u16) {
     DW_TAG_null = 0x00,
 
+// DWARF 1 only, obsolete in DWARF 2+.
+    DW_TAG_global_subroutine = 0x06,
+    DW_TAG_global_variable = 0x07,
+    DW_TAG_local_variable = 0x0c,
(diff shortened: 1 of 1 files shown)


philipc (approved):
Thanks!

PR #776, opened 2025-06-28
Add a llvm-dwarfdump inspired --debug-info [OFFSET] argument to dwarfdump.

khuey:
This dumps only .debug_info and if the optional offset is provided only the subtree of DIEs starting at that offset.

This relies on the way parallel_output() outputs each unit to its own individual buffer that can be discarded partway through. Implementing it for .debug_types isn't possible at the moment because dump_types() outputs directly to stdout.

### DIFF
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -1196,8 +1196,10 @@ fn dump_entries<R: Reader, W: Write>(
 
         let mut indent = if baseline_depth > 0 && baseline_depth >= depth {
             break;
-        } else {
+        } else if depth > 0 {
             (depth - baseline_depth) as usize * 2 + 2
+        } else {
+            2
         };
 
         if !found_desired_offset {


khuey:
Example output:

```
khuey@zhadum:~/dev/gimli$ ./target/debug/dwarfdump --debug-info=0x0000015b -G target/debug/dwarfdump

.debug_info

UNIT<.debug_info+0x00000000>: length = 0x36e, format = Dwarf32, version = 4, address_size = 8, abbrev_offset = 0x0
< 3><0x0000015b GOFF=0x0000015b>  DW_TAG_structure_type
                                    DW_AT_name                  PhantomData<core::hash::sip::Sip13Rounds>
                                    DW_AT_byte_size             0x00000000
                                    DW_AT_accessibility         DW_ACCESS_public
                                    DW_AT_alignment             0x00000001
< 4><0x00000163 GOFF=0x00000163>    DW_TAG_template_type_parameter
                                      DW_AT_type                  0x00000113<.debug_info+0x00000113>
                                      DW_AT_name                  T
< 4><0x0000016c GOFF=0x0000016c>    DW_TAG_null
< 3><0x0000016d GOFF=0x0000016d>  DW_TAG_null
```

crates/examples/src/bin/dwarfdump.rs
```diff
@@ -1139,22 +1180,38 @@ fn dump_entries<R: Reader, W: Write>(
...
         let depth = entries.next_depth();
         let abbrev = entries.read_abbreviation()?;
 
-        let mut indent = if depth >= 0 {
-            depth as usize * 2 + 2
+        let mut indent = if baseline_depth > 0 && baseline_depth >= depth {
+            break;
         } else {
-            2
+            (depth - baseline_depth) as usize * 2 + 2
```
philipc, line 1200: This can now be negative if `baseline_depth == 0 && depth < 0`, which is possible for malformed DWARF. It would be nice to keep the previous behaviour for this.

khuey: Ah, I was wondering what that was for.

philipc (approved):

khuey:
Thanks!

PR #783, opened 2025-08-15
Make EndianSlice::new() a const function

d-e-s-o:
Declare `EndianSlice::new()` as a const function, to make it usable from const contexts.


d-e-s-o:
Hm...I suppose that's not possible with the current MSRV :-(

d-e-s-o:
It appears to work with `1.61`. But it's not super important to me, as I can trivially work around.

PR #784, opened 2025-08-18
Keep the string form of `DW_LNCT_LLVM_source` on conversion

promaen:
With the existing conversion from a readable DWARF representation to one that is writable the `DW_LNCT_LLVM_source` string form is kept. When an empty source string is needed, it currently uses `DW_FORM_string` and a mismatch with `DW_LNCT_LLVM_source` results in an error when serializing.

Instead of always using `DW_FORM_string` when emitting empty source entries, use the same form as `DW_LNCT_LLVM_source` has.  This is only done on conversion, when creating a writable DWARF the existing logic is kept, and source expects `DW_FORM_string`.

### DIFF
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1204,7 +1204,7 @@ where
             .any(|x| x.content_type == constants::DW_LNCT_LLVM_source)
     }
 
-    pub(crate) fn source_form(&self) -> Option<constants::DwForm> {
+    pub(crate) fn file_source_form(&self) -> Option<constants::DwForm> {
         self.file_name_entry_format
             .iter()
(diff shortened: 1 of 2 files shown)


philipc (changes requested):
Thanks this is good. Just some cosmetic changes please.

promaen:
I have tried to address the changes you suggested. For the test, I have been unable to locate the referred place from the diff provided.

philipc:
Those changes look good. It's [this test](https://github.com/gimli-rs/gimli/blob/7335f00e7c39fd501511584fefb0ba974117c950/src/write/line.rs#L1227).

PR #786, opened 2025-08-20
write/line: support any form for file sources

### LINKED ISSUE #782: Reading then writing converted DWARF data can causes `Error::LineStringFormMismatch`
When processing DWARF data originating from a dSYM, reading and converting it to a writable `gimli::write::Dwarf` followed by a write results in `Error::LineStringFormMismatch`. The `DW_AT_producer` values are `"Apple clang version 17.0.0 (clang-1700.0.13.5)"` and `"Apple Swift version 6.1.2 effective-5.10 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)"`.

The cause appears to originate from this call site: https://github.com/gimli-rs/gimli/blob/71720829ef6f61ea5b0d4e1b9315534d2c8b9264/src/write/line.rs#L657-L670 and manifested here: https://github.com/gimli-rs/gimli/blob/71720829ef6f61ea5b0d4e1b9315534d2c8b9264/src/write/line.rs#L881-L892

The assumption here is that `source` is of form `constants::DW_FORM_string`. When this is not the case the precondition in `LineString::write` is broken. Passing `source.form()` instead avoids this issue for the described setup, but if that is correct or not is unclear. I see there is a comment on only supporting the `LineString::String` variant for the `source` field of `FileInfo`. In the failing cases the `source` is of type `LineString::LineStringRef`.

Are there any plans of supporting all the variants of `LineString` that can occur through conversions? Or perhaps, converting the input's `LineString` to `FileInfo`'s supported variant when using `gimli::write::Dwarf::from`?

philipc:
We can rely on the fact that if any source is set then the corresponding string table will have a null byte in it, and we can point to that for the empty string.

This is a bit of a hack, but it's workaround for an extension to DWARF 5 only. DWARF 6 should have a better way of handling this.

Closes #782


philipc:
@promaen Could you please test that this still works for your use?

PR #788, opened 2025-08-22
Release 0.32.1

philipc:


bugadani:
I believe this turned out to be a breaking update, see https://github.com/probe-rs/probe-rs/pull/3535 for an automated update PR, with an associated [build failure](https://github.com/probe-rs/probe-rs/actions/runs/17211988208/job/48825900476?pr=3535).

philipc:
Those new `PartialEq` impls don't even seem to be used. Would you agree that we should remove them and do another patch release?

bugadani:
> Would you agree that we should remove them

To me, this is always a tricky question. I can't tell why they were introduced, I'm not familiar enough with gimli. I'm not the one to say whether something should be kept or removed - I can only point at a possibly unintended build break :)

I'm personally fine with either rebranding the current state as 0.33, or reverting the breaking parts of the changes and releasing the result as 0.32.2, too.

PR #791, opened 2025-08-29
Accept -2 for tombstone values

philipc:
This has been seen in .debug_line, but the reason why is unknown. It is also unknown whether it occurs in other sections too, but allowing it for those is unlikely to cause a problem.


loewenheim (commented):
This looks sensible to me. Do you want further confirmation of where the -2 comes from before proceeding with this?

PR #797, opened 2025-09-11
Add support for DW_FORM_data16

crzysdrs:
`DW_FORM_data16` provided to `DW_AT_upper_bound` was being incorrectly being converted to an `AttributeValue::Block` causing the `exprloc_value` function to interpret it as a bunch of bytes representing an expression. This expression when parsed was an invalid set of bytes.

This can be demonstrated by running the `simple_convert` example over the following testcase.

Attached Example:
[dw_form.tar.gz](https://github.com/user-attachments/files/22285849/dw_form.tar.gz)

It should be noted that the example also has line programs that are newer versions than the units they are embedded in and will fail a consistency check when writing the `LineProgram`. This can be ignored by commenting out the check in `src/write/line.rs`. See #798.

Also we are using a cherry-picked version of https://github.com/gimli-rs/gimli/pull/787

### DIFF
--- a/src/read/unit.rs
+++ b/src/read/unit.rs
@@ -4400,7 +4400,7 @@ mod tests {
         ];
         let unit = test_parse_attribute_unit_default();
         let form = constants::DW_FORM_data16;
-        let value = AttributeValue::Data16(0x10_0f_0e0d_0c0b_0a09_0807_0605_0403_0201);
+        let value = AttributeValue::Data16(0x100f_0e0d_0c0b_0a09_0807_0605_0403_0201);
         test_parse_attribute(&buf, 16, &unit, form, value);
     }
 


philipc (changes requested):
Can you also add some tests for this? (They should have existed already but got missed). Two tests similar to the data8 tests should be enough ([`read`](https://github.com/gimli-rs/gimli/blob/146b123a96c58416476780d8dca36eccc4c580cc/src/read/unit.rs#L4387) and [write](https://github.com/gimli-rs/gimli/blob/146b123a96c58416476780d8dca36eccc4c580cc/src/write/unit.rs#L2322-L2326)).

Also note that the line program parser still reads `DW_FORM_data16` as a block, but we want it to keep doing that because it's used for MD5 hashes.

PR #803, opened 2025-09-19
add vector registers and AFRC to gimli::Register

marxin:


philipc (approved):
Thanks. Looking into it more, I don't think AFRC ever appears in the binary (CFI or DWARF). I guess it doesn't hurt though.

marxin:
Thank you. Any estimation when will you chop a new release?

Issue #469, opened 2020-01-22
Support for reading DWARF 5 .debug_names


luser:
I was looking up info on GDB's `.gdb_index` section today and found that it has been superseded by the DWARF 5 `.debug_names` section. gdb ships a helper tool (`gdb-add-index`) that can generate and add a `.gdb_index` section to a binary and in recent versions (I'm using gdb 8.3 on Ubuntu) it can instead generate `.debug_names`:
```
$ gdb-add-index -dwarf-5 hello
$ objdump --wide --section-headers hello- | grep debug_names
 32 .debug_names       000001b8  0000000000000000  0000000000000000  00003847  2**0  CONTENTS, READONLY, DEBUGGING
```

The `.debug_names` section format is specified in the DWARF 5 spec section 6.1.1, "Lookup by Name".

VorpalBlade:
@philipc (and other maintainers) I'm curious as to what would be needed to implement basic support for reading this section. I might consider this, though I will probably not add support for the hash table, just the basic linear tables.

Would that be a useful contribution, or would you want only the full support?

PR #807, opened 2025-10-01
Read debug_names section from DWARF 5.

tmcgilchrist:
Fix for https://github.com/gimli-rs/gimli/issues/469. This adds support for reading the debug_names section in DWARF 5 as documented in the DWARF 5 spec "6.1 Accelerated Access".  

I also added a sample implementation to dwarfdump that prints the same information as llvm-dwarfdump for MachO binaries. The section structure is the same for ELF binaries

This is my first significant contribution to a Rust library, so the API might not be ideal and I might have un-idomatic Rust code. I would appreciate specific suggestions to make it better. 

It needs further testing on Linux ELF and some of the corner cases.


src/read/dwarf.rs
```diff
@@ -754,6 +765,147 @@ impl<R: Reader> Dwarf<R> {
...
+                        }
+                    }
+                }
+            }
+        }
+
+        Ok(results)
+    }
+
+    /// Compute the DJB hash for a name string following the DWARF 5 specification.
```
VorpalBlade, line 897: Add a note that this is defined in section 7.33

src/read/names.rs
```diff
@@ -0,0 +1,2641 @@
...
+    /// This offset is relative to the start of the compilation unit's
+    /// debug information and can be used to locate the full DIE data.
+    pub die_offset: u32,
+
+    /// Parent information for hierarchical relationships.
+    ///
+    /// Contains the absolute offset of the parent entry in the entry pool
+    /// if this entry has a parent relationship. None indicates either
+    /// no parent (top-level entry) or that parent information is not indexed.
+    pub parent_info: Option<u64>,
```
VorpalBlade, line 156: I don't see a list of what attributes are allowed / required / expected for entries in the entry pool.

From 6.1.1.4.8:

> Each index entry in the series begins with an abbreviation code, and is followed
> by the attributes described by the abbreviation declaration for that code. The last
> index entry in the series is followed by a terminating entry whose abbreviation
> code is 0.

So I'm a bit confused by the ones you have added here. Would it be better to expose a more generic way to iterate over the attributes, just like is done for `.debug_info`?

src/read/names.rs
```diff
@@ -0,0 +1,2641 @@
...
+    /// Number of buckets in the hash table.
+    bucket_count: u32,
+    /// Number of unique name entries.
+    name_count: u32,
+    /// Size of the abbreviations table in bytes.
+    abbrev_table_size: u32,
+    /// Augmentation string size.
+    augmentation_string_size: u32,
+    /// The format of the unit.
+    format: Format,
```
VorpalBlade, line 369: I don't see format listed in 6.1.1.4.1?

src/read/names.rs
```diff
@@ -0,0 +1,2641 @@
...
+
+/// The header of a `.debug_names` unit.
+#[derive(Debug, Clone)]
+pub struct DebugNamesHeader<R: Reader> {
+    /// Unit length of this name index table.
+    length: R::Offset,
+    /// Version of the name index table format (should be 5 for DWARF 5).
+    version: u16,
+    /// Padding for alignment.
+    padding: u16,
```
VorpalBlade, line 353: Why is the padding here? This is not `repr(C)`, so we aren't trying to zero copy parse this data.

The spec says:

> padding (uhalf)
> Reserved to DWARF (must be zero).

src/read/names.rs
```diff
@@ -0,0 +1,2641 @@
...
+}
+
+impl<R: Reader> DebugNamesHeader<R> {
+    /// Return the version of this debug names table.
+    pub fn version(&self) -> u16 {
+        self.version
+    }
+
+    /// Return the amout of padding in this index.
+    pub fn padding(&self) -> u16 {
```
VorpalBlade, line 381: I think this is a misunderstanding, the field in the header is itself padding, not a field indicating the amount of padding found elsewhere.

crates/examples/src/bin/dwarfdump.rs
```diff
@@ -2469,3 +2487,453 @@ fn dump_macro<R: Reader, W: Write>(
...
+    // Add file format line to match system dwarfdump
+    let file_format = match file.format() {
+        object::BinaryFormat::Elf => "ELF",
+        object::BinaryFormat::MachO => match file.architecture() {
+            object::Architecture::Aarch64 => "Mach-O arm64",
+            object::Architecture::X86_64 => "Mach-O x86-64",
+            object::Architecture::I386 => "Mach-O i386",
+            _ => "Mach-O",
+        },
+        object::BinaryFormat::Pe => "PE",
```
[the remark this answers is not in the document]
bjorn3: Yes, the MinGW toolchain uses this. GDB doesn't support pdb files.

VorpalBlade:
> > I don't know what policy this project have on using AI tools.
> 
> No formal policy yet. I haven't noticed any prior AI contributions. I'd expect the following at least:
> 
> * Be up front about the use of AI. The PR should state how AI was used to assist.

Not done in this case. 

> * All AI generated code must be fully reviewed by the author prior to submission.

I think for the most part this was done, but perhaps not for the code documentation comments. (Or perhaps AI was only used for the comments?) The issues in the code itself could plausibly have been human errors that we all make.

The *obvious* AI generated parts were all in the documentation, which made me scrutinise it closer and start cross checking more references and find more issues. 

> Don't waste maintainer time on reviewing low quality AI generated code.
 
I don't think this PR necessarily feels low quality to me. It just feels like the AI wasn't fully checked. I would say it needs a bit of work. Plus checking by someone more familiar with gimli's internals than me. For example: There were some confusing constructs in the unit tests, but I have not compared to other unit tests in gimli to see how they work.

philipc:
> I don't think this PR necessarily feels low quality to me.

Just to be clear (since my previous comment wasn't): I wasn't making a judgement on this PR, just saying what the policy would be. I still haven't done a review.

crates/examples/src/bin/dwarfdump.rs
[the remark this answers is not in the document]
tmcgilchrist: There's some support for adding DWARF sections to PE/COFF executables in clang-cl and lldb has basic support for DWARF on Windows. It's an unusual combination for sure. I'll remove it.

tmcgilchrist:
Apologies for the confusion. I used AI tooling to extract/summarise parts of the DWARF spec and add them as code comments. They're too wordy (or inaccurate) for the style of this library, however I do find the links back to specific sections in the DWARF spec useful.

The Rust errors are all my own, unfortunately. Thank you for the review comments, I'll continue working on this.

crates/examples/src/bin/dwarfdump.rs
```diff
@@ -2469,3 +2487,453 @@ fn dump_macro<R: Reader, W: Write>(
...
+                    // Hit a hash from different bucket, stop chain traversal for this bucket
+                    break;
+                }
+            }
+        }
+    }
+
+    // Group entries by resolved string content to handle multiple entries per name
+    // Following DWARF 5 spec: "If two different symbol names produce the same hash value,
+    // that hash value will occur twice in the hashes array"
```
philipc, line 2791: I don't understand why we need to do this. The hash table is grouped by name, and its entries point back to the name, so we already have this grouping. Why are we building up these maps instead of simply iterating over the buckets as we display?

src/read/names.rs
```diff
@@ -0,0 +1,2641 @@
...
+
+/// A single name index table from the `.debug_names` section.
+///
+/// Each unit corresponds to one name index table within the debug_names section.
+/// It provides access to the compilation unit table, type unit tables,
+/// hash buckets, and name entries that make up the accelerated lookup structure.
+#[derive(Debug)]
+pub struct DebugNamesUnit<R: Reader> {
+    header: DebugNamesHeader<R>,
+    content: R,
```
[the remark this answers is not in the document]
tmcgilchrist: Ok, I didn't understand this pattern on my first reading of the code.

If I understand correctly I should store pre-sliced Reader instances for each section in `DebugNamesUnit<R>`, similar to the pattern in CFI (CommonInformationEntry, FrameDescriptionEntry).  For example:

```rust
pub struct DebugNamesUnit<R: Reader> {
    // Metadata
    version: u16,
    format: Format,
    augmentation_string: R,

    // Counts  for iteration
    comp_unit_count: u32,
    local_type_unit_count: u32,
    foreign_type_unit_count: u32,
    bucket_count: u32,
    name_count: u32,
    abbrev_table_size: u32,

    // Pre-sliced readers for each section
    comp_unit_list: R,
    local_type_unit_list: R,
    foreign_type_unit_list: R,
    hash_table_data: R,
    name_table_data: R,
    abbreviation_table: R,
    entry_pool: R,
}
```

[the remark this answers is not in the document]
philipc: Basically yes, but you should be able to avoid a lot of the counts. Probably the main one you need to keep is `bucket_count`, since that is needed for converting a hash to a bucket index. For the others, we only need to look entries up by index.

Related to this, I see a lot of functions for reading these tables and storing them as `Vec`s. We should avoid that for simple arrays. The abbreviation table will need to be parsed into another data structure, but the rest should be kept as `R` instead of parsing them into a `Vec`.

I haven't looked at this code closely, so I may be missing something, but the goal is to be zero copy where possible. These tables can potentially be quite large, and the user will only need to access a few entries at a time.

tmcgilchrist:
Thanks @philipc it's unlikely I'll have time to work on this and I don't have a good sense of how to match the Rust-style in this project. I had one more change [8148020](https://github.com/gimli-rs/gimli/pull/807/commits/8148020320216fe76f62fd64c42b2b3773bde4a1) that might be useful, which gets rid of some hardcoded values in favour of Labels. 

Good luck with the implementation, looking forward to reading the finished code :-)

Issue #806, opened 2025-09-23
Add an example of pruning dead code from the DWARF


philipc:
Once #804 is complete, it should be possible to write a simple conversion example that prunes DWARF containing tombstones.

PR #819, opened 2025-10-29
write: add ConvertUnit

### LINKED ISSUE #804: Lower level conversion API
The current conversion API is basically [`Dwarf:::from`](https://docs.rs/gimli/0.32.3/gimli/write/struct.Dwarf.html#method.from) (and similar methods on other types that are called by this). This method allows the caller to translate addresses, but nothing else. It would be useful to support other transformations during conversion. Currently the transform code in wasmtime duplicates a lot of `Dwarf::from`, but there is a desire to use the code in gimli instead (https://github.com/bytecodealliance/wasmtime/issues/5537).

The conversion needs to support at least:
- translation of addresses (not necessarily a one to one mapping)
- rewriting of expressions
- reshaping the DIE tree (pruning entries, or inserting new entries)
- replacing skeleton units with the DWO units

We should also try not to limit what can be done. I think this is best done by providing lower level APIs for possible actions during the conversion, while leaving the high level control flow to the caller.

The existing `Dwarf::from` will be kept unchanged for those who don't need the flexibility (but it will be internally rewritten to use these new APIs).

#745 provides an implementation of this for the line program conversion, but we need to support converting the DIE tree too (and the things it references).

philipc:
This allows reuse of the implementation of `Dwarf::from` as part of a more complex transformation.

This also provides support for conversion of split DWARF units.

Closes #804

### DIFF
--- a/src/write/dwarf.rs
+++ b/src/write/dwarf.rs
@@ -134,11 +134,12 @@ pub(crate) mod convert {
         ///
         /// `convert_address` should not be used for complex address transformations, as it
         /// will not be called for address offsets (such as in `DW_AT_high_pc`, line programs,
-        /// location lists, or range lists).
+        /// location lists, or range lists). If you need complex transformations, then you
+        /// need to use [`Dwarf::convert`] to enable you to transform the address offsets too.
         ///
         /// ## Example
         ///
-        /// Convert a DWARF section using `Dwarf::from`.
+        /// Convert DWARF sections using `Dwarf::from`.
         ///
         /// ```rust,no_run
         /// # fn example() -> Result<(), gimli::write::ConvertError> {
@@ -156,7 +157,7 @@ pub(crate) mod convert {
             convert_address: &dyn Fn(u64) -> Option<Address>,
         ) -> ConvertResult<Dwarf> {
             let mut dwarf = Dwarf::default();
-            let mut convert = dwarf.convert(from_dwarf, None)?;
+            let mut convert = dwarf.convert(from_dwarf)?;
             while let Some((mut unit, root_entry)) = convert.read_unit()? {
                 if let Some(convert_program) = unit.read_line_program(None, None)? {
                     let (program, files) = convert_program.convert_all(convert_address)?;
@@ -183,7 +184,7 @@ pub(crate) mod convert {
         ///
         /// ## Example
         ///
-        /// Convert a DWARF section using `convert`.
+        /// Convert DWARF sections using `convert`.
         /// See [`ConvertUnit`](crate::write::ConvertUnit) for an example of the unit
         /// conversion.
         ///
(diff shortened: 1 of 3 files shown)


philipc:
https://github.com/gimli-rs/gimli/commit/848229401b810355e28236c1adca142d213f5762 has an example for a complete DWARF conversion. I still need to extend it to handle split DWARF and implement the dead code detection to fix #806.

https://github.com/philipc/wasmtime/commit/a4cc3f25cb07acea20684db9694991a0990554d4 updates wasmtime to use this transformation support. It does result in some differences in the transformed DWARF, so I still need to spend some time to ensure those differences are all good.

src/write/dwarf.rs
```diff
@@ -129,49 +131,151 @@ pub(crate) mod convert {
...
+        /// let read_dwarf = gimli::read::Dwarf::load(loader)?;
+        /// let mut write_dwarf = gimli::write::Dwarf::new();
+        /// let mut convert = write_dwarf.convert(&read_dwarf, None)?;
+        /// while let Some((mut unit, root_entry)) = convert.read_unit()? {
+        ///     // Now you can convert the root DIE attributes, and other DIEs.
+        /// }
+        /// # unreachable!()
+        /// # }
+        /// ```
+        // TODO: specify `encoding` per unit instead?
```
fitzgen: Not sure when you would want to have different units have different encodings in the same set of debug info, but since it is a per-unit property, it makes sense to me to reflect that in these APIs.

src/write/unit.rs
```diff
@@ -248,6 +248,36 @@ impl Unit {
         self.root
     }
 
+    /// Reserve a `DebuggingInformationEntry` in this unit and return its id.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `parent` is invalid.
+    pub fn reserve(&mut self) -> UnitEntryId {
```
fitzgen: I don't see any `parent` here?

src/write/unit.rs
```diff
@@ -248,6 +248,36 @@ impl Unit {
...
+    ///
+    /// Panics if `child` or `parent` is invalid, or if `child` is not a reserved entry.
+    pub fn add_reserved(&mut self, child: UnitEntryId, parent: UnitEntryId, tag: constants::DwTag) {
+        let entry = self.get_mut(child);
+        debug_assert_eq!(entry.parent, None);
+        debug_assert_eq!(entry.tag, constants::DW_TAG_null);
+        entry.parent = Some(parent);
+        entry.tag = tag;
+        self.get_mut(parent).children.push(child);
+    }
```
fitzgen: I think `parent` is just here. Was the previous `parent` reference just out-of-sync with how you ended up structuring the APIs?

fitzgen (approved):
Apologies for the delay, I was at the Wasm CG meeting last week.

Changes here and in that Wasmtime commit LGTM, just some little nitpicks/questions below.

Thanks!

philipc:
I haven't looked at performance at all yet, but would like to in future. We probably want to move towards a streaming API; write to sections as we go instead of building up structures.

PR #368, opened 2019-01-29
write/range: fix conversion of relocatable address ranges

philipc:

PR #835, opened 2025-11-18
Delete `AddressOrOffsetPair` for range and location lists

### LINKED ISSUE #810: Is `read::RawRngListEntry::AddressOrOffsetPair` correct?
For DWARF 4, we currently allow range list entries to use a pair of addresses. However, the DWARF 4 standard (section 2.17.3) only talks about address offsets. We should check if our handling is correct. Maybe we should use `read::RawRngListEntry::OffsetPair` instead for DWARF 4.

Related to this, `read::Unit::low_pc` has type `u64`. This may be better as `Option<u64>`, and then an offset pair without a base address can be detected as an error.

philipc:
`AddressOrOffsetPair` was a hack to handle the way we always used `Reader::read_address` for ranges in DWARF version <= 4. We were doing this because we needed to use `read_address` for the case where the range is setting a base address.

A better way to handle this is to only use `read_address` if the first value is !0, signifying that the next value is a base address.

As part of this change, we no longer support writing `StartEnd` or `StartLength` for DWARF version <= 4. Our previous support for this wasn't correct, because consumers would still add the base address.

Closes #810. Originally added in #368.

### DIFF
--- a/crates/examples/src/bin/dwarfdump.rs
+++ b/crates/examples/src/bin/dwarfdump.rs
@@ -1841,12 +1841,7 @@ fn dump_loc_list<R: Reader, W: Write>(
                 dump_exprloc(w, unit, data)?;
                 writeln!(w)?;
             }
-            gimli::RawLocListEntry::AddressOrOffsetPair {
-                begin,
-                end,
-                ref data,
-            }
-            | gimli::RawLocListEntry::OffsetPair {
+            gimli::RawLocListEntry::OffsetPair {
                 begin,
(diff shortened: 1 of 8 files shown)


philipc:
This is wrong. In practice, compiler may emit relocations for offset pairs. For example, it happens when using `gcc-14 -O3 -gdwarf-4 -ffunction-sections`, and with this change the dwarfdump example reads the two zeroes instead of the relocated pair of values, and ends the range/location list immediately.

PR #841, opened 2025-12-09
Change `read::DebuggingInformationEntry` to store attributes in a Vec

philipc:
The convert code has identified the need to be able to store a parsed DIE. Its operation requires that it first parse and examine the attributes, and then later the user processes them. Currently this is handled with a bit of code repetition in [`FilterUnitEntry`](https://github.com/gimli-rs/gimli/blob/25efa5c2d148dda0f3ebb44dd53a09a1c7eeec10/src/write/unit.rs#L1973-L1985) and [`ConvertUnitEntry`](https://github.com/gimli-rs/gimli/blob/25efa5c2d148dda0f3ebb44dd53a09a1c7eeec10/src/write/unit.rs#L2980-L2992).

Additionally, the API for `DebuggingInformationEntry` has [methods](https://github.com/gimli-rs/gimli/blob/25efa5c2d148dda0f3ebb44dd53a09a1c7eeec10/src/read/unit.rs#L962-L986) that encourage repeated parsing of the attributes, along with implementation complexity to enable this. However, in practice this is inefficient, and it's a reason why `EntriesRaw` was added.

This PR changes `DebuggingInformationEntry` so that attributes are parsed immediately and stored in a `Vec`. This solves both of the above: `read::DebuggingInformationEntry` can now be used in the convert code, and repeated parsing of attributes is avoided.

The downside is that repeated allocation of this `Vec` can be expensive. This is mitigated by storing the DIE in `EntriesCursor` and `EntriesTree` so that it can be reused for every DIE. This results in a net performance increase for `EntriesCursor`/`EntriesTree` users who repeatedly call `DebuggingInformationEntry::attr`. Other users may have a small performance loss that can be regained by switching to `EntriesRaw`.

However, `UnitHeader::entry` and `DebuggingInformationEntry::clone` now have an increased cost, with no simple way to mitigate it.

### DIFF
--- a/src/read/unit.rs
+++ b/src/read/unit.rs
@@ -587,9 +587,9 @@ where
         abbreviations: &'abbrev Abbreviations,
         offset: UnitOffset<Offset>,
     ) -> Result<DebuggingInformationEntry<R>> {
-        let mut input = self.range_from(offset..)?;
         let mut entry = DebuggingInformationEntry::null();
-        entry.parse(&mut input, self.encoding, abbreviations, offset)?;
+        let mut input = self.entries_raw(abbreviations, Some(offset))?;
+        input.read_entry(&mut entry)?;
         if entry.is_null() {
             Err(Error::NoEntryAtGivenOffset)
         } else {
@@ -777,6 +777,7 @@ where
     has_children: bool,
     attrs: Vec<Attribute<R>>,
     offset: UnitOffset<Offset>,
+    depth: isize,
 }
 
 impl<R, Offset> Default for DebuggingInformationEntry<R, Offset>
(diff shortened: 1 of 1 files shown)


fitzgen:
> @fitzgen Maybe you have some thoughts about this change.

Thanks for the thorough explanations and benchmarks above. The reasoning makes sense to me, and benchmarks seem like they are what we would expect.

> This is changing one of the earliest design decisions made in this crate.

That is true, however if I remember correctly, at the time that these methods were established we also didn't have `EntriesRaw` yet and these methods needed to balance being both the most-convenient and fastest APIs possible. But given that `EntriesRaw` exists now and continues to exist in the future as a low-level, maximum-control escape hatch that is free to sacrifice usability and convenience at the alter of machine sympathy, then the changes in this PR seem very reasonable to me.

Thanks @philipc!

PR #847, opened 2025-12-18
Read debug_names section from DWARF 5

### LINKED ISSUE #469: Support for reading DWARF 5 .debug_names

philipc:
Adds sample implementation to dwarfdump that prints the same information as llvm-dwarfdump.

Closes #469


src/read/names.rs
```diff
@@ -0,0 +1,1923 @@
...
+#[derive(Debug, Clone)]
+pub struct NameLookupResult<R: Reader> {
+    /// The entry from the debug_names section.
+    pub entry: NameEntry<R>,
+
+    /// The resolved name string for this entry.
+    pub name: String,
+
+    /// The offset to the compilation unit header in .debug_info.
+    pub compilation_unit_offset: DebugInfoOffset<R::Offset>,
```
philipc: Not all entries require a CU offset. We'll need to do something different here but I'm not sure what. We could remove `NameLookupResult` and return only `NameEntry`, but we still need some way to also convert the CU index from the attribute into a CU offset, and similarly for TUs.

Maybe `DebugNames::find_entries_by_name` should return an iterator rather than a `Vec`, and that iterator could keep some state for looking up the indexes.

src/read/names.rs
```diff
@@ -0,0 +1,1923 @@
...
+        &self,
+        name: &str,
+        debug_str: &DebugStr<S>,
+        tag_filter: Option<constants::DwTag>,
+    ) -> Result<Vec<NameLookupResult<R>>>
+    where
+        S: Reader<Offset = R::Offset>,
+    {
+        // Compute the hash for the name
+        let hash = compute_djb_hash(name);
```
philipc: The hash must be of the case folded name. That means either `name` is already case folded, or we need to the case folding ourselves.

src/read/names.rs
philipc: I've removed `NameLookupResult` and the code that uses it. Also `NameEntry` now has methods to lookup CU and TU. I think there is scope for a `find_by_name` method, but I'll leave it to be designed by someone that is using it. For now, crates/examples/src/bin/names.rs contains a simple example of how to find a name, but users might want to do things differently (cache the `NameIndex`, handle source language differences).

src/read/names.rs
philipc: Removed this code, and provided `case_folding_djb_hash` for the user to do the hash calculation for `find_by_hash`.

PR #854, opened 2026-01-02
Fix no_std with write

the-ssd:
This allows for this crate to be use in a no_std environment.

Tbh I haven't tested the changes, only changed import of indexmap


philipc:
That's sounds reasonable. I'll try to have a look to see what's involved. The CFI writing is fairly independent, so maybe we could limit the no_std support to only that for starters.

the-ssd:
Wow, thanks a lot

PR #875, opened 2026-04-14
Optimize unsigned LEB128 decoding more

d-e-s-o:
Optimize LEB128 decoding some more (after https://github.com/gimli-rs/gimli/pull/795). Improvements are more significant as per my testing. Please refer to individual commits.


philipc (approved):
Thanks! Yeah those traits rely a lot on the optimiser, there might more improvements to be found.

PR #884, opened 2026-06-09
Remove unnecessary clone

FranciscoTGouveia:
If I am not overlooking anything, this clone can be removed, and `read_path` can instead be moved.

CC @nunoplopes


FranciscoTGouveia:
I believe this was a spurious failure on the CI.

PR #885, opened 2026-06-11
read: pre-reserve vecs for include_directories and file_names

Jannik2099:
The grow + realloc for these vecs makes up
almost 40% of total parse() runtime for me

### DIFF
--- a/src/read/line.rs
+++ b/src/read/line.rs
@@ -1311,6 +1311,7 @@ where
             comp_dir = None;
             directory_entry_format = FileEntryFormat::parse(rest)?;
             let count = rest.read_uleb128()?;
+            include_directories.reserve(count.min(rest.len().into_u64()) as usize);
             for _ in 0..count {
                 include_directories.push(parse_directory_v5(
                     rest,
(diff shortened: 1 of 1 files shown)


philipc (changes requested):
Thanks for the PR. We need to be a bit careful here, since malformed data could result in a panic or a large memory allocation. The existing code is fine since it is limited by the section length, so we should have a similar limit for any pre-allocation.

Jannik2099:
good point - is this better?

Jannik2099:
Actually, I think this is incorrect and would reserve too few elements - I'll need to look at it with more time :/

src/read/line.rs
```diff
@@ -1311,6 +1311,11 @@ where
             comp_dir = None;
             directory_entry_format = FileEntryFormat::parse(rest)?;
             let count = rest.read_uleb128()?;
+            include_directories.reserve(
+                count.min(
+                    rest.len().into_u64() / std::mem::size_of::<AttributeValue<R>>() as u64 + 1,
```
philipc:
```suggestion
                    rest.len().into_u64()
```
I think this should be fine, doesn't need to be too exact.

Jannik2099: Isn't `AttributeValue` a few hundred bytes in size? This'd allow malformed data to create an allocation orders of magnitude bigger than possibly needed.

philipc: Depending on the reader, it's around 24-64 bytes. However, I think it's enough to match the existing code's worst case, which is one byte per encoded entry (e.g. a single null byte for every path; the size of `AttributeValue` doesn't bear any relation to the encoded size).

Jannik2099: I see, I guess that sounds acceptable

philipc (approved):

PR #887, opened 2026-06-11
Handle bare S CFI augmentation

jsgf:
.debug_frame CIEs may use augmentation string "S" as a DWARF signal-frame marker without the GNU "z" augmentation prefix. The "z" prefix is what indicates z-style augmentation data, where the CIE and associated FDEs carry augmentation payloads with a uleb128 length prefix.

Gimli treated any non-empty parsed CIE augmentation as if it used that z-style format. When parsing an FDE for a bare "S" .debug_frame CIE, it unconditionally tried to read an FDE augmentation-data length and failed with UnexpectedEof because no such data exists.

Track whether the CIE augmentation actually used the z-style format, and only parse FDE augmentation data for z-style augmentations. Bare "S" still marks the FDE as a signal trampoline without consuming nonexistent per-FDE augmentation data.

Add a regression test for the .debug_frame bare-S case.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -1136,7 +1136,26 @@ where
     }))
 }
 
-/// We support the z-style augmentation [defined by `.eh_frame`][ehframe].
+/// We support the z-style augmentation [defined by `.eh_frame`][ehframe],
+/// and the bare `S` signal-frame augmentation emitted by GNU as for
+/// `.debug_frame`.
+///
+/// In the z-style augmentation format, a leading `z` means that the CIE and
+/// its FDEs carry augmentation data with a ULEB128 length prefix. Augmentation
+/// characters that consume data, such as `L`, `P`, and `R`, are only accepted
+/// in this z-style format.
+///
+/// GNU as has emitted a bare `S` augmentation for `.debug_frame` signal frames
+/// since 2006. This is not specified by DWARF or the LSB `.eh_frame`
+/// augmentation documentation, but it is de-facto GCC toolchain behavior.
+/// `S` is payload-free: it only marks the frame as a signal trampoline, so it
+/// is safe to handle without reading z-style CIE or FDE augmentation data.
+///
+/// Only payload-free augmentation flags can be handled without a leading `z`.
+/// Currently this means bare `S`. Other payload-free flags, such as `B` or `G`,
+/// could be handled the same way if support is added, but data-bearing
+/// augmentations such as `L`, `P`, and `R` require z-style augmentation data
+/// and are not accepted bare.
 ///
 /// [ehframe]: https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-Core-generic/LSB-Core-generic/ehframechpt.html
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


philipc (commented):
I've tried to find documentation for this, without any success. Do you know which toolchains produce bare "S" augmentation strings, and if there is any documentation? It would be good to update the `Augmentation` documentation to say more than "We support the z-style augmentation defined by `.eh_frame`.".

I also looked at some other parsers, and their handling is inconsistent. Some do handle this case, but at least LLVM looks like it would give a parsing error (I haven't tested though). At least one parser even handles things like bare "L", so I'm wondering if there are more fixes required here.

jsgf:
Looks like it's `gas` that does this.

```asm
# bare-s.s
.cfi_sections .debug_frame
.globl handler
.type handler, @function
handler:
  .cfi_startproc
  .cfi_signal_frame
  nop
  .cfi_endproc
```

```bash
# gas - target doesn't seem to matter (tried riscv + x86)
$ as -o bare-s.o bare-s.s

$ readelf --debug-dump=frames bare-s.o 
Contents of the .debug_frame section:


00000000 0000000000000014 ffffffff CIE
  Version:               1
  Augmentation:          "S"
[...]
```

I got Claude to research why gas does this, why other tools seem to work and if there's any documented expectations:

---

**Why GAS does this:** In GAS source ([`gas/dw2gencfi.c`](https://sources.debian.org/src/binutils/2.46-1/gas/dw2gencfi.c#L1958-L1959)), the `z` prefix and `P/L/R` augmentation characters are written only in the `.eh_frame` code path. The `S` flag is appended independently, outside the `.eh_frame`-specific block. For `.debug_frame`, GAS skips `z/P/L/R` entirely but still appends `S` if `.cfi_signal_frame` was used → bare `"S"`.

**Documentation status:** Bare "S" is **not documented by any spec**:
- DWARF 5 §6.4.1 defines the CIE augmentation field as a null-terminated string but defines none of z/L/P/R/S — vendor extensions are expected.
- The LSB Core "Exception Frames" chapter defines z/L/P/R for `.eh_frame` only. `L/P/R` are defined in the context of a leading `z`. The LSB does not define `S` at all.
- `S` was introduced by Jakub Jelinek (2006, GCC 4.3 / Binutils 2.16) with `.cfi_signal_frame` — so the unwinder uses the exact interrupted PC rather than RA-1 for FDE lookup.

GCC's libgcc (`unwind-dw2.c`, `extract_cie_info`) processes augmentation chars in a loop NOT gated on the `z` prefix, so bare `S` works. LLVM libunwind (`DwarfParser.hpp`, `parseCIE`) wraps the entire augmentation switch inside `if (starts with 'z')`, so it silently ignores bare `S`.

**Should bare L/R/P also be handled?** No. `L/R/P` each consume data bytes whose presence and length are announced by the `z` augmentation-data-length ULEB128. Without `z`, there is no data to consume and the layout is undefined. No mainstream producer emits bare `L/R/P` — GAS only writes them inside the `.eh_frame` branch, paired with `z`. The only other payload-free augmentation chars that could theoretically appear bare are `B` (AArch64 PAuth B-key) and `G` (MTE).

philipc (approved):
Thanks!

PR #888, opened 2026-06-13
feat: provide flags for PubNamesEntry/PubTypesEntry

marxin:
I would like to use the parser in the Wild linker and I just need to access the entire `u8` `flags` - no need for the actual parsing of the value.

### DIFF
--- a/src/read/lookup.rs
+++ b/src/read/lookup.rs
@@ -177,6 +177,10 @@ impl<R: Reader> PubSetEntry<R> {
         GdbIndexSymbolKind((self.flags >> 4) & 7)
     }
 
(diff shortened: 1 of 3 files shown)


philipc (approved):
Thanks!

philipc:
By the way, are the separate types for PubNamesEntry and PubTypesEntry problematic? I'm wondering if we should change them to type aliases for PubSetEntry.

marxin:
> By the way, are the separate types for PubNamesEntry and PubTypesEntry problematic?

Probably not, I decided to use just one parser for both of the sections as we don't need to go into the details of the `flags` encoding:
https://github.com/wild-linker/wild/pull/2060

PR #889, opened 2026-06-15
read: avoid overflow when computing index-based section offsets

alhudz:
Repro: resolve a `DW_FORM_addrx`/`strx`/`loclistx`/`rnglistx` whose index is around `0x2000_0000_0000_0000` via `DebugAddr::get_address` / `DebugStrOffsets::get_str_offset` / `LocationLists::get_offset` / `RangeLists::get_offset`.
Cause: each computes `index * size` (address size or DWARF word size) as an unchecked `u64` multiply before `from_u64`, so a large index wraps. With overflow checks that panics; in release it skips to a wrong in-bounds offset.
Fix: use `checked_mul` and return `Error::UnsupportedOffset` on overflow, the same way an out-of-range offset is already reported.

### DIFF
--- a/src/read/addr.rs
+++ b/src/read/addr.rs
@@ -31,7 +31,11 @@ impl<R: Reader> DebugAddr<R> {
         let input = &mut self.section.clone();
         input.skip(base.0)?;
         input.skip(R::Offset::from_u64(
-            index.0.into_u64() * u64::from(address_size),
+            index
+                .0
+                .into_u64()
+                .checked_mul(u64::from(address_size))
+                .ok_or(Error::UnsupportedOffset)?,
         )?)?;
         input.read_address(address_size)
     }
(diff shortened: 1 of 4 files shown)


src/read/addr.rs
```diff
@@ -310,6 +314,19 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_get_address_index_overflow() {
+        // `index * address_size` must not overflow when the index is attacker
+        // controlled; it should be reported as an unsupported offset instead.
+        let buf = [0u8; 64];
+        let debug_addr = DebugAddr::from(EndianSlice::new(&buf, LittleEndian));
+        let index = DebugAddrIndex(0x2000_0000_0000_0000usize);
```
philipc, line 324: This is failing CI. I don't think it's possible to achieve the overflow with 32-bit `EndianSlice`, so it's probably appropriate to add `#[cfg(target_pointer_width = "64")]` to these tests.

philipc (changes requested):
Thanks, this is a good correctness fix. I don't think the attacker controlled behaviour is a concern though; the worse it can cause is a panic if overflow checks are enabled. If they control the data then they can already cause the parser to return any address they want.

src/read/addr.rs
alhudz: Done, added `#[cfg(target_pointer_width = "64")]` to both. You're right that a 32-bit `usize` index can't reach the overflowing value, so the tests only make sense on 64-bit.

alhudz:
Agreed on the framing, the parser already trusts the data so this isn't a security boundary. I've dropped the attacker-controlled wording from the description and test comments and kept it as a plain correctness fix for the panic under overflow checks.

philipc (approved):

PR #895, opened 2026-06-25
read/dwarf: avoid overflow when computing high_pc range end

alhudz:
Repro: call `Dwarf::die_ranges` (or `unit_ranges`) on a DIE whose `DW_AT_low_pc` is near the top of the address space and whose `DW_AT_high_pc` is given as an offset (`DW_FORM_udata`), large enough that `low_pc + high_pc` is past `u64::MAX`.
Cause: the range end is computed with an unchecked `begin + size`, so it panics under overflow checks and wraps to an `end` below `begin` in release. Every other range computation here goes through `wrapping_add_sized`; this was the one site using a raw `+`.
Fix: compute the end with `begin.checked_add(size)` and drop the range when it overflows, matching the existing handling for a missing end.

### DIFF
--- a/src/read/dwarf.rs
+++ b/src/read/dwarf.rs
@@ -626,7 +626,10 @@ impl<R: Reader> Dwarf<R> {
             }
         }
         let range = low_pc.and_then(|begin| {
-            let end = size.map(|size| begin + size).or(high_pc);
+            let end = match size {
+                Some(size) => begin.checked_add(size),
+                None => high_pc,
+            };
             // TODO: perhaps return an error if `end` is `None`
             end.map(|end| Range { begin, end })
         });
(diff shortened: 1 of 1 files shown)


src/read/dwarf.rs
```diff
@@ -1740,4 +1743,79 @@ mod tests {
         }
         assert_eq!(dwarf.format_error(Error::Io), format!("{}", Error::Io));
     }
+
+    #[test]
+    fn test_die_ranges_high_pc_overflow() {
```
philipc: This is a lot of boilerplate just to check that `u64::checked_add` does what it is meant to, for a case that is unlikely to occur in practice. If we're going to add this sort of test, we should also cover the working cases at the same time. Also ideally the test would start with `Dwarf` instead of constructing a `Unit` by hand. It's probably fine to even use the `write` module to test this (`test_assembler` would be better, but we haven't got any concise extensions methods for this).

alhudz: Fair point. Reworked it to build the unit with the `write` module and start from `Dwarf`, which drops most of the boilerplate. It now exercises a normal `high_pc` offset (`begin..begin + offset`) alongside the overflow case in one helper. Pushed.

philipc (approved):
Thanks!

PR #896, opened 2026-07-04
read/cfi: avoid underflow when computing eh_frame offset

alhudz:
1. `pointer_to_offset` subtracts `eh_frame_ptr` from the looked-up FDE pointer with a raw `-`, and both come straight from the parsed `.eh_frame_hdr`.
2. an FDE pointer below `eh_frame_ptr` makes that underflow: panics under overflow checks, wraps to a bogus offset in release.
Used `checked_sub` returning `UnsupportedOffset`, the way `resolve_cie_offset` already handles the analogous subtraction.


philipc (approved):
Thanks. This is fine as a pedantic fix, but mostly would have been caught as an error by a subsequent fde_from_offset call, and additionally it is easy for the pointers here to have values that pass our checks but still cause the unwinder to crash.

PR #897, opened 2026-07-04
read/cfi: check for overflow in eh_frame_hdr table stride

alhudz:
Repro: parse a `.eh_frame_hdr` whose `fde_count` is near `u64::MAX` with a `udata8` table encoding, then call `EhHdrTable::lookup`.
Cause: `lookup` computes the binary-search stride as `(len / 2) * row_size` with `len` taken straight from the parsed `fde_count` (unbounded, up to `u64::MAX`) and `row_size` up to `16`, so the `u64` multiply overflows before the offset is bounds-checked by `split`. It panics under overflow checks and wraps to a bogus in-bounds offset in release. `EhHdrTableIter::nth` has the same unchecked `n * row_size`.
Fix: compute both strides with `checked_mul(...).ok_or(Error::UnsupportedOffset)?`, matching the index resolvers.

Added `test_eh_frame_hdr_lookup_fde_count_overflow`, which panics with `attempt to multiply with overflow` at `cfi.rs:363` before the change.

### DIFF
--- a/src/read/cfi.rs
+++ b/src/read/cfi.rs
@@ -166,6 +166,25 @@ impl<R: Reader> EhFrameHdr<R> {
             fde_count = parse_encoded_value(fde_count_enc, &parameters, &mut reader)?;
         }
 
+        // Validate that the table contains exactly `fde_count` entries so that
+        // later offset computations in `EhHdrTable` can't overflow. Trailing
+        // data after the table is permitted.
+        let mut table = reader;
+        if fde_count != 0 {
+            let row_size = match table_enc.format() {
+                constants::DW_EH_PE_sdata2 | constants::DW_EH_PE_udata2 => Some(4),
+                constants::DW_EH_PE_sdata4 | constants::DW_EH_PE_udata4 => Some(8),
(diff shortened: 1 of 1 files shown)


philipc (changes requested):
I think it's better to validate `fde_count` in `EhFrameHdr::parse` instead by ensuring that `ParsedEhFrameHdr::table` contains exactly `fde_count` entries (`split` at the required length, it's okay to have trailing data).

Additionally, `nth` should return `Ok(None)` for an out of range index (replace the `remain.saturating_sub(n)`).

alhudz:
Reworked it your way: `EhFrameHdr::parse` now truncates the table to `fde_count` entries (`checked_mul` for the length, then `truncate`), so a `fde_count` that overflows or runs past the section is rejected up front and the later `EhHdrTable` offsets can't overflow. Dropped the `checked_mul` guards in `lookup`/`nth` since they're bounded now, and `nth` returns `Ok(None)` for an out of range index. Trailing data after the table is still allowed. Added tests for the overflow, the too-large `fde_count`, and the out of range `nth`.

philipc (approved):
Thanks!

alhudz:
thanks for merging, validating `fde_count` at parse time ended up the tidier fix

